PR #282, opened 2021-06-15
Fix(FixedFooterLayout, FeedbackScreen): update some styles that were causing problems in page transition animations

pladaria:
Related WebApp PR: https://github.com/Telefonica/webapp/pull/577

Problem:
https://user-images.githubusercontent.com/579417/122053112-be12b000-cde6-11eb-8197-2a320aead73e.mp4

With this fix:
**(updated)**
https://user-images.githubusercontent.com/579417/122101502-abaf6b00-ce14-11eb-8365-6860c8342fe6.mp4

Also verified that nothing breaks in appointments booking

### DIFF
--- a/size-stats.json
+++ b/size-stats.json
@@ -1,11 +1,11 @@
 {
     "dist": {
-        "js": "5400.38 KB",
-        "jsNoMisticaIcons": "753.64 KB"
+        "js": "5400.60 KB",
+        "jsNoMisticaIcons": "753.85 KB"
     },
     "distEs": {
-        "js": "3091.88 KB",
-        "jsNoMisticaIcons": "556.37 KB"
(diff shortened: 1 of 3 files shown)


atabel:
I still see some weird behaviors in the transitions.
Just before transitioning:
image
A few frames after starting transition:
image
The navbar is removed and the whole content of the screen is moved up, also a blue div appears at the bottom

atabel (approved):

pladaria:
> The navbar is removed and the whole content of the screen is moved up, also a blue div appears at the bottom

hmm... will review that... instead of `top: 0`, perhaps it should be a `top: navbarHeight` (note the color change at the top in the "before" video)

PR #286, opened 2021-06-21
fix(ButtonFixedFooterLayout): align footer buttons in desktop

atabel:
jira: WEB-197


src/feedback-screen.tsx
```diff
@@ -77,7 +78,7 @@ const useStyles = createUseStyles((theme) => ({
 
     innerContainer: {
         textAlign: 'left',
-        padding: '64px 24px 16px',
+        padding: '64px 8px 16px',
```
pladaria, line 81: shouldn't this produce changes in mobile screenshots?

atabel: nop, I've added a `<ResponsiveLayout />` wich adds `16px` horizontal padding in mobile: `16 + 8 = 24`

pladaria (approved):

PR #287, opened 2021-06-21
fix(CreditCardExpirationField): show mandatory field error message when empty

atabel:
Fix cc expiration field error message and add tests
jira: WEB-185

### DIFF
--- a/src/__tests__/credit-card-fields-test.tsx
+++ b/src/__tests__/credit-card-fields-test.tsx
@@ -1,0 +1,212 @@
+import * as React from 'react';
+import {fireEvent, render, screen, waitFor} from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import {ButtonPrimary, Form, CreditCardFields} from '..';
+import ThemeContextProvider from '../theme-context-provider';
+import {makeTheme} from './test-utils';
+
+test('Credit card fields validation, all fields empty', async () => {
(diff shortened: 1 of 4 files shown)


src/__tests__/credit-card-fields-test.tsx
```diff
@@ -0,0 +1,212 @@
...
+                    ccExp: {
+                        month: 11,
+                        year: 2050,
+                        raw: '11/50',
+                    },
+                    ccCvv: cvv,
+                },
+                {ccNum: expect.any(String), ccExp: '11/50', ccCvv: cvv}
+            );
+            expect(onValidationErrorsSpy).toHaveBeenCalledWith({});
```
atabel, line 202: not sure if this call is expected here, but the current implementation does it

pladaria (approved):

src/__tests__/credit-card-fields-test.tsx
pladaria: I'd consider that a bug of the implementation. By looking at webapp usages of this prop, no one expects this to be called without errors.

There are two uses:
1. in appointment booking, to scroll into the field with the error (this shouldn't be needed because the form should automatically do this)
2. for analytics, to track form errors (if no errors, nothing is sent)

pladaria (approved):

PR #290, opened 2021-06-23
feat(Callout): new component

atabel:
jira: WEB-202


yceballost (changes requested):
I found a bug when you increase font size, closable action is displaced by the title
image

I'm going to change close icon spacing from
top: 8px
right: 8px

to

top: 16px
right: 12px

To avoid the align problem with one line version. I'm going to update specs.

Lastly, in storybook you marked description form as "optional" but description is always need it in Callouts. Title, asset, actions and dismiss action are the optional items.

I'm pending to review animation transition in dismiss action

src/callout.tsx
```diff
@@ -0,0 +1,99 @@
...
+import IconButton from './icon-button';
+import classNames from 'classnames';
+import {ButtonLink, ButtonPrimary, ButtonSecondary} from './button';
+import Inline from './inline';
+
+const useStyles = createUseStyles(({colors}) => ({
+    container: {
+        background: colors.backgroundAlternative,
+        borderRadius: 4,
+        padding: 16,
```
SergioMorchon, line 17: IIRC this can be done with `Boxed`, also dealing with choosing the outer inverse vs inner inverse

src/__screenshot_tests__/callout-screenshot-test.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+    await page.select(await screen.findByLabelText('actions'), 'none');
+
+    const callout = await screen.findByRole('alert');
+
+    const image = await callout.screenshot();
+    expect(image).toMatchImageSnapshot();
+});
+
+test('Callout only description', async () => {
```
pladaria, line 71: I'd make a test case "only description and closable"

Perhaps you could rewrite all these tests using a `test.each + table`

src/__screenshot_tests__/callout-screenshot-test.tsx
```diff
@@ -0,0 +1,85 @@
...
+    const page = await openStoryPage({
+        section: 'Components/Dialogs/Callout',
+        name: 'Callout',
+    });
+    await page.clear(await screen.findByLabelText('title (opcional)'));
+    await page.select(await screen.findByLabelText('actions'), 'none');
+    await page.click(await screen.findByLabelText('With icon'));
+    await page.click(await screen.findByLabelText('Is closable'));
```
pladaria, line 79: Hard to know if this is making it closable or not

src/callout.tsx
```diff
@@ -0,0 +1,99 @@
...
+                                    {buttonLink}
+                                </Inline>
+                            </div>
+                        )}
+                    </Stack>
+                </div>
+                {onClose && (
+                    <IconButton
+                        size={40}
+                        style={{
```
pladaria, line 82: This could be a class. If IconButton doesn't support a className prop, consider adding it

src/callout.tsx
```diff
@@ -0,0 +1,99 @@
...
+    const isInverse = useIsInverseVariant();
+    const classes = useStyles({needsButtonLinkAlignment});
+    const {colors, texts} = useTheme();
+    return (
+        <section role="alert" className={classNames(classes.container, {[classes.overInverse]: isInverse})}>
+            <ThemeVariant isInverse={false}>
+                {icon && <div className={classes.icon}>{icon}</div>}
+                <div className={classes.content}>
+                    <Stack space={16}>
+                        <div>
```
pladaria, line 61: I think this could be a fragment

src/callout.tsx
```diff
@@ -0,0 +1,99 @@
...
+                            </div>
+                        )}
+                    </Stack>
+                </div>
+                {onClose && (
+                    <IconButton
+                        size={40}
+                        style={{
+                            display: 'flex',
+                            margin: '-8px -12px -8px 0',
```
pladaria, line 84: perhaps it would be better a `position: 'absolute', right, top`

src/callout.tsx
atabel: IconButton uses `style` internally, so I can't override some styles with className :(
we could review IconButton, but that's a different task

src/callout.tsx
atabel: no, because we need a space of 8px between text and X button, so I think this is a better approach

src/callout.tsx
atabel: no, because Boxed uses different background color logic

yceballost (approved):
good job ✨

src/callout.tsx
atabel: yes, changed

src/__screenshot_tests__/callout-screenshot-test.tsx
atabel: Ok I'll include that new case, but I think I prefer having mistica form fields than story arg

src/__screenshot_tests__/callout-screenshot-test.tsx
atabel: fixed

src/__screenshot_tests__/callout-screenshot-test.tsx
atabel: done

SergioMorchon (approved):

pladaria (approved):

PR #293, opened 2021-07-01
chore(PreviewTools): Set controls theme to light

pladaria:
Controls theme was not forced to light, this fixes it.
Verified on Mac safari, using dark/light modes

### DIFF
--- a/playroom/components.tsx
+++ b/playroom/components.tsx
@@ -12,7 +12,6 @@ import {
     Checkbox,
     ThemeContextProvider,
 } from '../src';
-import {useIsOsDarkModeEnabled} from '../src/theme-context-provider';
 import {Movistar, Vivo, O2, Telefonica} from './themes';
 import {useOverrideTheme} from './frame-component';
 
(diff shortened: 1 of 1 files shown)


yceballost (approved):
working perfectly in Brand Factory

playroom/components.tsx
```diff
@@ -89,7 +88,7 @@ const useStyles = createUseStyles(() => ({
     },
 }));
 
-const themesMap: {[skinName in SkinName]: {themeConfig: ThemeConfig; text: string}} = {
+const themesMap: {[skinName: string]: {themeConfig: ThemeConfig; text: string}} = {
```
pladaria, line 91: TS was reporting this as an error, because O2-classic is missing in the themesMap object

atabel, line 91: why this change?

atabel (approved):

playroom/components.tsx
atabel: oh, there is a way to make this non-exhaustive?

atabel: also, why don't we run tsc on CI? :/

pladaria: we are running tsc but it only checks files used by the library

PR #294, opened 2021-07-01
fix(FeedbackScreen): text animation

atabel:
jira: WEB-256


src/feedback-screen.tsx
```diff
@@ -205,13 +203,13 @@ export const FeedbackScreen: React.FC<FeedbackScreenProps> = ({
             <ResponsiveLayout>
                 <div className={classes.innerContainer}>
                     {!!icon && <div className={classes.iconContainer}>{icon}</div>}
-                    <Text6>
-                        <span className={classes.title}>{title}</span>
```
atabel, line 209: css `transform: translate` doesn't affect to inline elements

yceballost (approved):
working in desktop, ios device and android device browsers

pladaria (approved):

PR #295, opened 2021-07-01
WEB-260 fix button layout

pladaria:


src/button-layout.tsx
```diff
@@ -161,6 +161,25 @@ const ButtonLayout: React.FC<ButtonLayoutProps> = ({
...
+    /**
+     * Listening to resize to force a layout calculation when the available space changes
+     * otherwise it can happen that small buttons with ellipsis keep the ellipsis when the screen size is increased
+     *
+     * Listening to focus/visibility change solves a corner that can be reproduced in Novum iOS webviews. Just after logging in,
+     * wait until everything loads (including hidden tabs), after a while, open Account tab (Mis productos) and the button appears
+     * with ellipsis, even it has enough space.
+     */
+    React.useEffect(() => {
+        window.addEventListener('resize', calcLayout);
```
atabel, line 173: maybe we should use a resize observer instead of just listening to window resize, but if this works it's fine to me. I think almost all the cases where the button layout resizes is because the window resizes, so maybe this is good enough

src/__stories__/button-layout-story.tsx
```diff
@@ -41,58 +41,58 @@ export const Default: StoryComponent = () => (
         <div data-testid="screenshot">
             <StorySection title="ButtonLayout">
                 <ButtonLayout>
-                    <ButtonSecondary fake>ok</ButtonSecondary>
-                    <ButtonPrimary fake>long long long</ButtonPrimary>
+                    <ButtonSecondary fake>Ok</ButtonSecondary>
+                    <ButtonPrimary fake>The text in this button is very long</ButtonPrimary>
```
atabel, line 45: In the mobile screenshots of this story I miss a case where both buttons are side by side

atabel (approved):

src/button-layout.tsx
pladaria: I considered to use the resize observer, but this is simpler and doesn't require polyfills. And the polyfill we're using is asynchronously loaded and that makes everything more complicated. Unless we find a case where this is not enough, I'd leave this as is.

src/__stories__/button-layout-story.tsx
pladaria: ok

atabel (approved):

PR #298, opened 2021-07-07
WEB-226 Tabs line animation, dataAttributes in touchable

pladaria:


src/tabs.tsx
```diff
@@ -117,26 +110,63 @@ export type TabsProps = {
...
 const Tabs: React.FC<TabsProps> = ({selectedIndex, onChange, tabs}: TabsProps) => {
+    const id = useAriaId();
     const {width, ref} = useElementDimensions();
-    const classes = useStyles({width, numTabs: tabs.length});
+    const animatedLineRef = React.useRef<HTMLDivElement>(null);
+    const scrollableContainerRef = React.useRef<HTMLDivElement>(null);
+    const [isAnimating, setIsAnimating] = React.useState(false);
+    const classes = useStyles({width, numTabs: tabs.length, isAnimating});
+
+    const animateLine = (fromIndex: number, toIndex: number) => {
```
pladaria, line 120: This function animates the line:
1. sets initial line position and hides the line from the tab (setting the isAnimating state)
2. sets final line position and gets animated
3. once the animation ends, animated line is hidden and the line from the tab is shown (set isAnimating false)

yceballost (approved):
smoooth! ✨

src/touchable.tsx
```diff
@@ -123,6 +126,16 @@ export interface PropsMaybeOnPress extends CommonProps {
 
 type Props = PropsHref | PropsTo | PropsOnPress | PropsMaybeHref | PropsMaybeTo | PropsMaybeOnPress;
 
+const getPrefixedDataAttributes = (attrs?: DataAttributes) => {
```
pladaria, line 129: this could be moved to utils/dom.tsx if needed by other components following this convention

atabel (approved):
nice!
Regarding `dataAttributes` prop, I'd create a ticket to standardize this approach, we should review other mistica components accepting `data-` props and refactor them to this approach to have a consistent api across components.

PR #300, opened 2021-07-13
WEB-271 migrate to dataAttributes

pladaria:
instead of defining `data-whatever` props in our components, we want to use a dictionary `dataProperties` to be able to pass any `data attribute` to the component

### DIFF
--- a/src/box.tsx
+++ b/src/box.tsx
@@ -1,6 +1,9 @@
 import React from 'react';
 import {createUseStyles} from './jss';
 import classnames from 'classnames';
+import {getPrefixedDataAttributes} from './utils/dom';
+
+import type {DataAttributes} from './utils/types';
 
 const useStyles = createUseStyles(() => ({
     box: {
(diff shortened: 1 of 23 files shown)


src/touchable.tsx
```diff
@@ -126,16 +127,6 @@ export interface PropsMaybeOnPress extends CommonProps {
 
 type Props = PropsHref | PropsTo | PropsOnPress | PropsMaybeHref | PropsMaybeTo | PropsMaybeOnPress;
 
-const getPrefixedDataAttributes = (attrs?: DataAttributes) => {
```
pladaria, line 129: moved to utils/dom.tsx

src/box.tsx
```diff
@@ -23,8 +26,11 @@ type Props = {
     children?: React.ReactNode;
     className?: string;
     role?: string;
+    // @deprecated use dataAttributes
     'data-testid'?: string;
+    // @deprecated use dataAttributes
     'data-qsysid'?: string;
+    dataAttributes?: DataAttributes;
```
atabel, line 33: did you miss the jsdoc here?

atabel (approved):

anabelengp (approved):

src/box.tsx
pladaria: added

PR #301, opened 2021-07-15
feat(MasterDetailLayout): new component. 6+4 and 4+6 GridLayout

atabel:
jira ticket: WEB-276 
This new component, based on `GridLayout` (used internally), covers the 4/6 case described in the [layout spec](https://brandfactory.telefonica.com/document/1846#/fundamentals/layout):
image


pladaria (approved):

yceballost (changes requested):
- Missing Grid Layout 4+6 in Storybook

Why do you create masterDetailLayout instead of include all the logic directly in GridLayout template="4+6"?

atabel:
> * Missing Grid Layout 4+6 in Storybook
> 
That will be done in a different PR
> Why do you create masterDetailLayout instead of include all the logic directly in GridLayout template="4+6"?
Because that would add additional logic to the `GridLayout` component, and we prefer to keep it focused in just the columns layout. Also, you may want a 4+6 layout without the logic of `MasterDetail` (two navigation levels in mobile). Pedro and I discussed that possibility, I paste here part of that conversation:
image
image

atabel:
I've completed the layout doc with this new component: https://github.com/Telefonica/mistica-web/blob/atoledano-master-detail/doc/layout.md#masterdetaillayout

pladaria (approved):

anabelengp (approved):

yceballost:
> > * Missing Grid Layout 4+6 in Storybook
> 
> Ok, I'll add it
> 
> > Why do you create masterDetailLayout instead of include all the logic directly in GridLayout template="4+6"?
> 
> Because that would add additional logic to the `GridLayout` component, and we prefer to keep it focused in just the columns layout. Also, you may want a 4+6 layout without the logic of `MasterDetail` (two navigation levels in mobile). Pedro and I discussed that possibility, I paste here part of that conversation:
> image
> image

check this @HectorSancho

yceballost (changes requested):
I think would be nice to add a vertical divider between "areas" like in this design
image

atabel:
> I think would be nice to add a vertical divider between "areas" like in this design

That can be done inside `master` view if needed, I wouldn't include that inside the `MasterDetailLayout`

atabel:
I've included the 6+4 layout in this PR too

playroom/use-scope.tsx
```diff
@@ -45,12 +45,14 @@ const usePlayroomState = () => {
...
 const useScope = (): {theme: Theme; colors: Colors} => {
     const theme = useTheme();
+    const screenSize = useScreenSize();
     const playroomState = usePlayroomState();
 
     return {
         theme,
         colors: theme.colors,
         ...playroomState,
+        ...screenSize,
```
atabel, line 55: @pladaria, this allows us to create playrooms using `isMobile`, `isTabletOrBigger`, etc

HectorSancho (approved):

yceballost (approved):

pladaria (approved):

PR #303, opened 2021-07-16
feat(Stack): Support `between`, `evenly` and `around` space

pladaria:


src/__stories__/stack-story.tsx
```diff
@@ -1,16 +1,29 @@
 import * as React from 'react';
```
atabel, line 1: please, review this story in dark mode, the text isn't readable

src/__stories__/stack-story.tsx
```diff
@@ -1,16 +1,29 @@
...
-import {Stack, Text2, useTheme} from '..';
-import {StorySection, useSelect} from './helpers';
-import {Placeholder} from '../placeholder';
+import {Stack, Text2} from '..';
 
 export default {
     title: 'Components/Layouts/Stack',
+    component: Stack,
+    argTypes: {
```
atabel, line 7: If we are using args here, we could do the same with the Inline story

atabel (approved):

yceballost (approved):
nice!

PR #306, opened 2021-07-28
feat(FormFields): use rifm to format CreditCard and Decimal fields

anabelengp:
related to #191 

Applied to CreditCardInput and DecimalInput (used by CvvField). Not applied to CreditCardExpirationField since it provokes more problems than what it solves...
**before**
creditcardfields-before-rifm

**after**
creditcardfields-after-rifm

### DIFF
--- a/src/__acceptance_tests__/form-fields-acceptance-test.tsx
+++ b/src/__acceptance_tests__/form-fields-acceptance-test.tsx
@@ -62,6 +62,15 @@ test.each(STORY_TYPES)('DecimalField (%s)', async (storyType) => {
 
     await clearAndType(page, field, '+-123e.4,5.6i');
     await expect(getValue(field)).resolves.toBe('123,456');
+
+    await clearAndType(page, field, '124,5');
+    await expect(getValue(field)).resolves.toBe('124,5');
+    await field.evaluate((el) => {
+        // set the caret position after the first block
+        (el as HTMLInputElement).setSelectionRange(2, 2);
+    });
+    await field.type('.3');
+    await expect(getValue(field)).resolves.toBe('12,345');
 });
 
 test.each(STORY_TYPES)('CreditCardNumberField (%s)', async (storyType) => {
--- a/src/__acceptance_tests__/form-fields-acceptance-test.tsx
+++ b/src/__acceptance_tests__/form-fields-acceptance-test.tsx
@@ -74,6 +74,15 @@ test.each(STORY_TYPES)('CreditCardNumberField (%s)', async (storyType) => {
 
     await clearAndType(page, field, '1234-567.8 9012/34abc567 890');
     await expect(getValue(field)).resolves.toBe('1234 5678 9012 3456');
+
+    await clearAndType(page, field, '123456789012');
+    await expect(getValue(field)).resolves.toBe('1234 5678 9012');
+    await field.evaluate((el) => {
+        // set the caret position after the first block
+        (el as HTMLInputElement).setSelectionRange(4, 4);
+    });
+    await field.type('1234');
+    await expect(getValue(field)).resolves.toBe('1234 1234 5678 9012');
 });
 
 test.each(STORY_TYPES)('CreditCardExpirationField (%s)', async (storyType) => {
--- a/src/__acceptance_tests__/form-fields-acceptance-test.tsx
+++ b/src/__acceptance_tests__/form-fields-acceptance-test.tsx
@@ -63,14 +63,25 @@ test.each(STORY_TYPES)('DecimalField (%s)', async (storyType) => {
     await clearAndType(page, field, '+-123e.4,5.6i');
     await expect(getValue(field)).resolves.toBe('123,456');
 
+    // test editing the number to set the decimal char in a previous pos
     await clearAndType(page, field, '124,5');
     await expect(getValue(field)).resolves.toBe('124,5');
     await field.evaluate((el) => {
-        // set the caret position after the first block
+        // set the caret position on a digit before the ,
         (el as HTMLInputElement).setSelectionRange(2, 2);
     });
     await field.type('.3');
     await expect(getValue(field)).resolves.toBe('12,345');
+
+    // test editing the number to set the decimal char in a later pos
+    await clearAndType(page, field, '1,24');
+    await expect(getValue(field)).resolves.toBe('1,24');
+    await field.evaluate((el) => {
+        // set the caret position on a digit after the ,
+        (el as HTMLInputElement).setSelectionRange(3, 3);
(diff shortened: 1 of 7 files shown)


src/credit-card-number-field.tsx
```diff
@@ -17,39 +17,57 @@ import {createUseStyles} from './jss';
...
-    // Naive implementation, some issues in cursor position when editing
-    const format = (s?: string) => {
-        const chars = String(s ?? '')
-            .replace(/[^\d]/g, '')
-            .slice(0, 16)
-            .split('');
+const format = (s?: string) => {
+    const sanitizedNumber = String(s ?? '').replace(/[^\d]/g, '');
+    return sanitizedNumber.match(/.{1,4}/g)?.join(' ') ?? sanitizedNumber;
+};
```
anabelengp, line 27: I've changed the code of format function, now it is based on `formatIban` implementation, to keep them similar: https://github.com/Telefonica/mistica-web/blob/master/src/iban-field.tsx#L76-L79

src/credit-card-number-field.tsx
```diff
@@ -17,39 +17,57 @@ import {createUseStyles} from './jss';
...
-            maxLength={maxLength ?? '19'} // 16 digits + 3 spaces
-            onInput={(e) => {
-                e.currentTarget.value = format(e.currentTarget.value);
-                onInput?.(e);
-            }}
-            value={value === undefined ? undefined : format(value)}
-            defaultValue={defaultValue === undefined ? undefined : format(defaultValue)}
-            ref={inputRef}
+            maxLength={getCreditCardNumberLength(rifm.value) + 3} // We have to take in account formatting spaces
+            onChange={rifm.onChange}
```
anabelengp, line 68: I've changed the event to listen, as I think `onChange` event is preferred to `onInput` in react, and I've seen it used in other similar components

src/decimal-field.tsx
```diff
@@ -48,19 +56,44 @@ export const DecimalInput: React.FC<DecimalInputProps> = ({inputRef, value, defa
         return parts.shift() + localDecimalChar + parts.join('');
     };
 
+    const replace = (value: any) => String(value ?? '').replace(/[.,]/g, localDecimalChar); // use local decimal char
```
anabelengp, line 59: replace the localDecimalChar in a `replace` input for rifm to prevent the caret from being kept in the position before the delimiter: https://github.com/realadvisor/rifm/issues/65#issuecomment-503073678

atabel (commented):
I've seen a weird behavior in DecimalFiled:
- write 1,23456
- place the cursor between 2 and 3
- start pressing the `.` key
The cursor moves one position on every keystroke and an `onChangeValue` receives a value with an additional comma

src/decimal-field.tsx
```diff
@@ -37,25 +56,6 @@ export const DecimalInput: React.FC<DecimalInputProps> = ({
...
-
-        if (parts.length === 1) {
-            // no fractional part, return "as is"
-            return parts[0];
-        }
-
-        // value includes one or more decimal separators, keep the first one
-        return parts.shift() + localDecimalChar + parts.join('');
-    };
-
```
anabelengp, line 50: The bug arose because I deleted the line which made the replacement by the localDelimiterChar. That replacement prevented rifm from working properly, that is why I moved it to rifm's replace input. It seems that rifm doesn't accept replacements in format functions. 

I've moved format function outside the component since localDecimalChar is no longer necessary in it. This way, in future revisions, developers won't be tempted to make that replacement in the format function

src/decimal-field.tsx
```diff
@@ -17,6 +17,25 @@ const getLocalDecimalChar = (locale: Locale): string => {
     }
 };
 
+const format = (value: any) => {
+    const sanitized = String(value ?? '').replace(/[^.,\d]/g, ''); // remove non digits or decimal separator chars;
+    const firstSeparator = /[.,]/.exec(sanitized);
+    const parts = sanitized.split(/[.,]/);
```
anabelengp, line 23: do not make the localDecimalChar  replacement here, instead, keep the one the user typed. Make that replacement in `replace` prost-processor

src/decimal-field.tsx
atabel: I'd also include a comment explaining this

src/__acceptance_tests__/form-fields-acceptance-test.tsx
```diff
@@ -62,6 +62,15 @@ test.each(STORY_TYPES)('DecimalField (%s)', async (storyType) => {
 
     await clearAndType(page, field, '+-123e.4,5.6i');
     await expect(getValue(field)).resolves.toBe('123,456');
+
+    await clearAndType(page, field, '124,5');
+    await expect(getValue(field)).resolves.toBe('124,5');
+    await field.evaluate((el) => {
+        // set the caret position after the first block
```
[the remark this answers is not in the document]
anabelengp: fixed

src/decimal-field.tsx
anabelengp: will this do?? https://github.com/Telefonica/mistica-web/pull/306/files#diff-9aec20dcbd46eed10b4693dec90d2e2a9970b54e570ff4b38f2b20495eaa622cR21-R22

atabel:
image
I think we shouldn't allow a comma as first char

atabel:
why don't we use `IntegerField` instead of `DecimalField` in `CvvField`? :/

atabel (commented):
see
https://github.com/Telefonica/mistica-web/pull/306#issuecomment-889887264
and
https://github.com/Telefonica/mistica-web/pull/306#issuecomment-889888563

anabelengp:
mm I thought accepting a comma as first char was an agreed behaviour. I saw this other implmementation which I think is more convenient: https://realadvisor.github.io/rifm/. It is more restrictive.
```
// allow only one floating point
format={v => (v.match(/\d+[.,]?\d*/) || []).join('')}
```
Shall I change it??

atabel:
> mm I thought accepting a comma as first char was an agreed behaviour. I saw this other implmementation which I think is more convenient: [realadvisor.github.io/rifm](https://realadvisor.github.io/rifm/). It is more restrictive.
> 
> ```
> // allow only one floating point
> format={v => (v.match(/\d+[.,]?\d*/) || []).join('')}
> ```
> 
> Shall I change it??

hmmm... well I'm not sure to be honest. Maybe we could accept `'.123'` as a valid version of `'0.123'`. Calculator use to accept that, so I think it could be fine.

atabel (approved):

pladaria (approved):

PR #315, opened 2021-08-16
feat(Rows): Allow extra content in rows

pladaria:


flow-defs/list.js.flow
```diff
@@ -10,10 +10,11 @@ declare type CommonProps = {|
   subtitleLinesMax?: number,
   description?: string | null,
   descriptionLinesMax?: number,
-  icon?: React.Element<any> | string | null,
+  icon?: React.Node,
   iconSize?: 24 | 40,
   badge?: boolean | number,
   role?: string,
+  extra?: React.Node,
```
yceballost, line 17: why don't we call these kind of areas "fragment" instead of "extra"?

In cards we called "Body" 
In headers we called "Extra"

yceballost (changes requested):

SergioMorchon (approved):

flow-defs/list.js.flow
pladaria: I used "extra" because headers use the "extra" naming. I don't like to use "fragment" because in react a "fragment" is something different

yceballost: should we change the "body" name on the card props to "extra"?

yceballost (approved):

flow-defs/list.js.flow
pladaria: ok, I will deprecate the body prop

pladaria: Done, I also prefer the "extra" naming because it refers to something "extraordinary"

yceballost: Cool, is 👍 for me. I prefer fragment to be the same as design documentation and libraries but If in react is used for something different we can go with extra, no problem.

PR #317, opened 2021-08-18
feat(Rows): Allow double interactions: onPress + switch/checkbox

pladaria:
Preview here: https://mistica-web-pladaria.vercel.app/

For some reason preview deploys in github action are failing, will investigate.

Looks like vercel tokens are no longer valid. I've updated the tokens to use my own account until @atabel comes back from vacations.


src/form.tsx
```diff
@@ -119,7 +119,7 @@ const Form: React.FC<FormProps> = ({
         }
 
         setFormErrors(errors);
-        if (onValidationErrors) {
+        if (onValidationErrors && Object.keys(errors).length > 0) {
```
pladaria, line 122: unrelated fix, Form was always calling onValidationErrors on submit, it should only be called if some error exists

.storybook/.babelrc.js
```diff
@@ -19,5 +19,9 @@ module.exports = {
             },
         ],
     ],
+    assumptions: {
+        arrayLikeIsIterable: false,
+        iterableIsArray: false,
+    },
```
pladaria, line 25: fixes a problem in storybook compilation

src/__tests__/credit-card-fields-test.tsx
```diff
@@ -122,7 +122,7 @@ test.each`
                     },
                     {ccNum: '4321 4321 4321 4321', ccExp: expirationDate, ccCvv: '123'}
                 );
-                expect(onValidationErrorsSpy).toHaveBeenCalledWith({});
+                expect(onValidationErrorsSpy).not.toHaveBeenCalled();
```
Marcosld, line 125: Nice improvement

Marcosld (approved):

flow-defs/list.js.flow
```diff
@@ -28,10 +28,12 @@ declare type BasicRowContentProps = {|
 |};
 declare type SwitchRowContentProps = {|
   ...$Exact<CommonProps>,
+  onPress?: () => void,
```
yceballost, line 31: 1. Doesn't works with `href` prop. Is it expected?

https://mistica-web-pladaria.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8AlCB3AMgSwM4BcA%2BAHQDsACc1DMyy7SUgXmFrsqQGFsAnMAGxjkARgEMwAawDmPCAFdSUThH4QeLSKp64AdKRhz8PUf0wYAvuVzYAXjBYAWAAzmSFdnSQBJRgGUAFhAADigwUnL8ojzkAPRuHlQx3HyC8XTmbPR%2BtvbAzhnulPjY%2BIJMxCAAKiWCFZnksLhgPNhBxRDMFQAiME0tbdgddYXk-jwwAGblIP74%2BEG48DEx6Ks6UhAQUoI6vcPsHQAK47i4LAAUAJTkTITkwOYF7LjoJWD%2BLMANk6IR%2BABqJjkMHg5AmJlwQkebDibGo6HqDA6LHqHGSAiEYkkMnkimUWg0KjUun0hmMpgsVhyjhcaQS3j8gRCYQiUVidM8SV4GI5TzoSNIvmpeRc9WKpXsFWqEv2dEazVa7U6IB6fUVg1IssoRxOZ2AVxudwefMoLzeH2AX1g4L%2BgP4wNB4P4kPI0PcsNISBiaCweCIZBA5iAA

2. Only works for switches? Should we have the same behavior to radio controls?

yceballost (changes requested):

flow-defs/list.js.flow
pladaria: This only works with switch/checkbox + onPress. 

We don't have any other cases in our code, if needed we could implement it... but if not needed I prefer to not complicate the code.

SergioMorchon (approved):
Reading the `RowContent` and all its possible combinations feels like learing to pilot a spaceship 🚀

yceballost (approved):

PR #319, opened 2021-08-20
Feat(Rows): dataAttributes support

pladaria:


src/inline.tsx
```diff
@@ -38,6 +41,7 @@ type Props = {
     role?: string;
     'aria-labelledby'?: string;
     fullWidth?: boolean;
+    dataAttributes?: DataAttributes;
```
anabelengp, line 44: a naive question: `Inline` is a layout component, isn't it? so, in which use case is `dataAttributes` applied? should then `dataAttributes` be applied also to other layout components?

anabelengp (approved):
just a question in comments

claramunt (approved):

src/inline.tsx
pladaria: `dataAttributes` could be added to any component, but instead of changing all components, we are adding the prop when we need it.

PR #320, opened 2021-08-23
feat(Rows): images with aspect ratio in rows

pladaria:


yceballost (approved):
p e r f e c t !

AnaMontes11 (approved):

src/list.tsx
```diff
@@ -173,18 +170,12 @@ const Content: React.FC<ContentProps> = ({
 
     return (
         <Box paddingY={16} className={classes.content}>
```
anabelengp, line 172: I think that in figma they have specified a special alignment for the items in the row, depending on how many items are displayed in rowBody:
**from Figma:**
image

**Our result:**
image

Should we review this with @yceballost ??

anabelengp (approved):

src/list.tsx
pladaria: image
It was an error in spec definition. Specs have been updated

anabelengp: solved: fixed in specifications

PR #313, opened 2021-08-13
feat(NavigationBar): new navigation bar components

atabel:
jira: WEB-294
spec: https://app.zeplin.io/project/5d653c69f828bf7299c551c1/dashboard?seid=610d1eead31de315fbce7392

PR #322, opened 2021-08-27
feat(EmptyState): new component

anabelengp:
- [x]  Deprecate EmptyStateScreen (I've deprecated it because it was once released, (although I think it is not being used anywhere...)

-  [x] Add EmptyState component which is not fullpage in mobile (moreover, the name changes)
-  [x] **Extra** EmptyState enhancements (check if it is better to undo them):

1.  force largeImage aspect ratio inside emptyState component. It is a restriction described inside the EmptyState component and it should be the responsible one.
2.  add `dataAttributes` prop to EmptyState

     ```
      <EmptyState
            icon={<IconBoxLight size={isTabletOrSmaller ? 64 : 80} color={colors.brand} />}
            title="Your cart is empty"
            description="Check our marketplaces and find something for you. Check our marketplaces and find something"
            button={<ButtonPrimary onPress={() => {}}>Explore marketplace</ButtonPrimary>}
            dataAttributes={{testid: 'empty-state-with-icon'}}
        />
    ```

- [x] **Extra** warn developers that  `small` prop for emptyStateCard button is needed.
- [x] force the correct size for icon in EmptyState/EmptyStateCard
- [x] remove weird padding-right definition in EmtpyStateCard tests box. (updated in [zeplin](https://app.zeplin.io/project/5d653c69f828bf7299c551c1/screen/609aa630c92b9bbe2d99759))
- [ ] TODO: check emptyState comp inside master-detail layout: @yceballost what do you thing about this? https://bit.ly/2WQjVZo


playroom/snippets.tsx
```diff
@@ -679,7 +679,7 @@ const layoutSnippets: Array<Snippet> = [
 
 const emptyStatesGroup: Array<Snippet> = [
     {
-        name: 'EmptyStateScreen',
+        name: 'EmptyStateScreen (deprecated)',
```
pladaria, line 682: please, remove

src/__screenshot_tests__/empty-state-screen-screenshot-test.tsx
```diff
@@ -3,35 +3,41 @@ import type {Device} from '../test-utils';
...
 
-test.each(devices)('EmptyStateScreen with image', async (device) => {
-    const page = await openStoryPage({
-        id: 'components-screens-emptystatescreen--with-image',
-        device,
+/**
+ * @deprecated
+ */
+// eslint-disable-next-line jest/no-disabled-tests
+describe.skip('EmptyStateScreen component (deprecated)', () => {
```
pladaria, line 10: please, remove

src/__stories__/empty-state-screen-story.tsx
```diff
@@ -1,5 +1,6 @@
 import * as React from 'react';
```
pladaria, line 1: please, remove

src/empty-state-card.tsx
```diff
@@ -70,24 +70,32 @@ const EmptyStateCard: React.FC<Props> = ({
...
 
+    let resizedIcon;
+    if (icon) {
+        resizedIcon = React.cloneElement(icon, {size: isTabletOrSmaller ? 64 : 80, color: colors.brand});
+    }
+
+    let smallButton;
+    if (button) {
+        smallButton = React.cloneElement(button, {small: true} as Partial<React.FC<ButtonProps>>);
+    }
```
pladaria, line 81: I don't really like this, you're assuming here that you will receive an Icon component with a size prop, but this component accepts anything as icon (it could be, for example, an ``)

src/empty-state-card.tsx
```diff
@@ -70,24 +70,32 @@ const EmptyStateCard: React.FC<Props> = ({
         image = <img className={classes.image} alt="" src={imageUrl} />;
     }
 
+    let resizedIcon;
+    if (icon) {
+        resizedIcon = React.cloneElement(icon, {size: isTabletOrSmaller ? 64 : 80, color: colors.brand});
```
pladaria, line 75: I don't really like this, you're assuming here that you will receive an Icon component with a size prop, but this component accepts anything as an icon (it could be, for example, an )

src/empty-state-card.tsx
```diff
@@ -70,24 +70,32 @@ const EmptyStateCard: React.FC<Props> = ({
...
+    }
+
+    let smallButton;
+    if (button) {
+        smallButton = React.cloneElement(button, {small: true} as Partial<React.FC<ButtonProps>>);
```
pladaria, line 80: I prefer to not modify the received component, instead of this we could make a runtime check and show an error if not a production build, so it will show that error while developing and make tests fail when using a wrong combination

src/empty-state-screen.tsx
```diff
@@ -1,3 +1,9 @@
+/*
+!!! IMPORTANT !!!
+Do not add new features here!
+This component is deprecated and will be removed
```
pladaria, line 4: if not used, I'd remove it now. Otherwise someone could decide to use it, even if deprecated.

src/empty-state.tsx
```diff
@@ -0,0 +1,169 @@
...
+}) => {
+    const {colors} = useTheme();
+    const needsButtonLinkAlignment = buttonLink && !button;
+    const classes = useStyles({largeImageUrl, needsButtonLinkAlignment});
+    const {isTabletOrSmaller} = useScreenSize();
+
+    let resizedIcon;
+    if (icon) {
+        resizedIcon = React.cloneElement(icon, {size: isTabletOrSmaller ? 64 : 80, color: colors.brand});
+    }
```
pladaria, line 103: as commented before, I'd avoid that

src/empty-state.tsx
```diff
@@ -0,0 +1,169 @@
...
+                        </Inline>
+                    )}
+                </Stack>
+            </section>
+        );
+    }
+
+    return (
+        <ResponsiveLayout dataAttributes={dataAttributes}>
+            <Boxed>
```
pladaria, line 141: This component shouldn't have a ResponsiveLayout. The layout should be given by the feature that uses this component.

src/empty-state.tsx
```diff
@@ -0,0 +1,169 @@
...
+                            {buttonLink}
+                        </Inline>
+                    )}
+                </Stack>
+            </section>
+        );
+    }
+
+    return (
+        <ResponsiveLayout dataAttributes={dataAttributes}>
```
pladaria, line 140: This component shouldn't have a ResponsiveLayout. The layout should be given by the feature that uses this component.

src/empty-state-card.tsx
anabelengp: Ok, In this case, I'll  revert the change so the size won't be controlled from the component, and we will expect developers to use the correct size

src/empty-state-screen.tsx
anabelengp: I deprecated it to follow a correct versioning and to remark the change in the changelog, but if it isn't important  I'll simply rename the component and make the new changes to it

src/empty-state.tsx
anabelengp: OK! I didn't know if Responsivelayout was needed here for something, so I will remove it from this component

src/empty-state-card.tsx
pladaria: perhaps a better option would be to set the size of the container and assume that the developer will use an asset with 100% width (or the correct size)

src/empty-state-card.tsx
```diff
@@ -69,7 +72,9 @@ const EmptyStateCard: React.FC<Props> = ({
     if (imageUrl) {
         image = <img className={classes.image} alt="" src={imageUrl} />;
     }
-
+    if (process.env.NODE_ENV !== 'production' && !button?.props?.small) {
+        console.error('button property in EmptyStateCard must be a a small Button. Set small prop to true');
```
atabel, line 76: I like this, we should add a similar error message for Cards

src/empty-state-screen.tsx
atabel: I think we should start thinking on prepare a new major version. We already have a bunch of deprecated stuff and I also have a branch with a breaking change which is difficult to avoid (https://github.com/Telefonica/mistica-web/pull/313#discussion_r700914561)

src/empty-state.tsx
```diff
@@ -0,0 +1,161 @@
...
+        height: 128,
+        [theme.mq.tabletOrSmaller]: {
+            height: 112,
+        },
+    },
+}));
+
+interface BaseProps {
+    title: string;
+    button?: React.ReactElement<typeof ButtonPrimary> | React.ReactElement<typeof ButtonSecondary>;
```
atabel, line 51: `React.ReactElement<ButtonProps, typeof ButtonPrimary>`

src/empty-state.tsx
```diff
@@ -0,0 +1,161 @@
...
+    buttonLink?: React.ReactElement<typeof ButtonLink>;
```
atabel, line 52: Same here. Maybe we should have an eslint rule to avoid using React.ReactElement with only one generic param

src/empty-state-card.tsx
```diff
@@ -29,7 +30,9 @@ const useStyles = createUseStyles((theme) => ({
 
 interface CommonProps {
     title: string;
-    button?: React.ReactElement<typeof ButtonPrimary> | React.ReactElement<typeof ButtonSecondary>;
+    button?:
+        | React.ReactElement<ButtonProps, typeof ButtonPrimary>
```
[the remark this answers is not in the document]
anabelengp: done! https://jira.tid.es/browse/WEB-322

atabel (approved):

src/empty-state-card.tsx
```diff
@@ -18,19 +19,30 @@ const useStyles = createUseStyles((theme) => ({
...
+            height: 64,
+        },
+    },
     actions: {
         marginLeft: ({needsButtonLinkAlignment}) => (needsButtonLinkAlignment ? -6 : 0),
     },
 }));
 
 interface CommonProps {
     title: string;
```
yceballost, line 41: Empty state use different text style than Empty State Card. Please, check specs https://zpl.io/a3E4GMY

yceballost (changes requested):

src/empty-state-card.tsx
anabelengp: oops! done! :fearful:

yceballost (approved):
Perfect! Thanks!

src/empty-state-card.tsx
yceballost: > * [ ]  TODO: check emptyState comp inside master-detail layout: @yceballost what do you thing about this? [bit.ly/2WQjVZo](https://bit.ly/2WQjVZo)

We can use EmptyStateCard by the moment, there aren't cases with this problem right now, so.. imo, I dont try to solve it now

PR #335, opened 2021-09-09
fix(Select, TextField): fix label text overflow

atabel:
jira: WEB-283

some examples: https://mistica-web-pah4v30y5-tuentisre.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8AhCAPABABwIZSgEsA7AcwF5gBGANgF8A%2BAHWM0yQGUAXXMAa0wBnPGBiVajFmzZIAIhACuAIwA2MAGKEYKqJgBmClSoDqhKFwAWmFbgCeiruSYgALAAYA9DTfPmraTIc2jBgXFIB0jZK2k4gQWqhmPYKAE76uABuECmEXDD6KQq5zuERmBDYXIQQxIKUANqlZZjAmBm4Kgow8JjO2bhkMM4ANJh56Fw9zgAyNaQFRVyYxLgAtkMgmHTDTWWt7Z3dvSBKA2cjYzATUyAoZyvOWzv%2BZQC6dLuYHn4BSB7yyjUmm0UB%2B0jkilUGi0On0hhMZks1jsDli7i8PhAYN%2BABUrlxgbCojFnPEQktkmk9Jlsrl8npCsVNt8mn8AVDCaDWeygTDdAYjKZzFYbMlHM50d5fJ8kHiJpzPmwYMQoABJSDEShIdU1ACCYHwMFWhDAmiMMF03w%2BLwixJUsTJiUp6SyOTyCyZnxZNrZkN5IOx7B50JBcMFiJFKIU4tcnilWJlcoJfMVQh4KS4Os1wG1Gv1huNpsI5stDGtzWR0XtpOCTsUVJpbvpjLCIC9gd9gJDOjBf24vD4fj%2BaHQfhAwxAliNMEECHqIAAshAMoRBOmQK8JwB3COz%2BD1ADMACY3O8gA


src/select.tsx
```diff
@@ -66,7 +68,7 @@ const useStyles = createUseStyles((theme) => ({
         position: 'absolute',
         top: ({optionsComputedProps}) => optionsComputedProps.top,
         left: ({optionsComputedProps}) => optionsComputedProps.left,
-        width: ({optionsComputedProps}) => optionsComputedProps.width,
```
yceballost, line 69: should we limit the label width in this cases? I think is not very important, but...
image

yceballost (approved):

src/select.tsx
atabel: I think I prefer no limiting it, because it looks better when using the select component:
Screenshot_20210909_172636

yceballost: agree

anabelengp: I think the ticket also refers to this problem 

image

which may come from a missing `word-space: nowrap`.

atabel: fixed ;)

anabelengp (approved):

PR #339, opened 2021-09-15
fix(PasswordField): keep caret position when switching visibility

anabelengp:


src/password-field.tsx
```diff
@@ -39,6 +43,7 @@ const PasswordAdornment: React.FC<{
     };
     return (
         <IconButton
+            dataAttributes={{testid: 'visibility-button'}}
             aria-label={texts.togglePasswordVisibilityLabel}
```
atabel, line 47: instead of using a data-testid, I'd use the aria-label to find this button in tests

src/password-field.tsx
```diff
@@ -69,23 +74,29 @@ const PasswordField: React.FC<PasswordFieldProps> = ({
     ...rest
 }) => {
     const [isVisible, setIsVisible] = React.useState(false);
+    const caretPositionRef = React.useRef<number>();
```
atabel, line 77: does TS allow this? if it's a ref of type number, shouln't you give it a intial value of type number?

src/password-field.tsx
```diff
@@ -69,23 +74,29 @@ const PasswordField: React.FC<PasswordFieldProps> = ({
...
     const inputRef = React.useRef<HTMLInputElement>(null);
+    const input = inputRef.current;
+    if (input?.selectionStart) {
+        caretPositionRef.current = input.selectionStart;
+    }
```
atabel, line 82: mutating a ref inside render is not safe. It may work today, but it probably won't work in future React versions (with concurrent mode on)

src/password-field.tsx
anabelengp: TS doesn't complaint because:
```
function useRef<T = undefined>(): MutableRefObject<T | undefined>;
```
I'll initialize with 0 anyway

atabel (approved):

PR #340, opened 2021-09-17
feat(Form fields): use grey color for icons inside disabled fields

atabel:
jira: WEB-337


yceballost (approved):

src/text-field-base.tsx
```diff
@@ -358,6 +359,21 @@ const TextFieldBaseComponent = React.forwardRef<any, TextFieldBaseProps>(
...
+            if (!rest.disabled) {
+                return icon;
+            }
+            if (!isElement(icon)) {
+                return icon;
+            }
+            if (icon.props.color) {
+                // if a explicit color is set, use it
+                return icon;
+            }
```
anabelengp, line 373: Could it be useful to reduce the opacity or something like that in the previous two cases? in order to let the textfield to control the disabled state even when the icon is not an icon:
image

pladaria: I think this could be the best approach for all cases. If disabled, set an opacity to the icon container. No need to define different color for enabled/disabled states

pladaria (commented):
github forces me to leave a comment to set the commented state to de PR... :shrug:

src/text-field-base.tsx
yceballost: I like this approach too. I talked with Abel moooonths ago about to have a constant style to disabled things.

Could be a good opportunity to have a "disabled" state with opacity for any thing disabled in Mística.

WDYT @atabel ?

atabel: Do you think that'd work in all the cases @yceballost? including all brands and dark mode?

yceballost: well, exists some cases like buttons disabled that have their own style. Maybe we can try if buttons works with this definition too and remove disabled buttons constant too.

atabel: applied opacity instead of different color constant. But only for textfield icons for the moment, please take a look

anabelengp (approved):

package.json
```diff
@@ -161,6 +161,7 @@
         "react-autosuggest": "^9.4.3",
         "react-datetime": "^3.0.4",
         "react-focus-lock": "^2.5.2",
+        "react-is": "^17.0.2",
```
anabelengp, line 164: no longer needed, is it?

src/text-field-base.tsx
```diff
@@ -237,6 +239,7 @@ const useStyles = createUseStyles((theme) => ({
         alignItems: 'center',
         height: '100%',
         position: 'absolute',
+        opacity: ({disabled}) => (disabled ? 0.3 : 1),
     },
     prefix: {
```
pladaria, line 244: this could also have a disabled opacity

pladaria (approved):

yceballost (approved):
works fine!

package.json
atabel: removed

src/text-field-base.tsx
atabel: applied textDisabled color in that case

PR #343, opened 2021-09-21
feat(FeedbackScreen): update desktop mode

anabelengp:
This PR updates Feedback screens in desktop mode. `SuccessFeedbackHeader` will come in a future PR 

- [x] Update SuccessFeedbackScreen in desktop
- [x] fix Buttons order in FeedbackScreen for desktop and dialogs
- [x] Remove grid 6+6 in mobile (no longer needed since desktop mode is rendered independently)
- [x] Implement `imageFit` prop for image assets in SuccessFeedback
- [x] Is a padding-top needed in desktop mode? ADDED 64px, as they are Screens
- [x]  Fix bug Success background in mobile?? https://bit.ly/3Cs9M4y
      **After**
image


src/boxed.tsx
```diff
@@ -13,16 +13,17 @@ type StylesProps = {
...
+        backgroundColor: ({isInverseInside}: StylesProps) =>
             isInverseInside && !isDarkMode ? colors.backgroundBrand : colors.backgroundContainer,
+        backgroundClip: ({isInverseInside}: StylesProps) => (isInverseInside ? 'padding-box' : 'border-box'), // for transparent borders, no background is drawn beneath the border.
 
         border: ({isInverseOutside, isInverseInside}: StylesProps) => {
             if (isInverseOutside && !isInverseInside) {
                 return `1px solid ${colors.backgroundContainer}`;
             }
 
-            if (isInverseInside && !isDarkMode) {
```
anabelengp, line 24: I think the border must be transparent when isInverseInside in both light and dark mode, but I'm not sure

src/__stories__/feedback-screen-story.tsx
```diff
@@ -16,8 +16,16 @@ export default {
...
     },
+    argTypes: {
+        imageFit: {
+            options: ['fit', 'fill'],
+            control: {type: 'select'},
+        },
+    },
 };
 
+type SuccessFeedbackScreenArgs = {imageUrl: string | null; imageFit: 'fit' | 'fill'};
```
pladaria, line 27: perhaps `SuccessWithAssetArgs` would be a better name

anabelengp:
A comment related with no buttons bug: https://github.com/Telefonica/mistica-web/pull/344#discussion_r714719581

flow-defs/feedback-screen.js.flow
```diff
@@ -11,16 +12,23 @@ declare type FeedbackProps = {|
...
+
   hapticFeedback?: HapticFeedback,
   icon?: React.Node,
   animateText?: boolean,
 |};
 declare export var FeedbackScreen: React.ComponentType<FeedbackScreenProps>;
-declare export var SuccessFeedbackScreen: React.ComponentType<FeedbackProps>;
+declare export var SuccessFeedbackScreen: React.ComponentType<AssetFeedbackProps>;
 declare export var ErrorFeedbackScreen: React.ComponentType<FeedbackProps>;
 declare export var InfoFeedbackScreen: React.ComponentType<FeedbackProps>;
```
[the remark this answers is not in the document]
anabelengp: I understood that the large image is only available in SuccessFeedbackScreens in zeplin specs. I though that showing a large image when an error occurs was like laughing at the user :sweat_smile: 

@HectorSancho should the image be allowed in error and info screens??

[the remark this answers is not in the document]
HectorSancho: Yes, it only applies to success feedback to reinforce the positive message and the moment when a process has been completed.

src/__stories__/no-asset-feedback-screen-story.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
```
atabel, line 1: instead of no-asset-feedback-screen-story, I'd create a diferent story file for every component, one for InfoFeedbackScreen and one for ErrorFeedbackScreen

src/__stories__/success-feedback-screen-story.tsx
```diff
@@ -0,0 +1,40 @@
...
+            control: {type: 'select'},
+        },
+    },
+};
+
+type SuccessArgs = {imageUrl: string | null; imageFit: 'fit' | 'fill'; multipleParagraphs: boolean};
+
+const primaryButton = <ButtonPrimary onPress={() => {}}>Action1</ButtonPrimary>;
+
+const secondaryButton = <ButtonSecondary onPress={() => {}}>Action2</ButtonSecondary>;
```
atabel, line 22: inline these buttons in the story

src/__stories__/success-feedback-screen-story.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+const defaultTitle = "I'm the title",
+    defaultDescription = "I'm the description";
```
atabel, line 25: why default? it can't be changed. I'd just inline it

src/__stories__/feedback-screen-story.tsx
```diff
@@ -1,66 +1,24 @@
...
+import {FeedbackScreen} from '../feedback-screen';
 import {ButtonLink, ButtonPrimary, ButtonSecondary} from '../button';
 import Stack from '../stack';
 import {ThemeVariant, useIsInverseVariant} from '../theme-variant-context';
 import {useTheme} from '../hooks';
 import {Box, FixedToTop, Text3, Text2} from '..';
 
 export default {
-    title: 'Components/Screens/FeedbackScreen',
+    title: 'Components/Screens/FeedbackScreen/FeedbackScreen',
```
atabel, line 10: is the last `/FeedbackScreen` needed?

src/boxed.tsx
```diff
@@ -13,16 +13,17 @@ type StylesProps = {
 
 const useStyles = createUseStyles(({colors, isDarkMode}) => ({
     boxed: {
-        background: ({isInverseInside}: StylesProps) =>
+        backgroundColor: ({isInverseInside}: StylesProps) =>
             isInverseInside && !isDarkMode ? colors.backgroundBrand : colors.backgroundContainer,
+        backgroundClip: ({isInverseInside}: StylesProps) => (isInverseInside ? 'padding-box' : 'border-box'), // for transparent borders, no background is drawn beneath the border.
```
atabel, line 18: I thought these changes were to avoid this bottom border here:
image

Is it working fine?

flow-defs/feedback-screen.js.flow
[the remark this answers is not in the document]
atabel: ok

src/boxed.tsx
anabelengp: Well, in this case is only that the choosen image has a bottom pixel with transparency:
image
without this fix, it was worse :stuck_out_tongue:  https://bit.ly/3ALqlIm

src/__stories__/feedback-screen-story.tsx
anabelengp: no! I had some problems with the url during the sepparation of stories and I must have forgotten it. Thanks!

src/__stories__/success-feedback-screen-story.tsx
```diff
@@ -0,0 +1,40 @@
+import * as React from 'react';
+import {SuccessFeedbackScreen} from '../feedback-screen';
+import {ButtonPrimary, ButtonSecondary} from '../button';
+
+export default {
+    title: 'Components/Screens/FeedbackScreen',
```
pladaria, line 6: You can't have two different story files with the same tile. This causes problems in storybook

src/feedback-screen.tsx
```diff
@@ -148,14 +155,19 @@ interface FeedbackProps {
     description?: string | Array<string>;
     children?: React.ReactNode;
     unstable_inlineInDesktop?: boolean;
+    dataAttributes?: DataAttributes;
+}
+
+interface AssetFeedbackProps extends FeedbackProps {
+    imageUrl?: string;
+    imageFit?: 'fit' | 'fill'; // defaults to 'fill'
```
pladaria, line 163: Perhaps `'contain'` and `'cover'` would be better values for this prop. At least the name explains what you get in css.

anabelengp: I used the names that are already being used in HighlightedCards. I think DesignCore members use that naming.

src/__stories__/success-feedback-screen-story.tsx
anabelengp: I did have problems while sepparating the story files until I gave a name to each story component correctly. In the end it seems to compose a correct index
```
- Screens
  - FeedbackScreen
        - ErrorFeedbackScreen
        - FeedbackScreen
        - InfoFeedbackScreen
        - SuccessFeedbackScreen
```
The preview is showing that index. Do you think there can be any other problem??

src/feedback-screen.tsx
pladaria: ok then, let's keep naming

src/__stories__/success-feedback-screen-story.tsx
pladaria: In the past, defining using the same title in different story files caused issues (like stories disappearing while developing). If you don't get any problems, perhaps storybook already fixed this issue.

anabelengp: I've seen some issues with hotreloading. Moreover, stories with no argTypes definition takes the definition that is set in other stories. So I'll follow your recomendation and give each story a different path. Thanks!

Anyway, I get some ugly urls with this last change:

`http://localhost:6006/?path=/story/components-screens-feedbackscreen-successfeedbackscreen--success`
I cannot get rid of that double "success" id.
- successFeedbackScreen comes from the url path: `title: 'Components/Screens/FeedbackScreen/SuccessFeedbackScreen'`
- the last `success` id comes from the name of the exported component in the story: `export const Success: StoryComponent<SuccessArgs> =`

This is my best trial... I cannot get storybook paths working correctly along with pretty urls...

flow-defs/feedback-screen.js.flow
```diff
@@ -2,6 +2,7 @@
```
yceballost, line 2: Dark mode color is quite weird, is it expected @HectorSancho ?
image

Headers works different in this case

---

A 1px white appear around de componente
image

--

Would be nice to have the distribution of buttons horizontally in mobile, instead of vertically. (I suppose that this depends on button layout)

yceballost (changes requested):

flow-defs/feedback-screen.js.flow
anabelengp: The background color in DarkMode for desktop is the one that is set by Boxed component. Is that the one that seems weird to you?

anabelengp: the 1px transparent was intentional because of this visible pixel around the image that does not appear in the specs:
https://github.com/Telefonica/mistica-web/pull/343/commits/84fe0f67cc75ecbd71b5a224cdc3fb58f76dd7e6#diff-da49445e16cc5dcbb9554ebb7566091f80c2bf044af0fc5f7841a6e19278e816

We have implemented that Boxed component in inverse mode has a `border: 1px solid transparent` and `backgroundClip: padding-box`.

Do you think it is not correct?

anabelengp: the distribution of buttons in mobile is given by ButtonFixedFooterLayout. This occurrs in this component since it is fullscreen. I have another PR for the "header" case Hector designed: #345. In that case, the buttons are aligned if they have enough space.

yceballost: 1. We can use boxed with isInverse, no problem.
2. We talked about transparent pixel and we are going to remove it

src/boxed.tsx
atabel: can you find a better image for the screenshot?

src/boxed.tsx
```diff
@@ -13,16 +13,17 @@ type StylesProps = {
 
 const useStyles = createUseStyles(({colors, isDarkMode}) => ({
     boxed: {
-        background: ({isInverseInside}: StylesProps) =>
+        backgroundColor: ({isInverseInside}: StylesProps) =>
```
atabel, line 16: I think this change will break o2-classic, because colors.backgroundBrand is a gradient in o2 classic

atabel (approved):

src/boxed.tsx
anabelengp: ok! backgroundColor is not needed any more. I had to specify it in order to set a specific backgroundClip for the transparent border. But as we agreed to remove the border, those changes have been reverted.

src/boxed.tsx
anabelengp: done

flow-defs/feedback-screen.js.flow
anabelengp: done!

yceballost (approved):

pladaria (approved):
nice!

atabel (approved):

PR #344, opened 2021-09-22
feat(FixedFooterLayout): stop using Portal for the fixed part

pladaria:
WEB-339

Related PR: https://github.com/Telefonica/webapp/pull/904

* [x] Verify in multiple webapp flows
* [x] Verify in mobile browsers (iOS + Android)
* [x] Verify in movistar.es website (embedded features)
* [x] Verify that [this iOS bug](https://jira.tid.es/browse/APPS-7485) is not reproduced. 
    * Verified in iOS 13 and 14


src/fixed-footer-layout.tsx
```diff
@@ -4,26 +4,18 @@ import debounce from 'lodash/debounce';
...
 
 const getScrollDistanceToBottom = () => document.body.scrollHeight - window.innerHeight - window.scrollY;
 
-const waitForSwitchTransitionToStart = (fn: () => void) => {
-    const timeoutId = setTimeout(fn, 0);
-    return {
-        cancel: () => clearTimeout(timeoutId),
-    };
-};
-
```
pladaria, line 18: I don't remember why was this needed

src/fixed-footer-layout.tsx
```diff
@@ -35,6 +34,7 @@ const useStyles = createUseStyles((theme) => ({
     containerSmall: {
         paddingBottom: ({height}) => height,
         backgroundColor: ({containerBgColor}) => containerBgColor || theme.colors.background,
+        minHeight: ({topDistance}) => `calc(100vh - ${topDistance}px)`,
```
anabelengp, line 37: I think that with this change, I can get rid of this div  I had to mainain in order to fill the background color in feedbackScreen.
https://github.com/Telefonica/mistica-web/pull/343/commits/782f7c84726f0947b2e0c4fe5b4badd83dca13d9#diff-a73f3b991d7fd52028c47266987d26b2caef049af23f23ec6adffd8c12d9231aR276
We have to remember to check it after mergin any of the two PRs.

anabelengp (approved):

src/fixed-footer-layout.tsx
pladaria: This could be not correct :/ At least not with some usages I'm seeing in webapp...

atabel (approved):

Marcosld (approved):

src/fixed-footer-layout.tsx
```diff
@@ -105,44 +100,49 @@ const FixedFooterLayout: React.FC<Props> = ({
         );
 
         const transitionAwaiter = waitForSwitchTransitionToStart(checkDisplayShadow);
-        addPassiveEventListener(window, 'resize', checkDisplayShadow);
-        addPassiveEventListener(window, 'scroll', checkDisplayShadow);
+        const scrollEventTarget = getScrollEventTarget(scrollable);
+        addPassiveEventListener(scrollEventTarget, 'resize', checkDisplayShadow);
```
atabel, line 104: resize event listener should be on window, you cannot listen to dom elements resize without a ResizeObserver

atabel (approved):

src/fixed-footer-layout.tsx
```diff
@@ -4,10 +4,13 @@ import debounce from 'lodash/debounce';
...
-import {Portal} from './portal';
-import {addPassiveEventListener, removePassiveEventListener} from './utils/dom';
+import {addPassiveEventListener, getScrollableParentElement, removePassiveEventListener} from './utils/dom';
 
-const getScrollDistanceToBottom = () => document.body.scrollHeight - window.innerHeight - window.scrollY;
+const getScrollDistanceToBottom = (el: HTMLElement) => el.scrollHeight - el.scrollTop - el.clientHeight;
+
+const getScrollEventTarget = (el: HTMLElement) => (el === document.documentElement ? window : el);
+
+const hasScroll = (el: HTMLElement) => el.scrollHeight > el.clientHeight;
```
anabelengp, line 13: could these functions be utility functions in dom.tsx?

anabelengp (approved):

src/fixed-footer-layout.tsx
anabelengp: I will keep the div then

src/fixed-footer-layout.tsx
pladaria: sure! moved `getScrollDistanceToBottom` and `hasScroll`

src/fixed-footer-layout.tsx
pladaria: not an issue for now. I'm not aware of a fixed footer usage where the scrollable parent is not the html element. In that case the scrollEventTarget is window.

PR #351, opened 2021-10-05
feat(ButtonGroup): internal component for related-component buttons

anabelengp:


src/button-group.tsx
```diff
@@ -0,0 +1,45 @@
...
+    fullWidthLink: {
+        marginLeft: -buttonLinkPadding,
+        width: '100%',
+        marginTop: buttonLayoutSpacing,
+    },
+}));
+
+export interface ButtonGroupProps {
+    primaryButton?: React.ReactElement<ButtonProps, typeof ButtonPrimary>;
+    secondaryButton?: React.ReactElement<ButtonProps, typeof ButtonSecondary>;
```
anabelengp, line 24: I'm checking if I shouldn't restrict these types to primary and secondary buttons since I see some usages in which the secondary button is configured with a ButtonPrimary:
 https://github.com/Telefonica/webapp/blob/master/web/src/pages/tariff/o2-de/components/tariff.js#L37-L40

PR #355, opened 2021-10-15
new visual for icon story

yceballost:
To-Do
- [ ] Dark mode using storybook color sheme instead inverse checkbox
- [ ] Increase size from center XY axis
- [ ] When increase size, increase circle too
- [ ] Circle is broken
- [ ] Icon area is broken
- [ ] Color selection


sergivillar:
@yceballost  maybe this PR could be closed?

PR #364, opened 2021-11-04
chore(Build): Fix problem in flow types generation

pladaria:
Fixes problem after generating flow defs by moving the overrides logic to the `ts-to-flow` library.

See https://github.com/Telefonica/js-toolbox/pull/8


src/portal.tsx
```diff
@@ -47,7 +47,7 @@ export const usePortalNodes = (): PortalNodesContext => React.useContext(PortalN
  * This is useful for components that need position fixed, for example
  * for menus, alerts, overlays, etc.
  *
- * @see https://reactjs.org/docs/portals.html
+ * See https://reactjs.org/docs/portals.html
```
pladaria, line 50: for some reason flow-gen was producing a strange output due to this annotation

flow-defs/dialog.js.flow
```diff
@@ -22,7 +22,8 @@ declare type DialogRootState = {
   instanceNumber: number,
 };
 declare export default class DialogRoot
-  mixins React.Component<DialogRootProps, DialogRootState> {
+  mixins React.Component<DialogRootProps, DialogRootState>
+{
```
atabel, line 26: weird formating

pladaria: this is autogenerated :shrug:

atabel: do you know why it changed?

atabel (approved):

flow-defs/dialog.js.flow
pladaria: I guess because prettier was updated

anabelengp (approved):

PR #377, opened 2021-12-03
feat(ForwardRef): Support forward ref in: Row, BoxedRow, MediaCard, Boxed, Box, ButtonLink

pladaria:


src/test-utils/index.tsx
```diff
@@ -2,7 +2,7 @@ import {
     openPage,
     serverHostName,
     screen,
-    getGlobalPage,
```
pladaria, line 5: this export doesn't exist

types in non source files were not checked in CI

flow-defs/box.js.flow
```diff
@@ -36,5 +36,8 @@ declare type Props = {
   dataAttributes?: DataAttributes,
   "aria-label"?: string,
 };
-declare var Box: React.ComponentType<Props>;
+declare var Box: React.Component<{
+  ...Props,
+  ...React.RefAttributes<HTMLDivElement>,
```
pladaria, line 41: this type is invalid for flow, I need to update the ts-to-flow lib

src/__tests__/popover-test.tsx
```diff
@@ -6,7 +6,7 @@ import {makeTheme} from './test-utils';
 
 type Props = Omit<React.ComponentProps<typeof Popover>, 'children' | 'description' | 'target'>;
 
-const TestPopover: React.FC<Props> = (props) => (
+const TestPopover: React.FC<Props> = ({children, ...props}) => (
```
pladaria, line 9: Unrelated. Popover doesn't accept children

src/test-utils/index.tsx
```diff
@@ -2,7 +2,7 @@ import {
...
+    globalPage,
```
atabel, line 5: hmmm... I'd say I removed the `globalPage` export from the lib

pladaria: It exists in the d.ts file

pladaria: ok, I had an old library version

atabel (approved):

PR #380, opened 2021-12-09
feat(DecimalField): Add maxDecimals prop to limit the number of decimal places

pladaria:


src/__tests__/decimal-field-test.tsx
```diff
@@ -0,0 +1,34 @@
...
+import {render, screen, waitFor} from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import ThemeContextProvider from '../theme-context-provider';
+import {makeTheme} from './test-utils';
+
+test.each`
+    input            | maxDecimals  | expected
+    ${'123.4567890'} | ${undefined} | ${'123,4567890'}
+    ${'123.4567890'} | ${2}         | ${'123,45'}
+    ${'123.4567890'} | ${0}         | ${'1234567890'}
```
atabel, line 12: `maxDecimarls={0}` is like using `IntegerField`?

atabel (approved):

src/__tests__/decimal-field-test.tsx
pladaria: yes

PR #384, opened 2021-12-23
chore(playroom): fix props autocomplete

atabel:
fixes https://jira.tid.es/browse/WEB-410
fixes https://jira.tid.es/browse/WEB-214


src/package-version.tsx
```diff
@@ -1,3 +1,3 @@
-import {version} from '../package.json';
+import pkg from '../package.json';
```
atabel, line 1: new playroom version uses webpack 5 and was causing this warning: https://jira.tid.es/browse/WEB-214

size-stats.json
```diff
@@ -1,15 +1,15 @@
...
-        "js": "5536.45 KB",
+        "js": "5536.44 KB",
         "jsNoMisticaIcons": "664.20 KB"
     },
     "libOverhead": {
         "initial": "127.79 KB",
-        "withMistica": "267.93 KB",
-        "difference": "140.14 KB"
+        "withMistica": "273.31 KB",
+        "difference": "145.52 KB"
```
atabel, line 13: I think this 5KB increase is due to the change in `package-version.tsx`
The CRA we use to measure this lib overhead is using webpack 4 and it seems it can not tree-shake the package.json

pladaria: we could make a script that runs before build and writes the version into a file.

yceballost (approved):

src/package-version.tsx
pladaria: can we use babel macro here?

atabel: we don't use babel, so no babel macro. I don't know if there is something similar for swc.
We could write a file with the version as part of the release workflow

atabel: this should do the trick: [`943bd12` (#384)](https://github.com/Telefonica/mistica-web/pull/384/commits/943bd12b2b5e7ce77dbb74817a954d2de599783c)

pladaria (approved):

PR #385, opened 2021-12-27
chore(eslint): update to v8

atabel:


.eslintrc.yml
```diff
@@ -1,2 +1,5 @@
 extends:
     - '@telefonica/eslint-config'
+overrides:
+    - files: ['*.js']
+      parser: espree
```
atabel, line 5: use default (espree) parser instead of babel parser for js files, because we don't have a babel config in this repo, because we don't use babel, but swc

pladaria (approved):

PR #391, opened 2022-01-13
feat(Video): Video component and Video support in Media Cards

pladaria:


flow-defs/image.js.flow
```diff
@@ -1,8 +1,10 @@
 // @flow
 
 import * as React from "react";
+declare export var useDisableBorderRadius: () => boolean;
+declare export var DisableBorderRadiusProvider: React.ComponentType<{}>;
 declare type AspectRatio = "1:1" | "16:9" | "7:10";
-declare type Props = {
+export type ImageProps = {
  url: string,
```
yceballost, line 8: would be nice to have the same way to call things between image and video? For example, in image componente we call "url" but in video we call it "src"

image

On the other hand, make sense to add aspect ratio props like image in video? To keep both component similar

pladaria: I prefer src, because you're defining a "source". Changing the image prop would be a breaking change.

For videos the aspect ratio is not needed (at least for now) because video dimensions are known

atabel: +1 to src and deprecate url

src/video.tsx
```diff
@@ -0,0 +1,91 @@
...
+    height?: number | string;
+    /** accepts multiple sources */
+    src: string | Array<string> | VideoSource | Array<VideoSource>;
+    /** defaults to true */
+    loop?: boolean;
+    /** defaults to true */
+    muted?: boolean;
+    /** defaults to true */
+    autoPlay?: boolean;
+    poster?: string;
```
atabel, line 36: thumbnail?

atabel (approved):

playroom/snippets.tsx
```diff
@@ -200,7 +200,7 @@ const listSnippets: Array<Snippet> = [
                 onPress={() => {}}
             />
```
yceballost, line 201: Can we add <Image> and <Video> as a playroom snippet component

src/__stories__/media-card-story.tsx
```diff
@@ -180,10 +182,15 @@ const useCarouselStyles = createUseStyles((theme) => ({
     },
 }));
```
yceballost, line 184: Would be nice to have a media card with image and video in story

src/__stories__/video-story.tsx
```diff
@@ -0,0 +1,89 @@
...
+                    >
+                        Play
+                    </ButtonPrimary>
+                    <ButtonPrimary
+                        small
+                        onPress={() => {
+                            videoRef.current?.pause();
+                        }}
+                    >
+                        Stop
```
yceballost, line 47: A little detail, but maybe is better "pause" instead of "stop"

src/__stories__/video-story.tsx
```diff
@@ -0,0 +1,89 @@
...
+                        Play
+                    </ButtonPrimary>
+                    <ButtonPrimary
+                        small
+                        onPress={() => {
+                            videoRefWithPoster.current?.pause();
+                            videoRefWithPoster.current?.load(); // to force showing the poster again
+                        }}
+                    >
+                        Stop
```
yceballost, line 81: This stop works different than the second example, it is expected? This one works as stop and the second example works as pause

playroom/snippets.tsx
yceballost: Also change Media Card snippet because is using a deprecated way

src/video.tsx
yceballost: cover?

pladaria: "poster" is the video attribute name

src/__stories__/video-story.tsx
pladaria: it is expected, if you see the story implementation, there's some extra code to force resetting the video

src/image.tsx
```diff
@@ -1,47 +1,84 @@
...
+
 const useStyles = createUseStyles(() => ({
     image: {
-        borderRadius: 4,
+        borderRadius: ({noBorderRadius}) => (noBorderRadius ? 0 : 4),
         display: 'block',
         objectFit: 'cover',
         maxWidth: '100%',
         maxHeight: '100%',
+        aspectRatio: ({aspectRatio}) => aspectRatio ?? 'unset',
```
pladaria, line 24: Note that this CSS property is only supported by modern browsers.

For browsers where this property is not supported (and no height or width are given), the media will be displayed in the original aspect ratio, which is good enough to me

src/video.tsx
```diff
@@ -0,0 +1,126 @@
...
+import {createUseStyles} from './jss';
+import {combineRefs} from './utils/common';
+
+export type AspectRatio = '1:1' | '16:9' | '12:5';
+
+export const RATIO = {
+    '1:1': 1,
+    '16:9': 16 / 9,
+    '12:5': 12 / 5,
+};
```
pladaria, line 12: I considered, instead of a fixed set of ratios, to accept a number. Anyway, extending this is easy

yceballost (approved):

src/image.tsx
```diff
@@ -58,19 +61,24 @@ type DeprecatedImageProps = {
...
     if (props.width !== undefined) {
         height = props.width / RATIO[aspectRatio];
     } else if (props.height !== undefined) {
         width = props.height * RATIO[aspectRatio];
+    } else {
+        width = '100%';
     }
 
-    return <img src={url} className={classes.image} alt={alt} width={width} height={height} />;
+    return <img src={url} className={classnames(classes.image)} alt={alt} width={width} height={height} />;
```
atabel, line 81: classnames call not needed here

atabel (approved):

RubenMorPov (approved):

src/image.tsx
pladaria: woops, leftovers

kydorn (approved):

PR #393, opened 2022-01-17
feat(Feedback): show info/error icons in vivo

RubenMorPov:
Now Vivo's info and error feedback screens will use the same icons as the others.


yceballost (approved):

atabel (approved):
please, review the CI build fail. It seems like you committed without prettier.

PR #394, opened 2022-01-18
feat(TextFieldBase): expose internal TextFieldBase component for advanced usage

sergivillar:


src/text-field-base.tsx
```diff
@@ -565,9 +565,9 @@ const TextFieldBase = React.forwardRef<any, TextFieldBaseProps>(
...
             </React.Suspense>
         ) : (
-            <TextFieldBaseComponent {...props} id={id} ref={ref} />
+            <TextFieldBase {...props} id={id} ref={ref} />
         );
     }
 );
 
-export default TextFieldBase;
+export default TextFieldBaseAutosuggest;
```
atabel, line 573: it's weird that the file is called text-field-base and the default export in this file is not the TextFieldBase, but there is also a named export with that name :/

sergivillar: totally true, I think the same, but.... any better naming? for me it make sense the names in the file, I could rename the file to 'text-field-base-autosuggest' but for me that's a bit weird too....

pladaria: I'd make both a named export (no default)

sergivillar: changed, take a look

atabel (approved):

PR #396, opened 2022-01-19
fix(ButtonLayout): missed button press in button layout when gaining focus

pladaria:
Related issue https://jira.tid.es/browse/WEB-432
See attached video to know what the problem is. Can also be reproduced using the dialog story in storybook


src/button-layout.tsx
```diff
@@ -161,8 +161,12 @@ const ButtonLayout: React.FC<ButtonLayoutProps> = ({
     }, [classes.link, isMeasuring]);
 
     const calcLayout = React.useCallback(() => {
-        updateButtonWidth(0);
-        updateIsMeasuring(true);
+        // These updates are executed inside a setTimeout to to avoid an immediate re-render on a visibility change
+        // This issue can be reproduced with Chrome. The button click is missed if pressed having the focus on the devtools
```
atabel, line 165: why does a visibility change trigger a re-render?

pladaria: see line 183. Those listeners are triggering a state change

andyindahouse (approved):
👌

RubenMorPov (approved):

atabel (approved):

PR #400, opened 2022-01-23
chore(Playroom): added hide prop to Preview Tools

yceballost:
I need this prop to show examples in documentation without brand tab but forcing other properties like `forceDesktop` or always light mode for example


playroom/components.tsx
```diff
@@ -336,6 +338,14 @@ export const PreviewTools: React.FC<PreviewToolsProps> = ({
                 )}
             </>
         );
+    }
+    if (hide) {
+        return (
+            <>
+                {children}
+                <div style={{display: 'none'}} />
```
atabel, line 346: I don't get this. Why the display none? isnt `return children` enough?

playroom/components.tsx
```diff
@@ -319,7 +321,7 @@ export const PreviewTools: React.FC<PreviewToolsProps> = ({
         </ThemeContextProvider>
     );
 
-    if (floating) {
+    if (floating && !hide) {
```
atabel, line 324: instead of adding the `&& hide` here, I'd add an early return before this

playroom/components.tsx
yceballost: true, I never used return xD

playroom/components.tsx
yceballost: perfect!

pladaria (approved):

atabel (approved):

PR #401, opened 2022-01-25
feat(Touchable): WEB-435 add hrefDecorator to theme

pladaria:


src/touchable.tsx
```diff
@@ -129,7 +129,7 @@ export type Props = PropsHref | PropsTo | PropsOnPress | PropsMaybeHref | PropsM
 
 const Touchable = React.forwardRef<HTMLDivElement | HTMLAnchorElement | HTMLButtonElement, Props>(
     (props, ref) => {
-        const {texts, analytics, platformOverrides, Link} = useTheme();
+        const {texts, analytics, platformOverrides, Link, hrefDecorator} = useTheme();
```
atabel, line 132: can `hrefDecorator` be a hook? It could be useful if we need to read some context in webapp.

pladaria: ok, I see

atabel (approved):

RubenMorPov (approved):

PR #404, opened 2022-01-27
fix(ButtonLayout): unstable in acceptance tests

pladaria:


src/button-layout.tsx
```diff
@@ -149,8 +151,10 @@ const ButtonLayout: React.FC<ButtonLayoutProps> = ({
                         el.classList.contains(classes.link) ? 0 : (el as HTMLElement).offsetWidth + 1
                     );
                     const maxChildWidth = Math.ceil(Math.max(...childrenWidths, BUTTON_MIN_WIDTH));
-                    updateButtonWidth(maxChildWidth);
-                    updateIsMeasuring(false);
+                    ReactDom.unstable_batchedUpdates(() => {
+                        updateButtonWidth(maxChildWidth);
+                        updateIsMeasuring(false);
+                    });
```
atabel, line 157: instead of batching updates, you could have a single state object with 2 keys

pladaria: done

atabel (approved):

PR #407, opened 2022-02-01
[Proposal] Row with renderRight and centerY

SergioMorchon:
Useful information to render properly the right side of the row.

## Issue

Custom text on the `right` side of a `Row` isn't properly aligned with the other texts on the main content.
Look at the "12,00 €" red texts here:
image

## Proposal

Expose the `centerY` flag so consumers can properly align their elements on the right side through the `renderRight` render callback.

| Single text | Multiple texts |
| - | - |
| single-text | multiple-texts |

### DIFF
--- a/src/list.tsx
+++ b/src/list.tsx
@@ -127,7 +127,7 @@ interface CommonProps {
 interface ContentProps extends CommonProps {
     isClickable?: boolean;
     type?: 'chevron' | 'basic' | 'custom' | 'control';
-    right?: React.ReactNode;
+    renderRight?: ({centerY}: {centerY: boolean}) => React.ReactNode;
     /** This id is to link the title with the related control */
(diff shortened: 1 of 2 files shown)


src/list.tsx
```diff
@@ -127,7 +127,7 @@ interface CommonProps {
 interface ContentProps extends CommonProps {
     isClickable?: boolean;
     type?: 'chevron' | 'basic' | 'custom' | 'control';
-    right?: React.ReactNode;
+    renderRight?: ({centerY}: {centerY: boolean}) => React.ReactNode;
```
SergioMorchon, line 130: This is the significant change

pladaria:
We're going to evaluate this: https://jira.tid.es/browse/WEB-447

PR #413, opened 2022-02-10
feat(types): better Flow types for ReactElement

atabel:
With this change, if a component prop is typed as `RendersElement<typeof Foo>`, it will accept `<Foo/>` elements but also elements of other components which end up rendering a `<Foo/>` element.

Some examples in this gist: https://gist.github.com/atabel/badefe5d16ad77d5310d58c4be3f2c6a

This only works in Flow for the moment, not in `TS`. I've still not found a way of TS to even work with the basic case (forcing the element to be of type `<Foo/>`), we have a different task to investigate further (WEB-456).


src/utils/renders-element.tsx
```diff
@@ -0,0 +1,4 @@
+import * as React from 'react';
```
pladaria, line 1: I'd move this to utils/types.tsx

atabel: it's reexported there, but I need this in a separated file to be able to have a flow override

pladaria (approved):

pladaria (approved):

PR #414, opened 2022-02-10
feat(Callout): Changed element distribution so link has more space

RubenMorPov:
[WEB-433](https://jira.tid.es/browse/WEB-433) Changed element distribution so link has more space


yceballost (approved):

src/callout.tsx
```diff
@@ -72,14 +73,32 @@ const Callout: React.FC<Props> = ({
...
+                            <Stack space={4}>
+                                <Text3 as="h2" regular>
+                                    {title}
+                                </Text3>
+                                <Text2 as="p" regular color={colors.textSecondary}>
+                                    {description}
+                                </Text2>
+                            </Stack>
+                            {onClose && (
+                                <div style={{textAlign: 'right'}}>
```
atabel, line 86: I think this style is not needed if you use `space="between"` in the `<Inline`

src/callout.tsx
```diff
@@ -72,14 +73,32 @@ const Callout: React.FC<Props> = ({
...
-                        <>
-                            <Text3 as="h2" regular>
-                                {title}
-                            </Text3>
-                            <Text2 as="p" regular color={colors.textSecondary}>
-                                {description}
-                            </Text2>
-                        </>
+                        <Inline fullWidth alignItems="flex-start" space={16}>
+                            <Stack space={4}>
```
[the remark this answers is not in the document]
RubenMorPov: I'm trying to put 4px between title and description, but that 4px should remain when there's no title, so the description is aligned with the icons. This solution is no working tho (the space only works when is more than one element), I'll remove it and try other things.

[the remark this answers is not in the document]
atabel: according to the spec, there is no space betwen title and description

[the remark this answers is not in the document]
yceballost: Specs visual guide for 4px was missing, but I added! Between Title and Description have to be 4px

src/callout.tsx
```diff
@@ -9,6 +9,7 @@ import IconButton from './icon-button';
 import classNames from 'classnames';
 import {ButtonLink, ButtonPrimary, ButtonSecondary} from './button';
 import ButtonGroup from './button-group';
+import {Inline} from '.';
 
 import type {ButtonProps, ButtonLinkProps} from './button';
```
yceballost, line 15: this screenshot hasn't a good alignment, right?
https://github.com/Telefonica/mistica-web/pull/414/files#diff-359bb804a1c6f7b7f2063c2a84611a054a27c892c8908dbf0efeb8869a22ce47

src/callout.tsx
```diff
@@ -72,14 +73,31 @@ const Callout: React.FC<Props> = ({
...
-                        </>
+                        <Inline fullWidth alignItems="flex-start" space="between">
+                            <Stack space={0}>
+                                <Text3 as="h2" regular>
+                                    {title}
+                                </Text3>
+                                <div style={{height: '4px'}}></div>
+                                <Text2 as="p" regular color={colors.textSecondary}>
+                                    {description}
+                                </Text2>
```
pladaria, line 84: instead of using a div to separate items, se the Stack space to 4

src/callout.tsx
```diff
@@ -72,14 +73,31 @@ const Callout: React.FC<Props> = ({
...
+                            </Stack>
+                            {onClose && (
+                                <IconButton
+                                    size={40}
+                                    style={{
+                                        margin: '-8px -12px -8px 0',
```
pladaria, line 90: better use `position: relative, top: xxx, left: xxx` to position the element

src/callout.tsx
```diff
@@ -72,14 +73,31 @@ const Callout: React.FC<Props> = ({
...
+                                        alignItems: 'center',
+                                        justifyContent: 'center',
+                                    }}
+                                    onPress={onClose}
+                                    aria-label={texts.closeButtonLabel}
+                                >
+                                    <IconCloseRegular color={colors.neutralHigh} />
```
pladaria, line 97: I'd set the size explicitly to 24

src/callout.tsx
```diff
@@ -72,14 +73,31 @@ const Callout: React.FC<Props> = ({
                 {icon && <div className={classes.icon}>{icon}</div>}
```
pladaria, line 73: perhaps you could replace this div with a `<Box paddingRight={16}>` and remove the icon class

src/callout.tsx
RubenMorPov: I think we're missing something, making this change breaks the icon alignment. Not sure why, but with the margin is fine but with this change is not.

atabel (commented):
Is this change expected? I'd say no
image

atabel (approved):

pladaria (approved):

yceballost:
> Is this change expected? I'd say no image

was fixed?

PR #419, opened 2022-02-16
fix(ThemeContextProvider): Support SSR + Strict Mode

pladaria:


flow-defs/theme-context-provider.js.flow
```diff
@@ -5,6 +5,17 @@ import type { ThemeConfig } from "./theme";
 declare export var useIsOsDarkModeEnabled: () => boolean;
 declare type Props = {
   theme: ThemeConfig,
+
+  /**
+   * You should use this prop if you use Strict Mode and Server Side Rendering together.
+   * This identifier will be used to generate unique class names for each instance of ThemeContextProvider.
+   * If no identifier is provided, this will fallback to an aunto-incremented id, which will cause
```
sergivillar, line 12: typo `auto`

sergivillar (approved):

src/theme-context-provider.tsx
```diff
@@ -70,16 +80,21 @@ const useDefaultHrefDecorator = () => {
...
 
-const ThemeContextProvider: React.FC<Props> = ({theme, children}) => {
+const ThemeContextProvider: React.FC<Props> = ({theme, children, providerId}) => {
+    const [instanceId] = React.useState(() => {
+        if (providerId) {
+            return providerId;
+        } else {
+            return isServerSide() ? 0 : nextJssInstanceId++;
+        }
+    });
```
atabel, line 90: do you need the useState? why don't you do this logic inside the useMemo?

atabel (approved):

src/theme-context-provider.tsx
pladaria: I don't want the useMemo to depend on the prop. Once the value is set I don't want to allow it to change. A better option would be to use a ref but useRef doesn't accept a callback

PR #422, opened 2022-02-17
feat(Rows): auto center right node when needed and allow custom render function

RubenMorPov:
[WEB-447](https://jira.tid.es/browse/WEB-447) - Added the shouldCenter flag on right prop. Now it centers properly. Also added a new option on the list story to check it.


src/__stories__/list-story.tsx
```diff
@@ -103,6 +106,19 @@ const Template: StoryComponent<Args & {boxed?: boolean}> = ({
...
+            case 'custom element with text':
+                controlProps = {
+                    right: (
+                        <div>
+                            <Inline space={0}>
+                                <Text3 color={colors.error} medium as="p">
+                                    12,00 €
+                                </Text3>
+                            </Inline>
+                        </div>
```
SergioMorchon: The real use case is that some rows have a navigation, that may show a chevron. While also have a money amount text that also may be in a customized color to show an extra charge.

```suggestion
                        <Inline space={0}>
                            <Text3 color={colors.error} medium as="p">
                                12,00 €
                            </Text3>
                            <IconChevron direction="right" color={colors.neutralMedium} />
                        </Inline>
```

This produces something like the following:
image

Before, the chevron was _always_ centered, but now we will move it to the top. I'm just trying to be sure about that change.

You can find this construction by searching for `<IconChevron` at webapp, being most of them either:
* Surrounded by a `{display: flex; align-items: center; height: 100%}` style, or
* With a shamy and manually hardcoded padding top, tied to the presence/absecnce ot the row subtitle.

SergioMorchon (approved):

src/__stories__/list-story.tsx
```diff
@@ -12,6 +12,9 @@ import {
     IconLikeFilled,
```
pladaria, line 12: Can you please add a screenshot test for your changes?

src/__stories__/list-story.tsx
yceballost: yes, chevron has to be always aligned to the center

SergioMorchon: Using `to`, `href` or `onPress` is an implementation detail.
But:
* Because it _navigates_, it has a chevron. and
* Because we need to add a red text at the right, we use the `right` prop, containing both.

And:
* Mística has no control about what's _inside_ `right`, and
* The consumer (filling the `right`) has no control over the _outside_ vertical alignement, then
* I see see way to conciliate both.

That's why I proposed to provide the `centerY` prop as a kind of _hint_ from Mística to the consumer, at https://github.com/Telefonica/mistica-web/pull/407

SergioMorchon: Another idea could be to render the Mïstica's chevron if required (based on the current criteria) even if the `right` prop exists

yceballost: I would like to propose something. Could we manage always the chevron with onPress?
Screen Shot 2022-02-21 at 23 12 14

In this way, we avoid Sergio's examples and we will have control of this chevron If in the future we want to modify or whatever

src/__stories__/list-story.tsx
```diff
@@ -103,6 +106,20 @@ const Template: StoryComponent<Args & {boxed?: boolean}> = ({
                     ),
                 };
                 break;
+            case 'custom element with text':
```
SergioMorchon, line 109: Nice, I only miss an update at line 80 (see https://github.com/Telefonica/mistica-web/pull/407/files#diff-b0914293aa384640f5a0b98c3be2713a6dbd19d63fdcdc8a8723cb3eab4d931d), that uses `right` and should be `renderRight: null`. Also the comment.

src/list.tsx
```diff
@@ -242,7 +218,30 @@ const Content: React.FC<ContentProps> = ({
...
+            {type === 'custom' && navigable === true && (
+                <div className={classNames(classes.right, {[classes.disabled]: disabled})}>
+                    <div style={centerY ? {display: 'flex', alignItems: 'center', height: '100%'} : {}}>
+                        <IconChevron
+                            color={isInverse ? colors.inverse : colors.neutralMedium}
+                            direction="right"
+                        />
+                    </div>
+                </div>
+            )}
```
SergioMorchon, line 244: This should be equal, while narrowing the conditional renderings:

```suggestion
            {type === 'custom' && (
                <>
                    <div className={classNames(classes.right, {[classes.disabled]: disabled})}>
                        {renderRight?.({centerY})}
                    </div>
                    {navigable && (
                        <div className={classNames(classes.right, {[classes.disabled]: disabled})}>
                            <div
                                style={centerY ? {display: 'flex', alignItems: 'center', height: '100%'} : {}}
                            >
                                <IconChevron
                                    color={isInverse ? colors.inverse : colors.neutralMedium}
                                    direction="right"
                                />
                            </div>
                        </div>
                    )}
                </>
            )}
```

SergioMorchon (approved):

src/list.tsx
```diff
@@ -122,17 +122,19 @@ interface CommonProps {
     extra?: React.ReactNode;
     dataAttributes?: DataAttributes;
     disabled?: boolean;
+    navigable?: boolean;
```
atabel, line 125: so, the idea of this prop is to only use it in conjunction with `renderRight` when we want the chevron to be rendered at the same time than the custom right element?

src/list.tsx
```diff
@@ -122,17 +122,19 @@ interface CommonProps {
...
 }
 
 interface ContentProps extends CommonProps {
     isClickable?: boolean;
     type?: 'chevron' | 'basic' | 'custom' | 'control';
-    right?: React.ReactNode;
+    renderRight?: ({centerY}: {centerY: boolean}) => React.ReactNode;
```
atabel, line 131: this is a breaking change, we should try to avoid it.
A posible solution would be to make `right` polymorphic, allow to receive `React.ReactNode | ({centerY}: {centerY: boolean})  => React.ReactNode`

src/list.tsx
RubenMorPov: Yep! That's right.

atabel: I'd add a comment explaining this. If you use jsdoc, vscode will show the comment to users of this component

SergioMorchon (commented):
I'm seeing that in some screenshot tests (that are passing) the _custom element_ dissappeared. For example, src/__screenshot_tests__/__image_snapshots__/list-screenshot-test-tsx-boxed-row-list-desktop-custom-element-extra-false-badge-true-1-snap.png

Can you check if that's expected?

Also the `badge` seems to be decoupled from the `centerY` criteria, even though it is part of the right side of the row (probably out of this CR, but just something I think that should be fixed at some point).

src/list.tsx
```diff
@@ -242,7 +218,34 @@ const Content: React.FC<ContentProps> = ({
...
+                    {navigable && (
+                        <div className={classNames(classes.right, {[classes.disabled]: disabled})}>
+                            <div
+                                style={centerY ? {display: 'flex', alignItems: 'center', height: '100%'} : {}}
+                            >
+                                <IconChevron
+                                    color={isInverse ? colors.inverse : colors.neutralMedium}
+                                    direction="right"
+                                />
+                            </div>
```
SergioMorchon, line 244: IIRC, the chevron should be *always* centered vertically, is that ok @yceballost?
If that's the case, then this should be inconditionally centered in the vertical:
```suggestion
                            <div
                                style={{display: 'flex', alignItems: 'center', height: '100%'}}
                            >
                                <IconChevron
                                    color={isInverse ? colors.inverse : colors.neutralMedium}
                                    direction="right"
                                />
                            </div>
```

SergioMorchon (approved):

src/list.tsx
```diff
@@ -122,17 +122,19 @@ interface CommonProps {
...
     dataAttributes?: DataAttributes;
     disabled?: boolean;
+    navigable?: boolean; // enables chevron on custom elements with text
 }
 
 interface ContentProps extends CommonProps {
     isClickable?: boolean;
     type?: 'chevron' | 'basic' | 'custom' | 'control';
-    right?: React.ReactNode;
+    right?: ({centerY}: {centerY: boolean}) => React.ReactNode | React.ReactNode;
```
atabel, line 131: I miss some parenthesis here, otherwise this wont work
```suggestion
    right?: (({centerY}: {centerY: boolean}) => React.ReactNode) | React.ReactNode;
```

src/__stories__/list-story.tsx
```diff
@@ -96,17 +99,31 @@ const Template: StoryComponent<Args & {boxed?: boolean}> = ({
                 break;
             case 'custom element':
                 controlProps = {
-                    right: (
+                    right: () => (
                         <div style={{display: 'flex', alignItems: 'center', height: '100%'}}>
                             <div style={{width: 32, height: 32, borderRadius: '50%', background: 'pink'}} />
                         </div>
```
pladaria, line 105: Please, use the callback parameter in the example

RubenMorPov: I kept this example as a React.Node instead the callback function, so we have both examples, with and without the new right callback with centerY.

src/list.tsx
yceballost: true, chevron always has to be placed centered

SergioMorchon (approved):
:rocket:

atabel (approved):

src/list.tsx
```diff
@@ -122,17 +122,24 @@ interface CommonProps {
     extra?: React.ReactNode;
     dataAttributes?: DataAttributes;
     disabled?: boolean;
+    navigable?: boolean; // enables chevron on custom elements with text
```
pladaria, line 125: if you rename this prop to "withChevron" you can remove the comment and it is self explanatory

RubenMorPov: Done!

yceballost (approved):
I didn't see buggy things 👌

PR #426, opened 2022-02-23
fix(select): issue with cropped text

RubenMorPov:
[WEB-441](https://jira.tid.es/browse/WEB-441) Changed line height and label height. Now text is not cropped when an option is selected.


src/select.tsx
```diff
@@ -52,7 +52,8 @@ const useStyles = createUseStyles((theme) => ({
         left: 12 + 1, // 12 for select paddingLeft and +1 for border
         right: 48 + 1, // 48 for icon and +1 for border
         overflow: 'hidden',
-        top: ({label}) => (label ? 27 : 17),
+        top: ({label}) => (label ? 26 : 16),
```
yceballost, line 55: It's better than before but I can still see that the text is clipped :/ 

On the other hand, space between label and input text is too small compare with the standard input
image
image

yceballost (changes requested):

src/select.tsx
```diff
@@ -52,7 +52,8 @@ const useStyles = createUseStyles((theme) => ({
...
+        lineHeight: '18px',
```
yceballost, line 56:
```suggestion
        lineHeight: '20px',
```
Rolling back your changes and change this line I think solve the problem and keep the same the distances between forms text and selects

image

yceballost (changes requested):

src/select.tsx
yceballost, line 55:
```suggestion
        top: ({label}) => (label ? 27 : 17),
```

src/select.tsx
```diff
@@ -52,7 +52,8 @@ const useStyles = createUseStyles((theme) => ({
         left: 12 + 1, // 12 for select paddingLeft and +1 for border
         right: 48 + 1, // 48 for icon and +1 for border
         overflow: 'hidden',
```
yceballost, line 54: I found problems using big sur and desktop views and I think is caused by https://jira.tid.es/browse/WEB-251 (that I don't know why @pladaria close this ticket xD)
https://tinyurl.com/yc4l7pe3

image

pladaria: I closed it because this seems to be something specific to Safari rendering, we can't do much there. 

I reopened it  but I think we can't do much there

pladaria (approved):

yceballost (approved):
Thank you! :)

PR #428, opened 2022-02-24
fix(Row): checkbox onChange always returning true

RubenMorPov:
[WEB-460](https://jira.tid.es/browse/WEB-460) Fixed onChange parameter, now passes "isChecked".


src/list.tsx
```diff
@@ -374,7 +374,7 @@ const useControlState = ({
             setIsChecked(!isChecked);
         }
         if (onChange) {
-            onChange(!value);
+            onChange(!isChecked);
```
atabel, line 377: I think this should be:
```ts
onChange(isControlledByParent ? !value : !isChecked)
```
Please, add some tests to check both cases (controlled by parent and uncontrolled)

atabel (changes requested):

src/list.tsx
pladaria: in this case some simple unit tests should be enough

src/__tests__/list-test.tsx
```diff
@@ -37,10 +37,16 @@ test('Row as a button', () => {
 });
 
 test('Row with switch', () => {
+    const onChangeFn = (isChecked?: boolean) => {
+        return isChecked;
+    };
+
+    const spyOnChange = jest.fn(onChangeFn);
```
pladaria, line 44: you can write this as:
```js
const onChangeSpy = jest.fn((isChecked?: boolean) => {}); // The callback doesn't need to return anything
```

src/__tests__/list-test.tsx
```diff
@@ -37,10 +37,16 @@ test('Row as a button', () => {
 });
 
 test('Row with switch', () => {
+    const onChangeFn = (isChecked?: boolean) => {
+        return isChecked;
```
pladaria, line 41: no need to return anything

pladaria (approved):

src/__tests__/list-test.tsx
```diff
@@ -37,10 +37,16 @@ test('Row as a button', () => {
 });
 
 test('Row with switch', () => {
```
pladaria, line 39: Looks like the test is not passing
image

atabel (approved):

PR #430, opened 2022-02-28
feat(Carousel): autoplay only move when carousel is visible

atabel:


src/carousel.tsx
```diff
@@ -138,6 +146,7 @@ const useStyles = createUseStyles((theme) => ({
         },
     },
     item: {
+        scrollSnapStop: isAndroid(theme.platformOverrides) ? 'always' : 'normal',
```
atabel, line 149: This is a small fix for android devices, where the scroll inertia is too high and it's difficult to scroll only one slide without forcing scroll snap to stop.

The default behavior works fine in desktop and in iOS so I've decided to only apply this to Android devices.

src/utils/document-visibility.tsx
```diff
@@ -0,0 +1,52 @@
+// @flow
+import * as React from 'react';
```
atabel, line 2: extracted from webapp

pladaria (approved):

kydorn (approved):

PR #433, opened 2022-03-01
WEB-457 - Changes on image component

RubenMorPov:
[WEB-457](https://jira.tid.es/browse/WEB-457) Changes on Image component

- [x] Accept string as width/height.
- [ ] Prop to force borderRadius state (enable/disable). By default it reads a context to know if the border radius should be used.
- [x] Allow any aspect ratio, perhaps adding `| number` to the AspectRatio type (so you can pass an aspect ratio as 16/9).

### DIFF
--- a/src/image.tsx
+++ b/src/image.tsx
@@ -37,13 +37,14 @@ export const RATIO = {
 };
 
 export type ImageProps = {
+    noBorderRadius?: boolean;
     src: string;
     url?: undefined;
     /** defaults to 100% when no width and no height are given */
-    width?: number;
-    height?: number;
+    width?: string | number;
+    height?: string | number;
     /** defaults to 1:1, if both width and height are given, aspectRatio is ignored */
-    aspectRatio?: AspectRatio;
+    aspectRatio: AspectRatio | number;
     /** defaults to empty string */
     alt?: string;
     children?: void;
@@ -67,9 +68,10 @@ type DeprecatedImageProps = {
 const Image = React.forwardRef<HTMLImageElement, ImageProps | DeprecatedImageProps>(
     ({aspectRatio = '1:1', alt = '', dataAttributes, ...props}, ref) => {
         const noBorderRadius = useDisableBorderRadius();
+        const ratio = typeof aspectRatio === 'number' ? aspectRatio : RATIO[aspectRatio];
         const classes = useStyles({
             noBorderRadius,
-            aspectRatio: !props.width && !props.height ? RATIO[aspectRatio] : undefined,
+            aspectRatio: !props.width && !props.height ? ratio : undefined,
         });
         const url = props.src || props.url;
 
@@ -80,9 +82,9 @@ const Image = React.forwardRef<HTMLImageElement, ImageProps | DeprecatedImagePro
             width = props.width;
             height = props.height;
         } else if (props.width !== undefined) {
-            height = props.width / RATIO[aspectRatio];
+            height = typeof props.width === 'number' ? props.width / ratio : props.width;
         } else if (props.height !== undefined) {
-            width = props.height * RATIO[aspectRatio];
+            width = typeof props.height === 'number' ? props.height * ratio : props.height;
         } else {
             width = '100%';
         }


src/image.tsx
```diff
@@ -37,13 +37,14 @@ export const RATIO = {
...
     src: string;
     url?: undefined;
     /** defaults to 100% when no width and no height are given */
-    width?: number;
-    height?: number;
+    width?: string | number;
+    height?: string | number;
     /** defaults to 1:1, if both width and height are given, aspectRatio is ignored */
-    aspectRatio?: AspectRatio;
+    aspectRatio: AspectRatio | number;
```
pladaria, line 47: keep this optional

src/image.tsx
```diff
@@ -80,9 +82,9 @@ const Image = React.forwardRef<HTMLImageElement, ImageProps | DeprecatedImagePro
             width = props.width;
             height = props.height;
         } else if (props.width !== undefined) {
-            height = props.width / RATIO[aspectRatio];
+            height = typeof props.width === 'number' ? props.width / ratio : props.width;
```
pladaria, line 85: when the height/width is a string it could be better to set it using calc, so the aspect ratio is preserved:
```js
typeof props.width === 'number' ? props.width / ratio : `calc(${props.width} / ratio)`
```

then, instead of passing the width/height as attributes to the image, it should be set using `style` or with jss

src/image.tsx
```diff
@@ -60,16 +61,17 @@ type DeprecatedImageProps = {
...
-    children?: void;
     dataAttributes?: DataAttributes;
+    children?: void;
+    borderRadius?: boolean;
 };
 
 const Image = React.forwardRef<HTMLImageElement, ImageProps | DeprecatedImageProps>(
-    ({aspectRatio = '1:1', alt = '', dataAttributes, ...props}, ref) => {
-        const noBorderRadius = useDisableBorderRadius();
+    ({aspectRatio = '1:1', alt = '', dataAttributes, borderRadius = false, ...props}, ref) => {
```
atabel, line 70: this is a breaking change, until now, Images had border radius by default.
How are we going to manage this? Should we release a major?

src/image.tsx
```diff
@@ -60,16 +61,17 @@ type DeprecatedImageProps = {
...
     alt?: string;
-    children?: void;
     dataAttributes?: DataAttributes;
+    children?: void;
+    noBorderRadius?: boolean;
 };
 
 const Image = React.forwardRef<HTMLImageElement, ImageProps | DeprecatedImageProps>(
-    ({aspectRatio = '1:1', alt = '', dataAttributes, ...props}, ref) => {
-        const noBorderRadius = useDisableBorderRadius();
```
atabel, line 69: I think we should still be listening to this context. It's being used in `MediaCard` right now.
Otherwise, in the places where we were using `MediaCards` with `Image` we'll see border radius if we don't explicitly set the new `noBorderRadius` prop.

pladaria (changes requested):
image
This is a bad example.

If you want to create different scenarios for the Image component, create a new story for Image.

src/image.tsx
pladaria: +1

This is the logic:
* default: with border radius
* if the useDisableBorderRadius returns true => remove border radius
* if the prop noBorderRadius is true => remove border radius

src/__stories__/list-story.tsx
```diff
@@ -189,7 +189,14 @@ const Template: StoryComponent<Args & {boxed?: boolean}> = ({
...
             <RowComponent
-                asset={<Image src="https://i.imgur.com/HRvhZ6F.jpeg" height={80} aspectRatio="16:9" />}
+                asset={
+                    <Image
+                        src="https://i.imgur.com/HRvhZ6F.jpeg"
+                        noBorderRadius
+                        height={80}
+                        aspectRatio="16:9"
+                    />
+                }
```
pladaria, line 199: this is a bad place to put a noBorderRadius example

atabel (approved):

PR #438, opened 2022-03-08
FixedFooterLayou Dark Mode improvement

yceballost:
image
vs
image

### DIFF
--- a/src/fixed-footer-layout.tsx
+++ b/src/fixed-footer-layout.tsx
@@ -130,7 +130,9 @@ const FixedFooterLayout: React.FC<Props> = ({
                  * This style is inline to avoid creating a class that may collide with
                  * other fixed footers during the page animation transition
                  */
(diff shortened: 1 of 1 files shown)


pladaria:
> @pladaria do you think that I need to change this color too? I don't really know where is applying
> 
> `37 backgroundColor: ({containerBgColor}) => containerBgColor || theme.colors.background,`

Look where this color is being used:
image
As you can see this is the container of the children, so it is the main content background. I wouldn't change it but try it and see what happens.

pladaria (approved):

yceballost:
> > @pladaria do you think that I need to change this color too? I don't really know where is applying
> > `37 backgroundColor: ({containerBgColor}) => containerBgColor || theme.colors.background,`
> 
> Look where this color is being used: It is inside a `className` definition (`containerSmall`):
> 
> image
> 
> And `containerSmall` it is being used here:
> 
> image As you can see this is the container of the children, so it is the main content background. I wouldn't change it but try it and see what happens.

true!

PR #441, opened 2022-03-15
feat(Chip): work like checkbox

atabel:
WEB-500


sergivillar:
should the cursor be 'pointer' if is a checkbox?

flow-defs/checkbox.js.flow
```diff
@@ -8,10 +8,12 @@ declare type RenderProps = {
   checked?: boolean,
   onChange?: (value: boolean) => void,
   id?: string,
-  render?: (
-    checkboxElement: React.Element<any>,
-    labelId: string
-  ) => React.Element<any>,
+  render?: (renderProps: {
```
atabel, line 11: this is a breaking change. Same for Switch and RadioButton

sergivillar:
I saw two thing that not sure if they are ok
first, in single/multiple selection, I'm missing pointer: 'cursor'. (in onPress it is ok)
second, when a chip is checked (with border blue) it seems kind of weird if I do hover and that border disappears, more over when I do click in a new item in Single selection and the border doesn't show till onBlur on that chip. Is this expected?

https://user-images.githubusercontent.com/6709771/158549804-050056b0-3497-4d47-aeee-982e1d16eaeb.mp4

src/chip.tsx
```diff
@@ -27,7 +28,8 @@ const useStyles = createUseStyles(({colors, mq}) => ({
         '& > span': {
             color: colors.textPrimary, // Giving color to text
         },
-
+    },
+    interactive: {
```
yceballost, line 32: can we remove hover state only when a chip is selected?

src/chip.tsx
```diff
@@ -27,7 +28,8 @@ const useStyles = createUseStyles(({colors, mq}) => ({
...
         [mq.supportsHover]: {
```
yceballost, line 33: missing cursor pointed in new behaviors

flow-defs/checkbox.js.flow
```diff
@@ -8,10 +8,12 @@ declare type RenderProps = {
...
+    iconElement: React.Element<any>,
```
pladaria, line 12: checkboxElement? to me the icon name is misleading

src/chip.tsx
```diff
@@ -39,29 +41,66 @@ const useStyles = createUseStyles(({colors, mq}) => ({
             },
         },
```
yceballost, line 42: weird colors in prominent and classic o2 while hover
image
image

I'm going to check constant usage

yceballost (changes requested):

flow-defs/checkbox.js.flow
atabel: I don't like `checkboxElement`, because it's not the checkbox. For me, the checkbox element is the element with the `role="checkbox"`, and this is just the checkbox icon.
About `actionElement`, hmm... not sure if it's better or worse

atabel:
> I saw two thing that not sure if they are ok first, in single/multiple selection, I'm missing pointer: 'cursor'. (in onPress it is ok) second, when a chip is checked (with border blue) it seems kind of weird if I do hover and that border disappears, more over when I do click in a new item in Single selection and the border doesn't show till onBlur on that chip. Is this expected?
> 
>  Kapture.2022-03-16.at.09.36.41.mp4

@yceballost wdyt?

src/chip.tsx
atabel: checkboxes don't use to have cursor pointer

src/chip.tsx
atabel: I don't understand this

yceballost: it is quite weird interaction removing border in hover state when a chip is selected. Keep the hover behavior but not the style that remove the border

https://user-images.githubusercontent.com/6722153/158561686-2608a846-63e3-4afe-b0d6-f15f7b8e9a64.mov

atabel: ok, try now

yceballost: perfect ✨

sergivillar (approved):

RubenMorPov (approved):

flow-defs/checkbox.js.flow
pladaria: controlElement ?

pladaria (approved):

yceballost (approved):

PR #451, opened 2022-03-23
fix(Video, Image): aspect ratio fix for old browsers

atabel:
`aspect-ratio` css property is not supported by Chrome < 89 and Safari < 15. We use that property for mistica Image and Video components.

This PR introduces an alternative workaround (padding technique) and applies it only when the browser doesn't have native support for `aspect-ratio`.


pladaria (approved):

src/video.tsx
```diff
@@ -17,12 +17,22 @@ export const RATIO = {
...
+
+    wrapper: {
+        borderRadius: ({noBorderRadius}) => (noBorderRadius ? 0 : 4),
+        overflow: 'hidden',
         maxWidth: '100%',
         maxHeight: '100%',
-        objectFit: 'cover',
-        aspectRatio: ({aspectRatio}) => aspectRatio ?? 'unset',
+        position: 'relative',
+        paddingTop: ({aspectRatio}) => (aspectRatio ? `${100 / aspectRatio}%` : 'initial'),
```
pladaria, line 35: I'd implement the same fix for images

atabel: Yes, I'm on it. But I don't like the implementation. I think I'd try to feature-detect the aspect-ratio support and only apply this hack when needed. I don't wan't every Image to have a div wrapper :(

src/image.tsx
```diff
@@ -69,24 +86,33 @@ const Image = React.forwardRef<HTMLImageElement, ImageProps>(
             width = props.width;
             height = props.height;
         } else if (props.width !== undefined) {
-            height = typeof props.width === 'number' ? props.width / ratio : `calc(${props.width} / ratio)`;
+            height = typeof props.width === 'number' ? props.width / ratio : undefined;
```
atabel, line 89: that `calc` makes no sense

src/image.tsx
```diff
@@ -52,15 +69,15 @@ export type ImageProps = {
...
 const Image = React.forwardRef<HTMLImageElement, ImageProps>(
-    ({aspectRatio = '1:1', alt = '', dataAttributes, noBorderRadius, ...props}, ref) => {
+    ({aspectRatio = '1:1', alt = '', dataAttributes, noBorderRadius, src, ...props}, ref) => {
+        const supportsAspectRatio = useSupportsAspectRatio();
         const ratio = typeof aspectRatio === 'number' ? aspectRatio : RATIO[aspectRatio];
         const noBorderRadiusContext = useDisableBorderRadius();
         const noBorderSetting = noBorderRadius ?? noBorderRadiusContext;
         const classes = useStyles({
             noBorderRadius: noBorderSetting,
             aspectRatio: !props.width && !props.height ? ratio : undefined,
```
atabel, line 79: I think I don't understand this condition :/
shouldn't we always pass the `aspectRatio`, @pladaria ?

src/utils/aspect-ratio-support.tsx
```diff
@@ -0,0 +1,28 @@
...
+    children: React.ReactNode;
+};
+
+export const AspectRatioSupportProvider: React.FC<Props> = ({children}) => {
+    // In SSR (and in first client side render) we assume the browser will support it
+    const [isSupported, setIsSupported] = React.useState(true);
+
+    useIsomorphicLayoutEffect(() => {
+        try {
+            if (!CSS.supports('aspect-ratio', '1 / 1')) {
```
pladaria, line 16: :O TIL

src/image.tsx
atabel: fixed the logic

RubenMorPov (approved):

pladaria (approved):

PR #463, opened 2022-04-06
fix(types): make skin in ThemeConfig readonly

atabel:
without this change, I have this kind of errors in flow:
image

This is the explanation of the error:
https://flow.org/en/docs/faq/#toc-why-can-t-i-pass-a-string-to-a-function-that-takes-a-string-number

### DIFF
--- a/flow-defs/theme.js.flow
+++ b/flow-defs/theme.js.flow
@@ -79,7 +79,7 @@ declare type LinkComponent = React.ComponentType<{
 declare export var AnchorLink: LinkComponent;
 export type ColorScheme = "dark" | "light" | "auto";
 export type ThemeConfig = {
-  skin: Skin,
+  skin: $ReadOnly<Skin>,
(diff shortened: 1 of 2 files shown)


flow-defs/theme.js.flow
```diff
@@ -79,7 +79,7 @@ declare type LinkComponent = React.ComponentType<{
 declare export var AnchorLink: LinkComponent;
 export type ColorScheme = "dark" | "light" | "auto";
 export type ThemeConfig = {
-  skin: Skin,
+  skin: $ReadOnly<Skin>,
```
SergioMorchon, line 82: Ahá! Nice catch (thanks for pointing to https://flow.org/en/docs/faq/#toc-why-can-t-i-pass-a-string-to-a-function-that-takes-a-string-number)

SergioMorchon (approved):

pladaria (approved):

PR #467, opened 2022-04-20
feat(ButtonLink): allow disabled state

atabel:
https://jira.tid.es/browse/WEB-518


yceballost (approved):

yceballost:
Please, change [button story](https://mistica-jfp62848j-tuentisre.vercel.app/?path=/story/components-touchables-button--type-of-buttons) to add disabled behavior to ButtonLink

pladaria (approved):

src/button.tsx
```diff
@@ -517,6 +519,7 @@ export const ButtonLink = React.forwardRef<
                 {props.children}
             </Text2>
         ),
+        disabled: props.disabled,
```
pladaria, line 522: You should also read the form state and if sending, automatically disable the button.

See the button implementation

atabel: I see we are looking for form status in TextLinks too, so it makes sense to be consistent here

atabel:
> Please, change [button story](https://mistica-jfp62848j-tuentisre.vercel.app/?path=/story/components-touchables-button--type-of-buttons) to add disabled behavior to ButtonLink

done

src/button.tsx
atabel: done

pladaria (approved):

PR #470, opened 2022-04-26
CHECKOUT-3340 Fixed background color for plain bakgrounds on feedback screens

Marcosld:


src/feedback.tsx
```diff
@@ -100,6 +100,18 @@ const useStyles = createUseStyles((theme) => ({
     },
 }));
 
+const BackgroundColor: React.FC = () => {
+    const {colors} = useTheme();
+    const isInverse = useIsInverseVariant();
+    React.useEffect(() => {
+        document.body.style.background = isInverse ? colors.backgroundBrand : colors.background;
+        return () => {
+            document.body.style.background = 'initial';
```
atabel, line 109: hmmm... this could break webapp code to set body backgound depending on light/dark mode
https://github.com/Telefonica/webapp/blob/master/web/src/common/components/body-background.js

I'd use a `<style>` tag instead.

atabel (changes requested):

pladaria (approved):

PR #475, opened 2022-05-05
update icon in password form

yceballost:
https://jira.tid.es/browse/WEB-545

Use mística icons in password form
Also adding a little fade animation in hover and adjusting hover size to 40


src/password-field.tsx
```diff
@@ -3,8 +3,8 @@ import {useFieldProps} from './form-context';
 import {TextFieldBaseAutosuggest} from './text-field-base';
 import {useTheme} from './hooks';
 import IconButton from './icon-button';
-import Visibility from './icons/icon-visibility';
-import VisibilityOff from './icons/icon-visibility-off';
```
atabel, line 7: I think we can remove these files

PR #478, opened 2022-05-11
feat(Image): expose onError and onLoad. Avoid broken icon

pladaria:
WEB-553


src/image.tsx
```diff
@@ -113,16 +119,31 @@ const Image = React.forwardRef<HTMLImageElement, ImageProps>(
...
+                    setIsError(false);
+                    onLoad?.(event);
+                }}
                 {...(!needsWrapper ? {width, height} : {})}
             />
         );
 
+        if (isError) {
+            // to avoid showing the broken image icon
+            return null;
```
atabel, line 144: hmmm... the problem is that you'll cause a layout shift if the image load fails

pladaria: same problem happens now if the image is broken

atabel (approved):

PR #479, opened 2022-05-13
chore(Examples): Update examples

pladaria:


examples/next/pages/_app.js
```diff
@@ -21,12 +21,12 @@ const calcLinkProps = (to) => {
     return {href: to};
 };
 
-const NextLinkAdapter = ({to, innerRef, children, ...props}) => (
-    <Link {...calcLinkProps(to)}>
+const Link = ({to, innerRef, children, ...props}) => (
+    <NextLink {...calcLinkProps(to)}>
```
atabel, line 25: I think the `calcLinkProps` is not needed anymore with last Next versions, because the `as` prop is not needed now.

atabel (approved):

PR #481, opened 2022-05-17
chore(playroom): update version

yceballost:
New features!! 🍿
- Now playroom support dark mode

- [x]  Preview now shows again the splash preview 😞 [#258](https://github.com/Telefonica/mistica-web/pull/258)


atabel:
do you think it makes sense to "propagate" the darkmode setting to the preview? I think we may be able to do it

playroom/frame-component.tsx
```diff
@@ -31,7 +31,7 @@ const App: React.FC = ({children}) => {
 
         ${isModalOpen ? 'body {overflow-y: hidden}' : ''}
 
-        *[class^='SplashScreen__'] {
+        *[class^='_'] {
```
yceballost, line 34: @atabel this solve the splashscreen.. but.. maybe is too generic? I don't know if this one could cause problems in other places

atabel: yes, this is too generic and may bite us

playroom/frame-component.tsx
```diff
@@ -31,7 +31,7 @@ const App: React.FC = ({children}) => {
 
         ${isModalOpen ? 'body {overflow-y: hidden}' : ''}
 
-        *[class^='SplashScreen__'] {
+        *[class^='_1fu0koy1'] {
```
yceballost, line 34: @atabel I tried with this class that is more specific. I tried to create new branches and build with `npx vercel` and kept the same class name. We can try with this solution and see what happen when we include new components. If the classes changes, maybe we will have to roll back de playroom version

atabel (approved):

PR #491, opened 2022-06-14
fix(Video): avoid flickering when playing video in Safari

pladaria:


kydorn (approved):

src/video.tsx
```diff
@@ -156,11 +163,11 @@ const Video = React.forwardRef<HTMLVideoElement, VideoProps>(
                 autoPlay={autoPlay}
                 muted={muted}
                 loop={loop}
-                {...(!needsWrapper ? {width, height} : {})}
+                {...(needsWrapper ? {} : {width, height})}
                 className={classes.video}
                 preload={preload}
                 // This transparent pixel fallback avoids showing the ugly "play" image in android webviews
-                poster={poster || TRANSPARENT_PIXEL}
+                poster={needsWrapper ? TRANSPARENT_PIXEL : poster || TRANSPARENT_PIXEL}
```
kydorn, line 170: if we call video.load(), does it display the first video frame or the poster?

kydorn (approved):

RubenMorPov (approved):

src/video.tsx
pladaria: video.load() pauses the video (if playing) and displays the poster. If the poster is a transparent pixel, the background image from the wrapper will be visible.

You can check here: https://t.ly/OmOD (use safari)

PR #492, opened 2022-06-14
feat(DataCard,MediaCard): allow dismissable cards

atabel:
jira: WEB-520


src/card.tsx
```diff
@@ -121,18 +141,33 @@ type MediaCardProps = {
     button?: RendersNullableElement<typeof ButtonPrimary>;
     buttonLink?: RendersNullableElement<typeof ButtonLink>;
     children?: void;
+    dataAttributes?: DataAttributes;
     'aria-label'?: string;
+    onClose?: () => void;
```
yceballost, line 146: image

www.rebrand.ly/577xrw2

close icon goes to Parla when you increase the width of the Media Card

yceballost (changes requested):

src/card.tsx
pladaria: se fue a echar gasolina

src/card.tsx
```diff
@@ -145,15 +180,23 @@ export const MediaCard = React.forwardRef<HTMLDivElement, MediaCardProps>(
...
+                </MaybeSection>
             </Boxed>
         );
+        return onClose ? (
+            <Dismissable onClose={onClose} aria-label={ariaLabel}>
+                {content}
+            </Dismissable>
+        ) : (
+            content
+        );
```
pladaria, line 192: perhaps the Dismissable component could perform this logic inside (in fact it would be a MaybeDismissable), but you don't need to repeat this logic in other places

pladaria (approved):

src/card.tsx
atabel: fixed

yceballost (approved):

RubenMorPov (approved):

PR #496, opened 2022-06-21
fix(MediaQuery): Improve media query for touchable-only devices

SergioMorchon:
This aims to prevent from Mística the same issue detected at ACCOUNT-17156.

### DIFF
--- a/flow-defs/utils/media-queries.js.flow
+++ b/flow-defs/utils/media-queries.js.flow
@@ -15,3 +15,8 @@ declare export var createMediaQueries: (x: {
   desktopOrBigger: string,
   supportsHover: string,
 };
+/**
+ * Scopes the styles to touchable devices.
+ * See: https://stackoverflow.com/questions/12469875/how-to-code-css-media-queries-targeting-all-mobile-devices-and-tablets/42835826#42835826
+ */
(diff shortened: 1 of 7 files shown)


flow-defs/utils/media-queries.js.flow
```diff
@@ -15,3 +15,8 @@ declare export var createMediaQueries: (x: {
   desktopOrBigger: string,
   supportsHover: string,
 };
+/**
+ * Scopes the styles to touchable devices.
+ * See: https://stackoverflow.com/questions/12469875/how-to-code-css-media-queries-targeting-all-mobile-devices-and-tablets/42835826#42835826
+ */
+declare export var TOUCHABLE_ONLY: "@media (pointer: coarse), @media (hover: none)";
```
atabel, line 22: For ergonomics, I'd move this to the theme context, as que do with other mq

atabel (approved):

flow-defs/utils/media-queries.js.flow
SergioMorchon: I like that, I missed it.
Indeed that exposes it to webapp and other consumers without duplicating the query

pladaria (approved):

PR #501, opened 2022-06-29
chore(playroom): added alerts/confirm snippets

yceballost:


playroom/snippets.tsx
```diff
@@ -1192,6 +1192,82 @@ const avatarSnippets = [
...
+      acceptText: "Continue",
+    })
+  }
+>
+  Open one button
+</ButtonPrimary>`,
+    },
+    {
+        group: 'Dialogs',
+        name: 'Alert two buttons',
```
atabel, line 1213: technically, this isn't an "alert", but a "confirm"

aweell (approved):

playroom/snippets.tsx
yceballost: @aweell do you think this detail should be documented in Brand Factory? the difference between alert and confirm

atabel (approved):

playroom/snippets.tsx
aweell: Maybe it would be more appropriate to name this component "modal dialog" and `alert` or `confirm` as variants of it. WDYT?

And then can be documented for which cases can each of them be used?

yceballost: At the same time, this component is tricky, because nowadays, this component is built to have a native equivalence and you can't use this modal dialog to introduce anything inside. So I think that detail is important.

I have in the scope of the roadmap to have an alternative Dialog component to allow include anything inside, for example, to design funnels in dialogs.

fyi

PR #502, opened 2022-07-04
chore(Storybook): reorganize stories

aweell:


pladaria:
Looks like you have some accessibility issues related to the `scrollable-region-focusable` rule.

Please, 

1. create a ticket to WEB so we can investigate and fix that issue
2. add that rule to the ignore list: https://github.com/Telefonica/mistica-web/blob/master/.github/actions/audit-accessibility/main.js#L79. Add a comment with the related ticket

pladaria (approved):

atabel (approved):
nice!

PR #511, opened 2022-07-29
Circles color combination

yceballost:
Circle's color combination as a default now is: `brandLow` + `brand`

This PR change stories example and playroom snippets.

I also added to the list story two new examples with list + avatar


yceballost:
- I added a rule to ignore this accessibility report
- I created a mistica-web ticket → https://jira.tid.es/browse/WEB-616

playroom/snippets.tsx
```diff
@@ -1048,9 +1069,10 @@ const navigationBarSnippets = [
         </Badge>
       </NavigationBarAction>
       <NavigationBarAction onPress={() => {}} aria-label="Open profile">
-        <Circle
-          backgroundImage="https://i.pravatar.cc/100?img=31"
+        <Avatar
           size={isDesktopOrBigger ? 32 : 24}
+          initials="ML"
+          src="https://i.pravatar.cc/100?img=31"
```
atabel, line 1075: does it make sense to have `intials` and `src` at the same time?

atabel (approved):

playroom/snippets.tsx
pladaria: Yes, if the image fails, it fallbacks to initials

https://github.com/Telefonica/mistica-web/blob/master/src/avatar.tsx#L104

pladaria (approved):

PR #517, opened 2022-08-16
prop table

yceballost:


sergivillar:
will this PR be merged @yceballost ?

yceballost:
> will this PR be merged @yceballost ?

I couldn't get the way to work properly this addon :(. Prop table generated is not correct in the most of stories

PR #518, opened 2022-08-17
fix(ButtonLink): Fix paddings

aweell:
The current buttonLink `padding` value is not following the specs.

https://www.figma.com/file/koROdh3HpEPG2O8jG52Emh/Buttons-Component-Specs?node-id=0%3A1

Expected: `padding: 6px 12px;`


pladaria:
You need to generate type definitions and commit them. From the repo root run `yarn build` then commit and push the updated files.

I guess you also need to regenerate some screenshot tests.

src/button.tsx
```diff
@@ -436,12 +436,13 @@ const Button: React.FC<
...
 const useButtonLinkStyles = createUseStyles((theme) => {
-    const padding = 6;
+    const paddingY = 6;
+    const paddingX = 12;
     return {
         link: {
             display: 'inline-block',
             width: 'auto',
-            padding,
+            padding: `${paddingY}px ${paddingX}px`,
```
yceballost, line 445: wrong alignment in callouts and cards
https://ibit.ly/HwBP
https://ibit.ly/YsS-

And hover in this case is quite "ugly"
https://ibit.ly/JDv_

yceballost (changes requested):

src/button.tsx
```diff
@@ -16,7 +16,7 @@ import type {DataAttributes, RendersElement, RendersNullableElement, TrackingEve
 import type {Location} from 'history';
 import type {Theme} from './theme';
 
-export const BUTTON_MIN_WIDTH = 136;
+export const BUTTON_MIN_WIDTH = 104;
```
yceballost, line 19: Specs say 136
image

now the buttonPrimary has weird min-width
https://ibit.ly/2oW_
image

(before)
https://ibit.ly/vY8v
image

yceballost (changes requested):

src/button.tsx
aweell: Fixed. This problem was related to the buttongrouop component

src/button.tsx
aweell: My error reading the specs, fixed primary min-width

yceballost (approved):
lgtm!

pladaria (approved):

PR #522, opened 2022-08-25
fix(Spinner): fix flicker pixel in spinner

yceballost:
https://user-images.githubusercontent.com/6722153/186729292-38382d19-c51c-4215-a0bf-d6e17d1989b5.mov

Exterminating this annoying pixel🤬

WEB-23


pladaria (approved):
good catch!

PR #527, opened 2022-09-01
fix(NavigationBarLogo): alignment and Blau's size

yceballost:
NavigationBarLogo component has two problems fixed:
- Blau logo size was wrong
- Logo wasn't aligned with the center

An extra ball, I removed evenodd from svgs and I exported the logos with the same dimensions as we have in Figma


pladaria (approved):
gr8

src/navigation-bar.tsx
```diff
@@ -138,25 +109,40 @@ type NavigationBarLogoProps = {
...
+            break;
         case 'Blau':
-            return <BlauLogo size={size} />;
+            logo = <BlauLogo size={size} />;
+            break;
         default:
-            return null;
+            logo = null;
     }
+    return <div className={classes.logoContainer}>{logo}</div>;
```
atabel, line 145: hmm... not sure if this is the best place for this div
I'd move it to the header, this component should just return an svg

yceballost: How can I resolve this alignment problem?
image

I can't found other alternative 🤔

https://ibit.ly/Qlrn

atabel (approved):

PR #528, opened 2022-09-06
fix(TextField): improve styles for readOnly state

yceballost:
I also fix this little issue, textArea was using native top padding.

https://user-images.githubusercontent.com/6722153/188650161-7aaf4c89-885c-4891-948c-31f236ec6e45.mov


src/text-field-components.tsx
```diff
@@ -191,7 +191,8 @@ const useFieldContainerStyles = createUseStyles((theme) => ({
         },
         display: 'flex',
         position: 'relative',
-        backgroundColor: theme.colors.backgroundContainer,
+        backgroundColor: ({readOnly}) =>
+            readOnly ? theme.colors.backgroundAlternative : theme.colors.backgroundContainer,
```
atabel, line 195: I think these are the same color in dark mode

atabel (approved):

src/text-field-components.tsx
yceballost: Yep, @aweell is reviewing this with the new inverse/alternative logic. Maybe in dark mode we have change boxed logic to keep border in this elements. But we are working on it.

Another solution to this is that readOnly appear with background color in dark mode to differentiate from default state
image

pladaria (approved):

src/text-field-components.tsx
yceballost: we are going to use neutralLow for readOnly by the moment, this color in dark mode is slightly different than `backgroundAlternative`, which is not the best solution but we have to improve a lot dark mode colors (in general)

The usage of readOnly prop is not very common and by the moment is a commitment that we are willing to have in dark mode

image

flow-defs/text-field-base.js.flow
```diff
@@ -52,7 +52,6 @@ export type CommonFormFieldProps = {
   value?: string,
   defaultValue?: string,
   onChange?: (event: SyntheticEvent<HTMLInputElement>) => void,
-  readOnly?: boolean,
```
SergioMorchon, line 55: I think this has removed the `readOnly` property from the `TextField` component props:
image

This is because `TextField` has `TextFieldProps` that extends from  `CommonFormFieldProps`, and none of them has the `readOnly` prop since this change.
@atabel @pladaria @yceballost

pladaria: https://jira.tid.es/browse/WEB-701

PR #529, opened 2022-09-07
fix input fields styles

yceballost:
- [x] perfect alignment in select input with text field in chrome
- [ ] perfect alignment in select input with text field in safari 
- [x] Changed input password text to use Lucida Grande instead of Arial in MacOS ([reverted in migration](https://github.com/Telefonica/mistica-web/commit/db5505d6e9626e17eba36d2dcb69fd37b5774d34))
- [x] I also branded cvv asset and changed this design to use text presets
- [x] Changed icons in CVV tooltips to use Mística colors
- [x] Changed icon in credit card field to use mistica icons. Removed the old one.
- [x] Changed search field icon to use mistica icons. Removed the old one.

image
https://rebrand.ly/y0alq3h


pladaria (approved):

src/select.tsx
```diff
@@ -52,14 +52,18 @@ const useStyles = createUseStyles((theme) => ({
...
         lineHeight: '20px',
-        fontSize: 16,
+        fontSize: 18,
         color: theme.colors.textPrimary,
         opacity: ({disabled}) => (disabled ? 0.5 : 1),
         maxWidth: '100%',
         textOverflow: 'ellipsis',
         whiteSpace: 'nowrap',
+        [theme.mq.tabletOrSmaller]: {
+            top: ({label}) => (label ? 27 : 27),
```
atabel, line 64: always 27?
```suggestion
            top: 27,
```

src/select.tsx
```diff
@@ -52,14 +52,18 @@ const useStyles = createUseStyles((theme) => ({
         left: 12 + 1, // 12 for select paddingLeft and +1 for border
         right: 48 + 1, // 48 for icon and +1 for border
         overflow: 'hidden',
-        top: ({label}) => (label ? 27 : 17),
+        top: ({label}) => (label ? 30 : 17),
```
yceballost, line 55: I don't understand why this label logic, I didn't see any changes if I change 17 value :/ @atabel. Maybe we can remove it?

PR #530, opened 2022-09-08
feat(Button): support for pre/post icons

atabel:
jira: WEB-625


src/button.tsx
```diff
@@ -21,19 +21,20 @@ export const BUTTON_MIN_WIDTH = 136;
...
 
 const BORDER_PX = 1.5;
-const SPINNER_MARGIN_PX = 8;
+const ICON_MARGIN_PX = 8;
 const X_PADDING_PX = 16 - BORDER_PX;
 const Y_PADDING_PX = 12 - BORDER_PX;
 const X_SMALL_PADDING_PX = 12 - BORDER_PX;
 const Y_SMALL_PADDING_PX = 6 - BORDER_PX;
+const SMALL_ICON_SIZE = 20;
+const ICON_SIZE = 24;
```
yceballost, line 31: something happens in playroom snippet
https://ibit.ly/7ygh

yceballost (changes requested):

src/button.tsx
atabel: weird... it works fine in storybook. I'll take a look

src/button.tsx
```diff
@@ -217,6 +219,62 @@ const useDangerButtonStyles = createUseStyles((theme) => ({
...
+        accText = [];
+    };
+    childrenArr.forEach((element, idx) => {
+        const isFirstChild = idx === 0;
+        const isLastChild = idx === length - 1;
+
+        const isIconElement =
+            React.isValidElement(element) &&
+            typeof element.type !== 'string' &&
+            element.type.name?.startsWith('Icon'); // all mistica icons start with Icon prefix
```
pladaria, line 248: You could be passing an element named IconWhatever that is not a mistica icon

src/button.tsx
```diff
@@ -217,6 +219,62 @@ const useDangerButtonStyles = createUseStyles((theme) => ({
...
+                    key={resultChildrenArr.length}
+                    style={{
+                        display: 'flex',
+                        alignItems: 'center',
+                        marginLeft: isFirstChild ? 0 : ICON_MARGIN_PX,
+                        marginRight: isLastChild ? 0 : ICON_MARGIN_PX,
+                    }}
+                >
+                    {React.cloneElement(element, {size: pxToRem(sizeInPx)})}
+                </div>
```
pladaria, line 266: mistica icons accept a style prop. The div wrapper shouldn't be needed

src/button.tsx
```diff
@@ -217,6 +219,62 @@ const useDangerButtonStyles = createUseStyles((theme) => ({
...
+    const childrenArr = React.Children.toArray(content);
+    const resultChildrenArr: Array<React.ReactNode> = [];
+    let accText: Array<React.ReactNode> = [];
+    const flushAccText = () => {
+        resultChildrenArr.push(
+            <React.Fragment key={resultChildrenArr.length}>{renderText(accText)}</React.Fragment>
+        );
+        accText = [];
+    };
+    childrenArr.forEach((element, idx) => {
```
pladaria, line 241: Instead of this, wouldn't be better to accept a startIcon / endIcon via props?

src/button.tsx
```diff
@@ -21,19 +21,20 @@ export const BUTTON_MIN_WIDTH = 136;
 const transitionTiming = '0.3s cubic-bezier(0.77, 0, 0.175, 1)';
```
pladaria, line 21: are the button width changes in screenshots expected?

src/button.tsx
atabel: that's fine, that element would likely be an icon too

src/button.tsx
atabel: yes, but we already supported icons as children so I didn't want to change the component api

src/button.tsx
atabel: As you said, it could be a non-mistica icon too, and the component may not have style prop. The extra div always works

aweell (approved):

src/button.tsx
```diff
@@ -217,6 +221,59 @@ const useDangerButtonStyles = createUseStyles((theme) => ({
...
+
+        if (isIconElement) {
+            if (accText.length) {
+                flushAccText();
+            }
+            const sizeInPx = element.props.size ?? defaultIconSize;
+            resultChildrenArr.push(
+                <div
+                    key={resultChildrenArr.length}
+                    style={{
```
brtbrt, line 257: just for understanding purposes: why the style in place of a class in this case?

brtbrt (approved):

src/button.tsx
```diff
@@ -21,19 +21,22 @@ export const BUTTON_MIN_WIDTH = 136;
...
 const BORDER_PX = 1.5;
-const SPINNER_MARGIN_PX = 8;
+const ICON_MARGIN_PX = 8;
 const X_PADDING_PX = 16 - BORDER_PX;
 const Y_PADDING_PX = 12 - BORDER_PX;
 const X_SMALL_PADDING_PX = 12 - BORDER_PX;
 const Y_SMALL_PADDING_PX = 6 - BORDER_PX;
+const ICON_SIZE = 24;
+const SMALL_ICON_SIZE = 20;
+const SPINNER_SIZE = 24;
```
yceballost, line 31:
```suggestion
const SPINNER_SIZE = 20;
```
Change this one to make the loading similar to icon size (icons have 2px padding inside, the real vector size is 20px)

atabel: hmmm... are you sure? it was 24 before my changes

src/button.tsx
atabel: `marginLeft` and `marginRight` are dynamic styles, so it's easier and more convenient to use `style` here, although I could have 3 different classNames and combine them too

src/button.tsx
yceballost: Yes, I know, we was talking @aweell and I to make this little change to keep a smooth consistency across icons and spinners

pladaria (approved):

yceballost:
I think this ticket is solved with this pr
https://jira.tid.es/browse/WEB-598

PR #531, opened 2022-09-08
feat(Skeletons): new components

brtbrt:
Added skeletons based on https://www.figma.com/file/w7E0mmB92eio0zHw7h9iS2/%5BREADY%5D-Skeletons-Specs?node-id=986%3A1161

in order to be passing this PR should be merged before: https://github.com/Telefonica/js-toolbox/pull/50


src/skeletons/skeleton-line.tsx
```diff
@@ -0,0 +1,51 @@
...
+    const isInverse = useIsInverseVariant();
+    const animationClasses = useAnimation({disableAnimation});
+    const classes = useLineStyle({width, height, isInverse});
+
+    return (
+        <div
+            className={`${classes.row} ${animationClasses.animation}`}
+            aria-valuetext={ariaValueText}
+            role={ariaValueText === undefined ? undefined : 'progressbar'}
+            // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
```
brtbrt, line 44: in case there's no role, as seen on the line above, tab index should not be defined

src/skeletons/styles.tsx
```diff
@@ -0,0 +1,29 @@
+import {createUseStyles} from '../jss';
+
+type AnimationProps = {
+    disableAnimation: boolean;
+};
+
+const transition = '1s linear';
```
yceballost, line 7: increase transition time to 1.5s (I will change specs)

src/skeletons/styles.tsx
```diff
@@ -0,0 +1,29 @@
+import {createUseStyles} from '../jss';
+
```
yceballost, line 2: This question is for @atabel 

In other components we have "base" component, should we have something similar with skeletons? skeleton-base.tsx

For example, textField use text-field-base, something similar for feedbacks, successFeedbackScreen use FeedbackScreen as a base.

src/skeletons/skeleton-text.tsx
```diff
@@ -0,0 +1,54 @@
...
+type TextStylesProps = {
+    isInverse: boolean;
+};
+
+const useTextStyles = createUseStyles(({colors}) => ({
+    line: {
+        borderRadius: 8,
+        height: 8,
+        width: '100%',
+        background: ({isInverse}: TextStylesProps) =>
```
yceballost, line 18: should be this style in skeletons styles across to any skeleton instead of repeat colors in each skeleton?

src/skeletons/skeleton-text.tsx
```diff
@@ -0,0 +1,54 @@
...
+            isInverse ? colors.backgroundSkeletonInverse : colors.backgroundSkeleton,
+    },
+    lastLine: {
+        width: '75%',
+    },
+}));
+
```
yceballost, line 25: https://ibit.ly/MmZD

I made this example, skeletons doesn't works with `<Inline>` component to make compositions like this
image


For other side, do you think is interesting to add `width` prop to media skeleton to make skeleton buttons instead of wrap skeletons with box with custom width?

```
<Box width={150}>
	<SkeletonMedia ariaValueText="loading" height={48} />
</Box>
```

image

yceballost (changes requested):

src/skeletons/skeleton-text.tsx
yceballost: And build for example row skeleton with two skeleton? like

```
<Inline space={16}>
	<SkeletonCircle />
	<SkeletonRect />
</Inline>
````

I think it is easier to build skeleton screens if you have some existing typology. I would keep SkeletonText and SkeletonRow as is. It is true that SkeletonMedia has a weird name if we want to use it as a button. Maybe rename SkeletonMedia to SkeletonRect? @aweell

Comment (aweell) [reply src/skeletons/skeleton-text.tsx:25]: I would definitely keep `text`, but I'm not sure about `row`. Since the typology doesn't cover all existing components I suggest a more agnostic approach: `square`, `circle` ( and maybe `rounded?)` (With a `width` and `height` props and build the replacement with the layout components. 

My point is: Why the `row` should have a direct replacement and not other components? Based on usage? Is it better to have a list of components that should have skeleton version or something that allows building any component?

File: src/skeletons/skeleton-circle.tsx [hunk 6]
```diff
@@ -0,0 +1,46 @@
...
+
+    return (
+        <div
+            className={`${classes.circle} ${animationClasses.animation}`}
+            aria-valuetext={ariaValueText}
+            role="progressbar"
+            tabIndex={0}
+            aria-busy
+            aria-valuemin={0}
+            aria-valuemax={100}
```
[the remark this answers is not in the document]
aweell: I suggest using `aria-hidden="true"` by default for all skeleton elements and instead show a visually hidden text (Is `aria-valuetext` for that?) that provides information about the loading state to screen readers, when the loading is finished, the content should flip the `aria-busy` from `true` to `false` and the `aria-valuetext` must be hidden.

src/skeletons/skeleton-text.tsx
yceballost: Yes, Row Skeleton is because row design is spread over a bunch of screens

src/skeletons/skeleton-circle.tsx
```diff
@@ -0,0 +1,46 @@
...
+        width: ({size}: CircleStyleProps) => size,
+        background: ({isInverse}: CircleStyleProps) =>
+            isInverse ? colors.backgroundSkeletonInverse : colors.backgroundSkeleton,
+    },
+}));
+
+type SkeletonCircleProps = {
+    size?: string | number;
+    ariaValueText?: string;
+    disableAnimation?: boolean;
```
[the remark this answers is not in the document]
brtbrt: basically it's for accessibility purposes. we discussed in mistica-web, but I can get rid off it

src/skeletons/skeleton-circle.tsx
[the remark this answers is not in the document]
brtbrt: what about https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/status_role?
i.e.
https://flowbite.com/docs/components/skeleton/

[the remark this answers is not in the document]
brtbrt: to recap, we could do:
`role="status"`
`aria-hidden="true"` if no `aria-valuetext` provided
`aria-busy` + `aria-valuetext` if provided

what do you think?

src/skeletons/skeleton-text.tsx
brtbrt: I'm unifying all the components into a single file, so yes, I'll apply DRY

pladaria (approved):

src/skeletons/styles.tsx
atabel: those are internal implementation details, not a big deal

src/skeletons/skeleton-text.tsx
yceballost: This issue with skeletonRectangle inside an inline doesn't work
https://ibit.ly/DZlC

src/__stories__/skeleton-circle-story.tsx
```diff
@@ -1,13 +1,13 @@
...
 export default {
     title: 'Components/Skeletons/Skeleton Circle',
 };
 
 type Args = {
-    size: number;
+    size?: number;
     inverse: boolean;
     disableAnimation: boolean;
     ariaValueText: string;
```
atabel, line 13: ariaLabel?

src/index.tsx
```diff
@@ -106,11 +106,7 @@ export {default as IconError} from './icons/icon-error';
 export {default as IconSuccess} from './icons/icon-success';
 export {default as IconSuccessVivo} from './icons/icon-success-vivo';
 
-// Skeletons
-export {default as SkeletonText} from './skeletons/skeleton-text';
-export {default as SkeletonCircle} from './skeletons/skeleton-circle';
-export {default as SkeletonRow} from './skeletons/skeleton-row';
-export {default as SkeletonMedia} from './skeletons/skeleton-media';
+export * from './skeletons';
```
atabel, line 109: I'd prefer to be explicit here and list all the exports, we could add a new export to skeletons in the future and want it to be used internally only, with the `*` we may end up making it public without noticing

src/skeletons.tsx
```diff
@@ -0,0 +1,184 @@
...
+            role="status"
+            aria-hidden={ariaLabel === undefined}
+            aria-busy={ariaLabel !== undefined}
+            aria-label={ariaLabel}
+        >
+            <SkeletonBase height={height} width={width} />
+        </div>
+    );
+};
+export {SkeletonLine, SkeletonText, SkeletonCircle, SkeletonRow, SkeletonRectangle};
```
atabel, line 184: any reason to not declare the `export` in the same line where the components are declared?

src/skeletons.tsx
```diff
@@ -0,0 +1,184 @@
...
+    return (
+        <div
+            className={classNames(animationClasses.animation, styleClass.wrap, rowClass.row)}
+            role="status"
+            aria-hidden={ariaLabel === undefined}
+            aria-busy={ariaLabel !== undefined}
+            aria-label={ariaLabel}
+        >
+            <SkeletonBase height={40} width={40} radius="50%" className={rowClass.circle} />
+            <SkeletonBase width={width} className={rowClass.line} />
```
atabel, line 156: I'd expect this to be a SkeletonCircle + SkeletonLine

src/skeletons.tsx
```diff
@@ -0,0 +1,184 @@
...
+    return (
+        <Stack
+            className={`${animationClasses.animation} ${styleClass.wrap}`}
+            space={16}
+            role="status"
+            aria-hidden={ariaLabel === undefined}
+            aria-busy={ariaLabel !== undefined}
+            aria-label={ariaLabel}
+        >
+            <SkeletonBase />
```
atabel, line 90: I'd expect this to be 3 SkeletonLines

src/skeletons.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+    return (
+        <div
+            className={`${animationClasses.animation} ${styleClass.wrap}`}
+            role="status"
+            aria-hidden={ariaLabel === undefined}
+            aria-busy={ariaLabel !== undefined}
+            aria-label={ariaLabel}
+        >
+            <SkeletonBase></SkeletonBase>
```
atabel, line 70:
```suggestion
            <SkeletonBase/>
```

atabel (commented):
I see some screenshots which are identical, what's the reason?

src/index.tsx
brtbrt: makes sense, alright!

src/skeletons.tsx
brtbrt: no, simply coding routine. I can change it

src/skeletons/skeleton-text.tsx
brtbrt: you should place a defined width on the rectangle in this case, and you'll have the expected behavior

brtbrt: https://mistica-mrgj5x2d2-tuentisre.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8BJAdgGwJZpgAgDMBXDDAdyygBcALPAZwAcBDMGAXmACYAGAXwB8AHTR48SAMoBrGBhhUIaAEowwVZmgDmckWLHMATlmYAZZgCNZ7ISAwRmUHJpu69NGFk00qnAIw9%2BVzEKahprEF4ANzIXUT1Hegs5AEE0LABbZiosRU4CZgx6GD5XAHphOMkZOQVlVXUtHTj9I1MLKxs7BydYvTx3T28-AJLmvBDacKiYkCC8BKSYVIysnLQ8gqLRsXK5qtl5RRU1DW0YOcNjM0sMcK7HLV63Dy8fYH9AsYmwm2mnsQW5hSaUy2VywHyhWKZQqfTE%2BxqR3qpyafUubRud3sD2cszGA1ew0%2BfW%2BUx40X%2B8ywiSBSxBq3BkK2MJESFK6GwuAEIAANCBaDB0jB6AgANogACyEEi1PUBhAAF0%2Bd8RfBRQBmABsPB5vi1AA4FXwgA

brtbrt: https://mistica-mrgj5x2d2-tuentisre.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8BJAdgGwJZpgAgGcAHAQzBgF5gAmABgF88AzAVwwwHUsoAXACwB8AHTR48SAMoBrGBhg8IaAEowwPEmgDmckWLEkATlhIAZEgCNZFISAwQSUHJpu69fGFk18eVAIy0GVzFHAgs5AEE0LABbEh4sRSomEgwCGHpXAHphUXFpWXlFFTUNbRggvENjM0sMa1t7Ry0XXLF3T28-AIzWvBCwmEiYuIS0JJS0nrFskSRM9GxcHJAAGhB%2BGGiYAgQAbRAAWQgANywCdQMQAF01gHdufh34XYBmADZaFd93gA4r%2BiAA

src/skeletons.tsx
brtbrt: that was my very first thought, but then by reiterating I figured out that this approach would be better:

- aria attributes are only defined on the final element, the one that would wrap all the graphical components, whereas SkeletonBase does not have them
- the same for the animation: this way I make sure that all the graphical components that make up a skeleton will be flashing at the same pace. 

let's say I'm seeing the skeletons as a final class, wheresas a BaseSkeleton as an abstract one

src/skeletons/skeleton-text.tsx
yceballost: mmm in my opinion,  this way has a different behavior than the rest of the components and changes the way to build, but If @pladaria and @atabel I will ok

src/__screenshot_tests__/skeletons-screenshot-test.tsx
```diff
@@ -0,0 +1,89 @@
+import {openStoryPage, screen} from '../test-utils';
+
+test.each`
+    givenStoryPageArgs
+    ${{inverse: false, disableAnimation: true}}
+    ${{inverse: true, disableAnimation: true}}
+    ${{inverse: false, size: 60, disableAnimation: true}}
+    ${{inverse: true, size: 60, disableAnimation: true}}
+    ${{inverse: false, size: 100, disableAnimation: true}}
+    ${{inverse: true, size: 100, disableAnimation: true}}
```
pladaria, line 10: is it relevant to screenshot test the `disableAnimation` state?

src/__screenshot_tests__/skeletons-screenshot-test.tsx
```diff
@@ -0,0 +1,89 @@
...
+        await openStoryPage({
+            id: 'components-skeletons-skeleton-circle',
+            device: 'MOBILE_IOS',
+            args: givenStoryPageArgs,
+        });
+
+        const element = await screen.findByTestId('skeleton-circle');
+        expect(await element.screenshot()).toMatchImageSnapshot();
+    }
+);
```
pladaria, line 23: I'd rewrite these tests as:
```
test.each`
    inverse  | size
    ${false} | ${undefined}
    ${true}  | ${undefined}
    ${false} | ${60}
    ${true}  | ${60}
    ...
`(
    `Skeleton circle given inverse $inverse and size of $size`,
    async ({givenStoryPageArgs}) => {
        await openStoryPage({
            id: 'components-skeletons-skeleton-circle',
            device: 'MOBILE_IOS',
            args: {inverse, size, disableAnimation: true},
        });

        const element = await screen.findByTestId('skeleton-circle');
        expect(await element.screenshot()).toMatchImageSnapshot();
    }
);
```

src/skeletons.tsx
```diff
@@ -0,0 +1,184 @@
...
+    const animationClasses = useAnimation({disableAnimation});
+    const styleClass = useSkeletonStyles({width, height: 'content-initial'});
+    const rowClass = useSkeletonRowStyles();
+
+    return (
+        <div
+            className={classNames(animationClasses.animation, styleClass.wrap, rowClass.row)}
+            role="status"
+            aria-hidden={ariaLabel === undefined}
+            aria-busy={ariaLabel !== undefined}
```
pladaria, line 152: shouldn't `aria-busy` be always `true` for skeletons?

src/skeletons.tsx
pladaria, line 70:
```suggestion
            <SkeletonBase />
```

src/skeletons.tsx
```diff
@@ -0,0 +1,184 @@
...
+    height: number | string;
+    width: number | string;
+    isInverse: boolean;
+};
+
+const useSkeletonStyles = createUseStyles(() => ({
+    wrap: {
+        borderRadius: 8,
+        height: ({height = 8}: SkeletonStylesProps) => height,
+        width: ({width = 'content-initial'}) => width,
```
pladaria, line 44: I'm unable to find documentation about `content-initial`

https://developer.mozilla.org/en-US/docs/Web/CSS/width
https://www.w3schools.com/cssref/pr_dim_width.asp

src/__screenshot_tests__/skeletons-screenshot-test.tsx
```diff
@@ -0,0 +1,89 @@
+import {openStoryPage, screen} from '../test-utils';
```
pladaria, line 1: There is a lot of garbage in the `__image_snapshots__` directory. Probably caused by updating the screenshots-tests with titles, etc.

I took your branch and there should be just 17 screenshots:
image

src/skeletons.tsx
pladaria: I see no problem in composing existing skeletons. If the aria-label is not given, the aria-hidden for inner elements would be "true"

you could also pass a role=presentation for child skeletons

src/__screenshot_tests__/skeletons-screenshot-test.tsx
brtbrt: there's some issue with the undefined, which does not make the element render properly for the test. this only happens within the test env, the components themselves accept undefined without any problem. that's why I've used the objet in order to control the undefined values as a missing ptop

src/skeletons.tsx
brtbrt: this means that if we don't pass a label a <SkeletonLine> would be aria-hidden in place of having a semantic element that tells the user "you should wait while we're charging this". I'd avoid this

src/skeletons.tsx
```diff
@@ -0,0 +1,184 @@
...
+    },
+}));
+
+type SkeletonLineProps = {
+    width?: string | number;
+    disableAnimation?: boolean;
+    ariaLabel?: string;
+};
+
+const SkeletonLine = ({
```
pladaria, line 54: please, add the dataAttributes prop to skeleton components (see other components to know how)

src/skeletons.tsx
brtbrt: not really, at least in my mind: if it's hidden, then the component should be totally ignored by the assistant. But if we pass a label, it means we want to give it a semantic meaning, and then we do want to be showed!

yceballost (approved):
gg! 🤟

src/skeletons.tsx
```diff
@@ -0,0 +1,200 @@
...
+        '50%': {
+            opacity: 0.5,
+        },
+
+        '100%': {
+            opacity: 1,
+        },
+    },
+}));
+
```
aweell, line 36: Did you consider apart of providing the possibility via prop disabling the animation by default based on reduced motion preferences?

brtbrt: I've considered it but it's still an open issue AFAIK! I'd say we can proceed with this solution, because the father using these elements can control it, according to the needs (disabling, or other needs)

brtbrt: we've decided to get rid off the disable animation feature completely

atabel (approved):

PR #536, opened 2022-09-14
feat(dialog): allow extra content in web

sergivillar:


src/dialog.tsx
```diff
@@ -168,6 +185,7 @@ const Dialog: React.FC<DialogProps> = (props) => {
                     {message}
                 </Text3>
             </div>
+            {extra && <Box paddingTop={8}>{extra}</Box>}
```
sergivillar, line 188: this is the only UI change

src/dialog.tsx
```diff
@@ -257,6 +275,34 @@ const useNativeDialog = ({
     }, [onAcceptRef, onCancelRef, acceptText, cancelText, showCancel, message, title, renderNative]);
 };
 
+type NativeModalDialogProps = ModalDialogProps & {
+    dialogAcceptButton: string;
+    dialogCancelButton: string;
+};
+
+const NativeModalDialog = (props: NativeModalDialogProps) => {
```
sergivillar, line 283: this is just a Wrapper to the native behaviour

playroom/snippets.tsx
```diff
@@ -1367,9 +1367,29 @@ const alertSnippets = [
   }
 >
   Open two buttons destructive
-</ButtonPrimary>  
+</ButtonPrimary>
       `,
     },
```
yceballost, line 1372: snippet doesn't works in playroom
https://ibit.ly/bQN1

yceballost (changes requested):

src/__stories__/dialog-story.tsx
```diff
@@ -46,6 +47,21 @@ export const Default: StoryComponent = () => (
                 Open two buttons destructive
             </ButtonPrimary>
         </ButtonLayout>
+        <ButtonLayout>
+            <ButtonPrimary
+                onPress={() =>
```
yceballost, line 52: This example should not have cancel button

yceballost (changes requested):

yceballost (approved):
🤟

src/dialog.tsx
```diff
@@ -513,22 +554,27 @@ export default class DialogRoot extends React.Component<DialogRootProps, DialogR
 }
 
 const showDialog =
-    (showCancel = false) =>
+    (showCancel = false, forceNonNative: boolean) =>
```
pladaria, line 557: I prefer an object:
```js
const showDialog = 
    ({showCancel = false, forceNonNative = false}: {showCancel: boolean, forceNonNative: boolean}) => ...
```

This way instead of:
```js
showDialog(true, false)
```
you would read:
```js
showDialog({showCancel: true, forceNonNative: false})
```

src/dialog.tsx
```diff
@@ -272,7 +322,7 @@ const ModalDialog = (props: ModalDialogProps) => {
             `To use @telefonica/mistica components you must instantiate <ThemeContextProvider> as their parent.`
         );
     }
-    const renderNative = isWebViewBridgeAvailable();
+    const renderNative = !props.forceNonNative && isWebViewBridgeAvailable();
```
pladaria, line 325: double negation, perhaps you could rename `forceNonNative` to `forceWeb`

pladaria (approved):

src/dialog.tsx
```diff
@@ -257,6 +279,34 @@ const useNativeDialog = ({
     }, [onAcceptRef, onCancelRef, acceptText, cancelText, showCancel, message, title, renderNative]);
 };
 
+type NativeModalDialogProps = ModalDialogProps & {
+    dialogAcceptButton: string;
+    dialogCancelButton: string;
+};
+
+const NativeModalDialog = (props: NativeModalDialogProps) => {
+    const renderNative = isWebViewBridgeAvailable();
```
atabel, line 288: why is this condition repeated now?

src/dialog.tsx
```diff
@@ -312,34 +362,19 @@ const ModalDialog = (props: ModalDialogProps) => {
...
         if (!animationsSupported(platformOverrides)) {
             addKeyDownListener();
         }
 
-        if (
-            (renderNative || !animationsSupported(platformOverrides)) &&
-            props.isClosing &&
-            props.onCloseTransitionEnd
-        ) {
+        if (!animationsSupported(platformOverrides) && props.isClosing && props.onCloseTransitionEnd) {
```
atabel, line 370: why this change?

src/dialog.tsx
sergivillar: true, not needed here, because will be always `true`, removed

src/dialog.tsx
sergivillar: just move the `renderNative` to new component, but as renderNative is not needed there, this could be restored as it was too

atabel (approved):

atabel:
I miss an screenshot for the new case

atabel (approved):

PR #540, opened 2022-09-16
test | use mistica icons chevron in lists

yceballost:


yceballost:
we can close this draft too. The generated link is enough

Issue #546, opened 2022-09-27
Make components to adapt to the container width instead of screen size, and using this capability on carousel's arrows


wyunreal:
Problem:
As components are using the screen size to adapt to different viewports, if you use a grid layout, the components on the columns won't adapt well to the available space.
An example of this can be shown on the carousel: the arrows aren't positioned correctly.

Solution: Create a context to pass down the available space components have, in terms of small/med/wide desktop column, mobile or tablet.

A considered alternative could be just passing the exact column width, but discarded due to not needed.

sergivillar:
@wyunreal  could this issue be closed? as the related PR seems merged

PR #547, opened 2022-09-27
feat(Carousel): improve styles inside GridLayout columns

wyunreal:
issue: https://github.com/Telefonica/mistica-web/issues/546

pending:
tests for the new hook,
tests for the carousels (I've changed it and no test failed, so I think I need to add tests).

Some result screenshots:

Small desktop:
carousel-on-small-desktop

Largue desktop:
carousels-on-largue-desktop

Mobile:
image

Tablet:
image


src/carousel.tsx
```diff
@@ -101,6 +101,15 @@ const arrowButtonStyle = (theme: Theme) => ({
...
 
+const arrowButtonSeparation = (containerType: ContainerType, isLargueDesktop: boolean) => {
+    switch (containerType) {
+        case 'desktop-wide-column':
+            return isLargueDesktop ? -(24 + arrowButtonSize) : -arrowButtonSize / 2;
+        default:
+            return -arrowButtonSize / 2;
+    }
+};
+
```
wyunreal, line 112: @yceballost we need your opinion here, as this is not yet defined by design rules, this can change depending on your opinion (see screenshots attached).

src/grid-layout.tsx
```diff
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import ContainerSizeContext from './container-type-context';
```
wyunreal, line 2: @yceballost we need definition on which size (small, mid, wide should be each column on the template, I assumed small from 1 to 5, mid from 6 to 9 and wide just 10 and when not inside a grid layout.

yceballost: we can go with this by the moment and we will review when we can work on it

yceballost (changes requested):

flow-defs/screen-size-context.js.flow
```diff
@@ -7,6 +7,7 @@ export type ScreenSizeContextType = {
   isTabletOrBigger: boolean,
   isTabletOrSmaller: boolean,
   isDesktopOrBigger: boolean,
+  isLargueDesktop: boolean,
```
Winde, line 10: typo in Large

src/hooks.tsx
```diff
@@ -185,3 +187,14 @@ export const useIsInViewport = (
 
     return isInViewport;
 };
+
+export const useContainerType = (): ContainerType => {
+    const containerType = React.useContext(ContainerTypeContext);
+    if (!containerType) {
+        throw Error(
```
Winde, line 194: Are we 100% sure that we want this to throw an error, there's no possibility of using one of the mistica atoms that use containerType without having this context?

src/responsive-layout.tsx
```diff
@@ -64,16 +65,20 @@ const ResponsiveLayout: React.FC<Props> = ({children, className, fullWidth, data
 
     const sideMargin = isMobile ? MOBILE_SIDE_MARGIN : isTablet ? TABLET_SIDE_MARGIN : null;
 
+    const containerType = isMobile ? 'mobile-column' : isTablet ? 'tablet-column' : 'desktop-wide-column';
```
Winde, line 68: hard to read ternary inside other ternary. 

I'd suggest to reword this to not chain them

src/screen-size-context-provider.tsx
```diff
@@ -27,8 +27,15 @@ const ScreenSizeContextProvider: React.FC<Props> = ({children}) => {
...
             tabletOrSmaller: stripMedia(theme.mq.tabletOrSmaller),
+            largueDesktop: stripMedia(theme.mq.largeDesktop),
         }),
-        [theme.mq.mobile, theme.mq.tablet, theme.mq.tabletOrBigger, theme.mq.tabletOrSmaller]
+        [
+            theme.mq.mobile,
+            theme.mq.tablet,
+            theme.mq.tabletOrBigger,
+            theme.mq.tabletOrSmaller,
+            theme.mq.largeDesktop,
```
Winde, line 37: is this a new mediaquery? shouldn't it be a new definition part of this PR?

Winde (approved):

src/hooks.tsx
wyunreal: mmm, the error is raised only when the hook is used on a component not placed on a responsive layout, nor a grid layout. By now, only the carousels are using it.

Question @yceballost: if not in a responsive layout nor grid layout, what value should arrive to the component via this hook ? maybe depending on the screen media queries ? I can do that if you think thats the way to go:
- mobile for mobile screen
- tablet for tablet
- desktop and wide desktop for desktop and large desktop respectively. 

WDYT ?

src/responsive-layout.tsx
wyunreal: I just copy pasted the pattern being seen on the above line ;) but you're right, its difficult to read.

src/screen-size-context-provider.tsx
wyunreal: No no, the mediaquery already exists, I'm just adding it to this hook to make it available to components, as the carousel requires it.

src/hooks.tsx
yceballost: for me it is ok!

In this way, the actual behavior keeps the same behavior, only changing when it is on a responsive or grid layout and the container is less than the mq defined, right?

yceballost (approved):
After some test, i didn't see anything weird! Thanks W <3

src/screen-size-context-provider.tsx
```diff
@@ -27,8 +27,15 @@ const ScreenSizeContextProvider: React.FC<Props> = ({children}) => {
             tablet: stripMedia(theme.mq.tablet),
             tabletOrBigger: stripMedia(theme.mq.tabletOrBigger),
             tabletOrSmaller: stripMedia(theme.mq.tabletOrSmaller),
+            largueDesktop: stripMedia(theme.mq.largeDesktop),
```
pladaria, line 30: typo: largue

src/carousel.tsx
```diff
@@ -114,22 +128,12 @@ const useStyles = createUseStyles((theme) => ({
         zIndex: 1,
         top: `calc(50% - ${arrowButtonSize / 2}px)`,
         '&.prev': {
-            left: -arrowButtonSize / 2,
-            [theme.mq.tabletOrSmaller]: {
```
atabel, line 118: I think I'd keep mediaqueries and just apply overrides in desktop when inside medium/small a grid column

src/responsive-layout.tsx
```diff
@@ -64,16 +65,20 @@ const ResponsiveLayout: React.FC<Props> = ({children, className, fullWidth, data
...
+
     return (
         <ResponsiveLayoutMarginContext.Provider value={sideMargin}>
-            <div
-                className={classnames(classes.container, className)}
-                {...getPrefixedDataAttributes(dataAttributes)}
-            >
-                <div className={classnames(classes.layout, {[classes.layoutFullWidth]: fullWidth})}>
-                    {children}
+            <ContainerTypeContext.Provider value={containerType}>
```
atabel, line 72: I wouldn't apply this to responsive layout, only to grid layout

src/container-type-context.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
+import {useScreenSize} from './hooks';
+
+import type {ContainerType} from './utils/types';
+
+const ContainerTypeContext = React.createContext<ContainerType | null>(null);
```
atabel, line 6: I'd call it `ContainerWidthContext` with values `'large' | 'medium' | 'small'` and only listen for them in desktop viewports. I won't mix screen size with container width

src/container-type-context.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export const useContainerType = (): ContainerType => {
+    const containerType = React.useContext(ContainerTypeContext);
+    const {isTablet, isDesktopOrBigger, isLargeDesktop} = useScreenSize();
```
atabel, line 10: I won't mix screen size with container width, see my previous comment

src/utils/types.tsx
```diff
@@ -10,3 +10,10 @@ export type IconProps = {
     className?: string;
     style?: React.CSSProperties;
 };
+
+export type ContainerType =
+    | 'desktop-small-column'
+    | 'desktop-mid-column'
```
pladaria, line 16: "medium" instead of "mid" ?

src/utils/types.tsx
```diff
@@ -10,3 +10,10 @@ export type IconProps = {
...
+    | 'desktop-wide-column'
```
pladaria, line 17: "large" instead of "wide" ?

pladaria (approved):

yceballost:
> image @yceballost @wyunreal, is this expected?

yes, talked with @wyunreal and we will improve when we can work on it later. Wiljan promised me that he will make the changes :P

src/carousel.tsx
```diff
@@ -101,6 +102,19 @@ const arrowButtonStyle = (theme: Theme) => ({
     },
 });
 
+const arrowButtonSeparation = (containerType: ContainerType, isLargeDesktop: boolean, sideMargin: number) => {
```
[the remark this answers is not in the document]
wyunreal: Yes, I've decided to leave the balls the same (depending on the screen size) to let them feel the same (It would be weird to see some balls with different styles at the same time.
IMHO, we should not change styles of the bullets depending on the column.

src/carousel.tsx
```diff
@@ -101,6 +102,19 @@ const arrowButtonStyle = (theme: Theme) => ({
...
+    console.log(containerType);
```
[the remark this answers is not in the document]
wyunreal: yes !! removed !!

src/container-type-context.tsx
wyunreal: And what about the grid layout when its displayed in a mobile or tablet ? In these cases, the expected behavior is to be displayed as if the grid won't exists.

src/container-type-context.tsx
wyunreal: I've answered, I agree we should not mix, thats why I'm not mixing at all XD, I've created the Columns alike enum to just denote a different concept: not the same (conceptually) a mobile screen than a mobile column

src/responsive-layout.tsx
wyunreal: Appliyng to responsive layout we can also enable the components to use the Context on all devices, simplifying the development of the components (they can then take into account only the container type).

src/screen-size-context-provider.tsx
wyunreal: changed, thanks ;)

src/utils/types.tsx
wyunreal: yes, will change !!

atabel (approved):

PR #548, opened 2022-09-29
Move header height to context

sergivillar:
# Proposal to move header height to context

Why? First because we have two sources of truth  for mobile height. Second reason, because it will be nice to have also desktop height in that context (to use in zeus, for example)


src/theme.tsx
```diff
@@ -228,7 +229,7 @@ export type ThemeConfig = {
         logEvent: (trackingEvent: TrackingEvent) => Promise<void>;
         eventFormat?: EventFormat;
     };
-    dimensions?: {headerMobileHeight: number};
+    dimensions?: {headerMobileHeight: number; headerDesktopHeight: number};
```
pladaria, line 232: as commented in Teams, I wouldn't allow to configure the `headerDesktopHeight`. Use the internally defined value and expose it via context, so you can use it if needed

atabel (approved):

pladaria (approved):

PR #550, opened 2022-09-30
feat(Tooltip): open on hover in desktop

sergivillar:


atabel (commented):
In desktop, I'd apply a timeout when the user hovers the target until the tooltip is shown.

Please include someone from Design Core to review these behavior changes

sergivillar:
> In desktop, I'd apply a timeout when the user hovers the target until the tooltip is shown.
> 
> Please include someone from Design Core to review these behavior changes

@HectorSancho told me to avoid that delay, but let's wait for his opinion

pladaria (approved):

yceballost:
missing slot feature in tooltip
image

src/tooltip.tsx
```diff
@@ -97,12 +104,152 @@ const useStyles = createUseStyles((theme) => {
     };
```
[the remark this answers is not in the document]
sergivillar: by default the width is 340px in desktop, do you want to change that?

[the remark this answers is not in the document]
yceballost: in desktop would be adapted to the content width up to size 343px
image

src/tooltip.tsx
```diff
@@ -282,10 +434,10 @@ const Tooltip: React.FC<Props> = ({children, description, target, title, targetL
                                 {description && <p className={classes.description}>{description}</p>}
```
yceballost, line 469: tooltip have to use text-preset-2/regular instead of `<p>`

src/__stories__/tooltip-story.tsx
```diff
@@ -65,9 +65,14 @@ export const Default: StoryComponent = () => {
                     target={<Text2 regular>Tooltip with link</Text2>}
                     position="top"
                     description={description}
-                >
-                    <a href="#whatever">Ir a la web</a>
-                </Tooltip>
+                    extra={<a href="#whatever">Ir a la web</a>}
```
yceballost, line 68: better use textLink for this example 👀

yceballost (approved):
nice!! 😍

src/tooltip.tsx
```diff
@@ -478,12 +454,12 @@ const Tooltip: React.FC<Props> = ({children, description, target, title, targetL
...
-                                {title && <p className={classes.title}>{title}</p>}
-                                {description && <p className={classes.description}>{description}</p>}
-                            </>
+                            <Stack space={4}>
+                                {title && <Text2 medium>{title}</Text2>}
+                                {description && <Text2 regular>{description}</Text2>}
+                            </Stack>
                         )}
-                        {children}
+                        {extra || children}
```
atabel, line 462: why do we need `extra`? just use `children`

src/tooltip.tsx
```diff
@@ -75,48 +85,150 @@ const useStyles = createUseStyles((theme) => {
...
 
-const Tooltip: React.FC<Props> = ({children, description, target, title, targetLabel, ...rest}) => {
+const Tooltip: React.FC<Props> = ({
+    children,
+    extra,
+    description,
+    target,
+    title,
+    targetLabel,
+    delay = 0,
```
atabel, line 222: isn't `delay` a boolean?
it should have delay by default

src/tooltip.tsx
```diff
@@ -1,20 +1,30 @@
...
+const distanceToTarget = 4 + arrowSize;
 const marginLeftRightMobile = 16;
 const defaultWidthDesktop = 340;
 const arrowWrapperWidth = arrowSize * 2;
 const arrowWrapperHeight = arrowSize;
 
+const transitionDurationMs = 500;
+const animationMovement = 12;
+const animationTiming = 'cubic-bezier(0.215, 0.61, 0.355, 1)';
+const defaultShowTooltipDelayMs = 2000;
```
atabel, line 24: too much, I'd say something like 500ms should be fine

src/tooltip.tsx
yceballost: looking for feature parity

src/tooltip.tsx
yceballost: we can change it to 500ms

But when you hover some elements for example, PR Chip in this page, the delay is more than 500ms, I think is a native delay, just comment
image

src/tooltip.tsx
atabel: it doesn't make sense to have both, if we add extra we should remove children. But removing children is a breaking change, so if we prefer extra, then mark children as deprecated

atabel (approved):

src/__stories__/tooltip-story.tsx
```diff
@@ -62,12 +62,26 @@ export const Default: StoryComponent = () => {
...
                     position="top"
                     description={description}
-                >
-                    <a href="#whatever">Ir a la web</a>
-                </Tooltip>
+                    extra={<Placeholder />}
+                />
+                <Tooltip
+                    targetLabel="help text"
+                    target={<Text2 regular>Tooltip with delay</Text2>}
```
atabel, line 81: does this example makes sense? all the Tooltips have delay now

sergivillar: changed

PR #552, opened 2022-10-06
support custom color

sergivillar:
This PR tries to fix an issue found in the fragment part of NavigationBar

https://mistica-web.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8AFATjAbgSxgdwBUIIAbAZwAIB6APgB0A7JAQUwEMAXNtCs7ALxgBeYAGYATAF9q9BoyQBZNtgYA5NjgDmnbBAYAhbowq8YYDroZkRAbTogAylzQd7AGgr3mYSAFcGriAe9gCiAB4ADiQQGO6ejr4RETGBALoAdAC2bBEAFLkWHCQwHthQYQCUFEI0FLnAxiYUhcVujSZ66DBkZPB1VTWmHE6cMLn2KrBhcWWVbQwmkhUVko1kMMXmMFAAkgxTIpoww1wcYxP7MNNBFAAMK41o2JoAFhwi7RRI6lo6eoZobwWPQAcTQEESsiaTW%2BGmefwM3CBlgonQwPREuQGtWAkmk3GwbAAtCQ2AAjDZCewAeQiMAWEXBADNsMV7FDoTDWJxuLwBMIxFIZJ8TEgqD94cDEYDzJYOaLxXDtFKAcjQeDIY1VgxaIwgiAOC8YJluggbCAFBAcGRnCBUm4QHgyoayGaJLc3ABGW7iAAsqUkQA

In zeus we need to draw an Avatar like this
Screen Shot 2022-10-06 at 12 39 02

Thing that is not possible with the current implementation and it should be as talked with @HectorSancho


src/navigation-bar.tsx
```diff
@@ -581,7 +581,10 @@ export const NavigationBarAction: React.FC<NavigationBarActionProps> = ({childre
     const isInverse = useIsInverseVariant();
     const classes = useNavigationBarActionStyles({isInverse});
     return (
-        <Touchable {...touchableProps} className={classes.touchable}>
+        <Touchable
+            {...touchableProps}
+            className={touchableProps.className ? touchableProps.className : classes.touchable}
+        >
```
atabel, line 587: this should be equivalent, but please, consider my other comment first:
```suggestion
        <Touchable className={classes.touchable} {...touchableProps}>
```

src/navigation-bar.tsx
```diff
@@ -581,7 +581,10 @@ export const NavigationBarAction: React.FC<NavigationBarActionProps> = ({childre
     const isInverse = useIsInverseVariant();
     const classes = useNavigationBarActionStyles({isInverse});
     return (
-        <Touchable {...touchableProps} className={classes.touchable}>
```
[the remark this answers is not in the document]
atabel: I think that's intended, because `NavigationBarAction` tries to follow the specs regarding actions with icon in navbars.
I'd say if you don't want these styles, you'd better use a custom component built with a `Touchable` with your own styles, instead of using `NavigationBarAction`, wdyt?

pladaria, line 584: I think the problem is here:
```js
        '& svg': {
            color: ({isInverse}) => (isInverse ? theme.colors.inverse : theme.colors.neutralHigh),
        },
```
The avatar icon color shouldn't be overridden

atabel (approved):

pladaria (approved):

PR #554, opened 2022-10-10
feat(Dialog): design improvements

sergivillar:
https://jira.tid.es/browse/WEB-615


src/dialog.tsx
```diff
@@ -191,22 +200,26 @@ const Dialog: React.FC<DialogProps> = (props) => {
...
-                        <ButtonSecondary
-                            tabIndex={2} // eslint-disable-line jsx-a11y/tabindex-no-positive
-                            onPress={handleCancel}
-                            dataAttributes={{testid: 'dialog-cancel-button'}}
-                        >
-                            {cancelText}
-                        </ButtonSecondary>
-                    )}
-                </ButtonLayout>
+                <ButtonGroup
```
sergivillar, line 203: There's one missing thing in this PR, and is the need to support a link (up to 3 actions in dialog), what will be the best way to support this? adding all the needed props of link (to, href, newTab....) ? add a prop that receive a ButtonLink component? this dialog component works only in web side (no native dialog) so.,.. do we really need to support this new link?

src/dialog.tsx
```diff
@@ -124,26 +124,35 @@ interface BaseDialogProps {
     title?: string;
```
yceballost, line 124: title in alert & confirm dialog have different text-preset than Dialog
image

yceballost (changes requested):

src/dialog.tsx
atabel: why?

atabel: hmmm... that argument sounds weak to me. I'd say it's way better to have consistence across dialogs in the same platform (web) than having consistence between iOS, Android and web

pladaria (approved):

src/dialog.tsx
yceballost: I think is better to break the consistency between this type of dialogs because the usage is very different. 

Confirm and alerts it's for immediate actions, however, dialogs could have much plane information without a required action. The moment for the user is very different.

src/dialog.tsx
```diff
@@ -10,14 +10,16 @@ import {isWebViewBridgeAvailable, nativeConfirm, nativeAlert} from '@tef-novum/w
...
+import {Text5, Text4, Text3} from './text';
 import {ESC} from './utils/key-codes';
 import Box from './box';
 import {isOldChrome, isRunningAcceptanceTest} from './utils/platform';
 import {useSetModalState} from './modal-context-provider';
+import {pxToRem} from './utils/css';
 import Stack from './stack';
 
 import type {Theme} from './theme';
+import type {RendersNullableElement} from './utils/types';
```
yceballost, line 23: I found a weird space above the title in confirm and alerts in mobile 
image

in specs 
image

I think the cause of this is when the modal doesn't have dismiss, you don't need to include this extra space

yceballost (changes requested):

sergivillar:
> In this example, the subtitle is missing. Captura de Pantalla 2022-10-20 a las 15 25 02
> 
> Take a look at this specs screenshot: Captura de Pantalla 2022-10-20 a las 15 35 43

this examples were created by @yceballost, maybe he dont want to include that for some reason

yceballost:
> > In this example, the subtitle is missing. Captura de Pantalla 2022-10-20 a las 15 25 02
> > Take a look at this specs screenshot: Captura de Pantalla 2022-10-20 a las 15 35 43
> 
> this examples were created by @yceballost, maybe he dont want to include that for some reason

I forgot completely xD. Nice catch @AnaMontes11, thanks

atabel (approved):

AnaMontes11 (approved):
Great job! ✨

PR #556, opened 2022-10-11
WEB-471 added google analytics to the storybook

brtbrt:


yceballost (approved):
should we add a gdpr warning in storybook?
https://github.com/storybookjs/addon-google-analytics/issues/3#issuecomment-1145100383

atabel:
> should we add a gdpr warning in storybook? [storybookjs/addon-google-analytics#3 (comment)](https://github.com/storybookjs/addon-google-analytics/issues/3#issuecomment-1145100383)

good point... I guess we should, but does this storybook addon allows to load conditionally?

PR #560, opened 2022-10-24
feat(Carousel): initialActiveItem prop

wyunreal:


src/carousel.tsx
```diff
@@ -353,6 +355,11 @@ const BaseCarousel: React.FC<BaseCarouselProps> = ({
     const showNextArrow = scrollRight !== 0;
     const showPrevArrow = scrollLeft !== 0;
 
+    const initialItemInfo = React.useRef({
+        initialItem: 0,
+        itemsPerPage: itemsPerPageFloor,
+    });
+
```
wyunreal, line 362: required to assure the scroll is updated only when the prop is changed, but also when the items per page changes (to avoid leaving the carousel on a corrupt state).

SergioMorchon (approved):

kydorn (approved):

src/carousel.tsx
```diff
@@ -457,6 +464,24 @@ const BaseCarousel: React.FC<BaseCarouselProps> = ({
...
+            initialItemInfo.current.itemsPerPage = itemsPerPageFloor;
+        }
+    }, [
+        initialActiveItem,
+        itemsPerPageFloor,
+        goToPage,
+        initialItemInfo.current.initialItem,
+        initialItemInfo.current.initialItem,
+    ]);
+
```
pladaria, line 484: I'm probably missing something but if you want to react to the `initialActiveItem` change, the ref is not needed, you just need an effect like:
```js
React.useEffect(() => {
    goToPage(Math.floor(initialActiveItem / itemsPerPageFloor));
}, [initialActiveItem, goToPage, itemsPerPageFloor]);
```

But I think you shouldn't react to `initialActiveItem` changes. This value should be used only for the initial render and ignore changes. Then you should use a ref and use its value to set the initial position. Don't react to changes.

To update the storybook carousel when changing the `arg`, perhaps you want to set a `key` value to the carousel to remount it when the `initialActiveItem` changes

src/carousel.tsx
```diff
@@ -457,6 +464,24 @@ const BaseCarousel: React.FC<BaseCarouselProps> = ({
...
+            initialItemInfo.current.initialItem = initialActiveItem;
+            initialItemInfo.current.itemsPerPage = itemsPerPageFloor;
+        }
+    }, [
+        initialActiveItem,
+        itemsPerPageFloor,
+        goToPage,
+        initialItemInfo.current.initialItem,
+        initialItemInfo.current.initialItem,
+    ]);
```
atabel, line 483:
```suggestion
    }, [initialActiveItem, itemsPerPageFloor, goToPage]);
```
You don't need refs in `useEffect` dependencies

atabel (commented):
I'd include a screenshot test to cover this feature

src/carousel.tsx
wyunreal: I need to react to itemsPerPageFloor because the carousel renders with 1 in this variable the first time and then it re'renders with the correct value. so I need to add dependencies array, so I need to add all dependencies to make the warning disappear.
Also if the prop changes, its desirable to move to the page the prop reffers to (in account there is a use case in the header of the invoicing page (if the selected invoice changes, the carousel must go to the page in where the invoice is present).

src/__screenshot_tests__/carousel-screenshot-test.tsx
```diff
@@ -17,6 +17,16 @@ test('Carousel mobile', async () => {
     expect(await page.screenshot()).toMatchImageSnapshot();
 });
 
+test('Carousel mobile selecting second page initially', async () => {
```
atabel, line 20: I'd rename the test to
```suggestion
test('Carousel mobile with initialActiveItem=1', async () => {
```
Because we can select the active item, not the page.
Same for the other tests

src/carousel.tsx
```diff
@@ -311,6 +312,7 @@ const BaseCarousel: React.FC<BaseCarouselProps> = ({
     itemClassName,
     withBullets,
     renderBullets,
+    initialActiveItem,
```
pladaria, line 315: if you are reacting to changes of this prop, don't call it `initialActiveItem`. `activeItem` would be a better name because this is not the "initial value", it is the current one.

src/carousel.tsx
```diff
@@ -457,6 +465,20 @@ const BaseCarousel: React.FC<BaseCarouselProps> = ({
 
     const shouldAutoplay = useShouldAutoplay(!!autoplay, carouselRef);
 
+    React.useEffect(() => {
+        if (
+            initialActiveItem &&
```
pladaria, line 470: `0` is a valid value. But this logic discards it.

src/carousel.tsx
```diff
@@ -457,6 +465,20 @@ const BaseCarousel: React.FC<BaseCarouselProps> = ({
...
+            (initialActiveItem !== initialItemInfo.current.initialItem ||
+                itemsPerPageFloor !== initialItemInfo.current.itemsPerPage ||
+                itemScrollPositions.length !== initialItemInfo.current.scrollPositionsLength)
+        ) {
+            goToPage(Math.floor(initialActiveItem / itemsPerPageFloor));
+            initialItemInfo.current.initialItem = initialActiveItem;
+            initialItemInfo.current.itemsPerPage = itemsPerPageFloor;
+            initialItemInfo.current.scrollPositionsLength = itemScrollPositions.length;
+        }
+    }, [initialActiveItem, itemsPerPageFloor, goToPage, itemScrollPositions.length]);
```
pladaria, line 480: I think this could be replaced by:
```
React.useEffect(() => {
    if (initialActiveItem !== undefined) {
        goToPage(Math.floor(initialActiveItem / itemsPerPageFloor));
    }
}, [initialActiveItem, goToPage, itemsPerPageFloor]);
```

But you also need to memoize the `pagesScrollPositions` value, because currently gets calculated in each render and `goToPage` callback depends on this value
```diff
-    const pagesScrollPositions = calcPagesScrollPositions(itemScrollPositions, pagesCount);
+    const pagesScrollPositions = React.useMemo(
+        () => calcPagesScrollPositions(itemScrollPositions, pagesCount),
+        [itemScrollPositions, pagesCount]
+    );
```

I've committed both changes to this branch `pladaria/carousel` and it seems to work as expected

src/carousel.tsx
wyunreal: its the initial,, because the carousel will be allowed to change the slide without requiting the prop to change (by swiping), thats where the initial word comes from. IMHO the naming is correct.

src/carousel.tsx
wyunreal: you right, will fix it (I was assuming this logic was safe because the initial slide is always 0, but if changing the prop from non zero to zero will not trigger the effect. I'll change this.

src/carousel.tsx
wyunreal: Will test this approach, coded something similar at the first attempt, and the carousel started to behave very bad.

wyunreal: It works perfectly, I think what was missing in my first approach was the useMemo

src/carousel.tsx
wyunreal: Hook changed to what you suggested in the other comment ;)

src/carousel.tsx
wyunreal: fixed

atabel (approved):

pladaria (approved):

PR #561, opened 2022-10-25
feat(Vanilla Extract): as a future replacement of jss

atabel:
### To do:
- [x] Make storybook work
- [x] Make playroom work
- [x] Setup vanilla extract build and generate `mistica.css` file
- [x] Make jest unit test work with vanilla extract
- [x] Configure eslint import/extensions rule to work with `.css.ts` extensions
	- configured locally, but we should consider moving that config to `@telefonica/eslint-config`
- [x] Vendor prefixes
- [x] Automatic `box-sizing: border-box`
- [x] Figure out the best way to work with media queries
	-  media queries are now static and not configurable. I've deprecated `ThemeConfig['mediaQueries']` and will be removed in future versions
	- We were using `ThemeConfig['mediaQueries']` configuration in Playroom for two things: override `desktopOrTabletMinHeight` to 0 to not fallback to mobile break point when the bottom code panel is shown, and for the `PreviewTools` `forceMobile` and `forceDesktop` props. Those props are heavily used in Mistica docs in brand factory site. To support these use cases the solution I've found is making two additional Playroom static builds served in `/playroom-desktop` and `/playroom-mobile` which use custom media query breakpoints (see `playroom.config.js` changes). The downside is that this will increase the preview deployments time. A future improvement could be to only make these additional playroom builds when deploying to prod, and skip them in PR previews.
- [x] Make ssr tests work

### Nice to have:
- [ ] Make storybook run with vite instead of webpack?

## Release plan
We can release a first version with all the setup and then migrate component by component in future PRs

### Major version?
Some breaking changes:
- Once we remove jss completely from mistica we'll need to remove the `createUseStyles` function, so we'll need a major version for sure. But this isn't done in this PR yet and we can do it in future PRs once we have migrated all components
- To use components that use vanilla extract, the app that uses mistica needs to include/serve the new `css/mistica.css` file. This is a breaking change. We could find a way to mitigate this by inlining the css in the js or something. It isn't the more performance solution but could be a good transient solution that doesn't force us to release a major version yet

### The plan:
- Merge this branch in a first minor version:
	- the generated Vanilla Extract css will be injected in js and rendered in a `<style>` tag by `ThemeContextProvider`.
	- Deprecate `ThemeConfig['mediaQueries']` (see details above). 
	- Deprecate `createUseStyles`
- In future PRs, migrate all the components to VE
- Write docs with a migration path for apps using `createUseStyles` from mistica
- In a major release, remove JSS from mistica.


.gitignore
```diff
@@ -18,6 +18,7 @@ storybook-static/
 .vercel
 src/generated/mistica-icons/index.tsx.txt
 reports
+css/mistica.css
```
atabel, line 21: this file is generated by `yarn build`

src/avatar.tsx
```diff
@@ -99,13 +80,20 @@ const Avatar = ({
 
     return (
         <Badge value={badgeValue} top={badgePosition} right={badgePosition}>
-            <div className={classes.avatar} role="img" aria-label={ariaLabel ?? initials}>
+            <div
+                className={classes.avatar}
+                role="img"
+                aria-label={ariaLabel ?? initials}
+                style={{width: size, height: size, color: textColor, background: backgroundColor}}
```
sergivillar, line 87: so now we will use more and more inline styles to support things like 'isInverse' colors right? or could we create a common set of styles with sprinkles for that?

src/sprinkles.css.ts
```diff
@@ -0,0 +1,57 @@
...
+
+const commonProperties = defineProperties({
+    properties: {
+        position: ['relative', 'absolute', 'fixed', 'static'],
+        display: ['none', 'flex', 'inline-flex', 'block', 'inline', 'inline-block'],
+        flexDirection: ['row', 'column'],
+        justifyContent: ['stretch', 'flex-start', 'center', 'flex-end', 'space-around', 'space-between'],
+        alignItems: ['stretch', 'flex-start', 'center', 'flex-end'],
+        flexShrink: [0],
+        flexGrow: [0, 1],
```
sergivillar, line 28: algo `flex: 1`?

src/sprinkles.css.ts
```diff
@@ -0,0 +1,57 @@
...
+    properties: {
+        color: colors,
+        background: colors,
+        backgroundColor: colors,
+    },
+});
+
+const sizes = [0, 2, 4, 8, 12, 16, 24, 32, 40, 48, 56, 64, 72, 80] as const;
+
+const commonProperties = defineProperties({
```
sergivillar, line 20: from the documentation

```
 conditions: {
    mobile: {},
    tablet: { '@media': 'screen and (min-width: 768px)' },
    desktop: { '@media': 'screen and (min-width: 1024px)' }
  }
```
  
  i think this is also interesting

src/theme-context-provider.tsx
```diff
@@ -161,6 +164,8 @@ const ThemeContextProvider: React.FC<Props> = ({theme, children, providerId}) =>
         };
     }, [theme, isOsDarkModeEnabled]);
 
+    const themeStyle = `:root {${assignInlineVars(vars, {colors: contextTheme.colors})}}`;
```
sergivillar, line 167: is this needed in snapshot tests? there is a hash in this styles that will change in each css change? maybe not needed for tests snapshots?

src/sprinkles.css.ts
atabel: I still need to figure out how to work with mediaqueries, because until now mistica media queries have been configurable with context, but I don't know how to make them configurable and at the same time be able to use them in vanilla extract / sprinkles. (css doesn't allow variables `@media` :disappointed: )

src/theme-context-provider.tsx
atabel: maybe we could start thinking if those test make sense today or we should migrate them to screenshot tests

src/avatar.tsx
atabel: we can use sprinkles when possible (as an alternative to inline styles), but only when dynamic values are available in our sprinkles definitions. For example, if a component receives a prop `color`, we can use it with sprinkles if the color is one of the skin, but if the `color` prop accepts any string then we need to use inline styles

sergivillar:
really nice, two questions

- how vanilla-extract solve the issue of hash classnames generation in SSR? we had issues with this in JSS
- what will happen with the currently createUserStyles?

src/sprinkles.css.ts
atabel: I've added a TODO to the PR list

atabel:
> really nice, two questions
> 
> * how vanilla-extract solve the issue of hash classnames generation in SSR? we had issues with this in JSS

Classnames are generated at build time, so there should never be hydration missmatches

> * what will happen with the currently createUserStyles?

As said in "Release plan", once we migrate all components to VS, we'll remove JSS from mistica, so the apps using mistica which were using `createUseStyles` will need to use JSS directly or migrate to other style solution.

.storybook/main.js
```diff
@@ -17,9 +19,49 @@ const replaceBabelWithSwc = (config) => {
...
+                ? rule.test.some((exp) => exp.test('random.css'))
+                : rule.test.test('random.css'))
+        ) {
+            const previousExclude = rule.exclude || [];
+            rule.exclude = [
+                ...(Array.isArray(previousExclude) ? previousExclude : [previousExclude]),
+                /\.vanilla\.css$/,
+            ];
+        }
+    });
```
pladaria, line 36: can you add a comment explaining why is this needed?

.eslintrc.yml
```diff
@@ -3,3 +3,5 @@ extends:
 overrides:
     - files: ['*.js']
       parser: espree
+rules:
+    import/extensions: [error, {js: never, json: always, css: always}]
```
brtbrt, line 7: why JSON?

postcss.config.js
```diff
@@ -0,0 +1,4 @@
+// vite automatically apply postcss to css files, like the generated mistica.css file from vanilla extract.
```
pladaria, line 1: automatically applies

pladaria (approved):

src/__tests__/touchable-test.tsx
```diff
@@ -38,6 +38,9 @@ test('<a> element is rendered when "to" prop is used and no Link component injec
...
+          </style>
```
brtbrt, line 43: 🤔 this could potentially be duplicated more and more in the future. I'd extract it to a constant or something more maintainable

sergivillar (approved):

brtbrt (approved):

brtbrt (approved):

src/__tests__/touchable-test.tsx
atabel: what do you mean? this is an inline snapshot, it's autogenerated by jest.
Anyway, I'm going to remove this from tests

src/theme-context-provider.tsx
atabel: I've finally excluded this from tests, as Pedro suggested

.storybook/main.js
atabel: added

.eslintrc.yml
atabel: just copy&pasted from https://github.com/Telefonica/js-toolbox/blob/master/packages/eslint-config/import.yml#L5 and added the css extension

src/__tests__/touchable-test.tsx
atabel: done

src/snackbar.css.ts
```diff
@@ -0,0 +1,92 @@
...
+        minWidth: 360,
+        minHeight: 48,
+        padding: `14px 16px`,
+        bottom: 24,
+        zIndex: 1000, // above anything
+        opacity: 0,
+        transform: 'translateY(100px)',
+        transition: `transform ${TRANSITION_TIME_IN_MS}ms ease-in-out, opacity ${TRANSITION_TIME_IN_MS}ms ease-in-out`,
+
+        '@media': {
```
sergivillar, line 41: why not add mq to sprinkles?

playroom.config.js
```diff
@@ -1,4 +1,51 @@
...
+
+const getOutputPath = () => {
+    if (process.env.FORCE_MOBILE) {
+        return './public/playroom-mobile/';
+    }
+    if (process.env.FORCE_DESKTOP) {
+        return './public/playroom-desktop/';
+    }
+    return './public/playroom/';
+};
```
sergivillar, line 36: maybe this file could include a comment explaining the reasons of this, as stated in the PR description

sergivillar:
where is done the double build that you talked about in the description? I cant see it @atabel

brtbrt (approved):

atabel:
> where is done the double build that you talked about in the description? I cant see it @atabel

see `package.json` `vercel-build` script

pladaria (approved):

src/index.tsx
```diff
@@ -15,6 +15,7 @@ export {
     ServerSideStyles,
 } from './jss';
 
+export {vars as skinVars} from './skins/skin-contract.css';
```
atabel, line 18: Export vars (skin colors) to be usable by apps using Vanilla Extract

vite.config.js
```diff
@@ -5,7 +5,14 @@ import {vanillaExtractPlugin} from '@vanilla-extract/vite-plugin';
 import noBundlePlugin from 'vite-plugin-no-bundle';
 
 export default defineConfig({
-    plugins: [react(), vanillaExtractPlugin(), noBundlePlugin()],
+    plugins: [
+        react(),
+        vanillaExtractPlugin(),
+        noBundlePlugin({
+            // Change .css.js files to something else so that they don't get re-processed by consumer apps using vanilla extract too
+            fileNames: ({name}) => `${name.replace(/\.css$/, '.css.vanilla')}.js`,
```
atabel, line 13: Vanilla Extract Next plugin applies to node_modules too, so it tries to re-process mistica files (which are already compiled by mistica build) and result in errors. With this change, we make those files invisible for the Next plugin in apps using mistica and VE (zeus, for example).

pladaria: perhaps a better name would be `.css.mistica`

pladaria (approved):

vite.config.js
```diff
@@ -0,0 +1,24 @@
...
+import react from '@vitejs/plugin-react';
+import {vanillaExtractPlugin} from '@vanilla-extract/vite-plugin';
+import noBundlePlugin from 'vite-plugin-no-bundle';
+
+export default defineConfig({
+    plugins: [
+        react(),
+        vanillaExtractPlugin(),
+        noBundlePlugin({
+            // Change .css.js files to something else so that they don't get re-processed by consumer apps using vanilla extract too
```
sergivillar, line 12: this is because VE process all files ended in `css`? ir is because of vite?

PR #562, opened 2022-10-25
feat(Image): add loading/error experience

sergivillar:
https://www.figma.com/file/3iEvVzfHBtg3sROXBHzkn7/Image-%26-Video-Specs?node-id=306%3A1704


yceballost (changes requested):

src/carousel.tsx
```diff
@@ -33,6 +33,7 @@ const useBulletsStyles = createUseStyles((theme) => ({
         height: 8,
         borderRadius: '50%',
         transition: 'transform 0.3s ease-in-out, background-color 0.3s ease-in-out',
+        zIndex: 2,
```
sergivillar, line 36: changes in this file, zIndex 1 -> 2 are needed because now images has a zIndex 1, otherwise arrows and bullets are hidden bellow the image, any other idea?

pladaria: please, leave a comment explaining the reason of this

yceballost (approved):
magnifique✨

src/image.tsx
```diff
@@ -17,44 +24,66 @@ export const DisableBorderRadiusProvider: React.FC = ({children}) => (
     <DisableBorderRadiusContext.Provider value>{children}</DisableBorderRadiusContext.Provider>
 );
 
+type VivoLogoPRops = {
```
pladaria, line 27: PRops

src/image.tsx
```diff
@@ -81,77 +105,93 @@ export type ImageProps = {
...
                 src={src}
                 className={classes.image}
                 alt={alt}
                 onError={(event) => {
-                    setIsError(true);
+                    setTimeout(() => {
+                        setIsError(true);
+                        setIsLoading(false);
+                        setHideLoadingFallback(true);
+                    }, 1000);
```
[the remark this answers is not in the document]
sergivillar: removed, was a test

src/image.tsx
```diff
@@ -81,77 +105,93 @@ export type ImageProps = {
...
+            ...props
+        },
         ref
     ) => {
-        const supportsAspectRatio = useSupportsAspectRatio();
         const noBorderRadiusContext = useDisableBorderRadius();
         const noBorderSetting = noBorderRadius ?? noBorderRadiusContext;
         const [isError, setIsError] = React.useState(false);
+        const [isLoading, setIsLoading] = React.useState(true);
+        const [hideLoadingFallback, setHideLoadingFallback] = React.useState(false);
```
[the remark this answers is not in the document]
sergivillar: not exactly, I use the isLoading to start the opacity transition of the image, and the hideLoadingFallback to remove the loading skeleton when the fadeIn ends

src/utils/aspect-ratio-support.tsx
```diff
@@ -26,3 +27,92 @@ export const AspectRatioSupportProvider: React.FC<Props> = ({children}) => {
...
+    if (needsWrapper) {
+        return (
+            <div style={{width, height}} className={classes.wrapper}>
+                {container}
+            </div>
+        );
+    } else {
+        return container;
+    }
+};
```
brtbrt, line 118: alright! But: don't you think we'd need some unit test coverage here?

src/video.tsx
```diff
@@ -177,26 +121,31 @@ const Video = React.forwardRef<HTMLVideoElement, VideoProps>(
...
+                {isSafari() ? (
+                    <div
+                        style={{
+                            width: '100%',
+                            height: '100%',
+                            /**
+                             * In safari, when using a video with poster, the transition from pause to play does a flicker,
+                             * To avoid this, in Safari browsers, instead of using the poster attribute, we use a
+                             * wrapper with the poster as background image
+                             */
```
brtbrt, line 136: !https://media.giphy.com/media/0NwSQpGY6ipgOSt8LL/giphy.gif!

src/utils/aspect-ratio-support.tsx
```diff
@@ -26,3 +27,92 @@ export const AspectRatioSupportProvider: React.FC<Props> = ({children}) => {
 };
 
 export const useSupportsAspectRatio = (): boolean => React.useContext(AspectRatioSupport);
+
+const useAspectRatioStyles = createUseStyles(() => ({
```
brtbrt, line 31: if the intension is only exporting this very vsriable, I'd call the file the same name and I'd export it as default, unless there's some reason to follow this as a standard (in that case let me know!)

src/utils/aspect-ratio-support.tsx
```diff
@@ -26,3 +27,92 @@ export const AspectRatioSupportProvider: React.FC<Props> = ({children}) => {
...
+    container: {
+        '@supports (aspect-ratio: 1 / 1)': {
+            aspectRatio: ({aspectRatio}) => aspectRatio ?? 'unset',
```
brtbrt, line 34: I'm not that sure but maybe you can set `undefined` in case it's not needed, and the property won't be set at all?

src/utils/aspect-ratio-support.tsx
```diff
@@ -26,3 +27,92 @@ export const AspectRatioSupportProvider: React.FC<Props> = ({children}) => {
...
+        },
+        '$wrapper &': {
+            position: ({aspectRatio}) => (aspectRatio ? 'absolute' : 'static'),
+            width: '100%',
+            height: '100%',
+            top: 0,
```
brtbrt, line 40: 🤔 maybe you can get rid off `top` and `left` in case `aspectRatio` is not defined?

src/utils/aspect-ratio-support.tsx
```diff
@@ -26,3 +27,92 @@ export const AspectRatioSupportProvider: React.FC<Props> = ({children}) => {
...
+    children: React.ReactNode;
+    as?: React.ComponentType<any> | string;
+    style?: React.CSSProperties;
+    className?: string;
+};
+
+export const AspectRatioElement = (props: AspectRatioElementProps): JSX.Element => {
+    const supportsAspectRatio = useSupportsAspectRatio();
+    const withCssAspectRatio =
+        typeof props.width !== 'number' && typeof props.height !== 'number' && props.aspectRatio !== 0;
```
brtbrt, line 75: not an expert at all about the aspect ratio, so asking to make sure I got this: in case relative measures are set for both height and width, and aspectRatio is set: the height will be ignored in order to follow the ratio set? Did I get right?

src/utils/aspect-ratio-support.tsx
```diff
@@ -26,3 +27,92 @@ export const AspectRatioSupportProvider: React.FC<Props> = ({children}) => {
...
+    } else {
+        width = props.width || '100%';
+    }
+
+    const needsWrapper = withCssAspectRatio && !supportsAspectRatio;
+
+    const container = React.createElement(
+        props.as ?? 'div',
+        {
+            className: props.className ? `${props.className} ${classes.container}` : classes.container,
```
brtbrt, line 101: what about:

```
        className: classnames({
            [props.className]: props.className,
          },
          classes.container
        ),
```

brtbrt (approved):

src/utils/aspect-ratio-support.tsx
sergivillar: well this file has two exports know, the intention is to gather the aspect ratio functionality ins the same file

src/utils/aspect-ratio-support.tsx
sergivillar: `unset` is the value we should use here, https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio#syntax

and as we are switching from jss to vanilla extract, better keep this with the standard

src/utils/aspect-ratio-support.tsx
sergivillar: I lost this explanation

```  
// if width or height are numeric, we can calculate the other with the ratio without css.
        // if aspect ratio is 0, we use the original image proportions

src/utils/aspect-ratio-support.tsx
sergivillar: I'm not a big fan of using unit test for styles, and this case is tested with a screenshot of storybook covering all the cases

src/image.tsx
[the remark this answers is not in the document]
atabel: We need a state machine here :P

atabel (approved):

pladaria (approved):

src/utils/aspect-ratio-support.tsx
brtbrt: wat

PR #567, opened 2022-11-02
Google analytics with GDPR banner

brtbrt:


.storybook/manager.js
```diff
@@ -1,9 +1,15 @@
...
 // https://storybook.js.org/docs/react/configure/features-and-behavior
 addons.setConfig({
     panelPosition: 'right',
     theme: createStorybookTheme('Movistar'),
 });
+
+if (checkAcceptedCookies()) {
+    window.STORYBOOK_GA_ID = 'G-G928X876KM';
+    window.STORYBOOK_REACT_GA_OPTIONS = {};
+}
```
atabel, line 15: why is this needed both in `main.js` and `manager.js` ?

.storybook/cookie-policy.ts
```diff
@@ -0,0 +1,49 @@
+const COOKIE_ACCEPT_GDPR_NAME = 'GDPR';
+const COOKIE_ACCEPT_VALUE = 'accepted';
+
+function setGDPRCookie(expirationInDays = 365) {
```
atabel, line 4: follow coding styles
```suggestion
const setGdprCookie = (expirationInDays = 365) => {
```

atabel (commented):
Please test that this works in the GA account that DA gave us

pladaria (approved):

.storybook/cookie-policy.ts
```diff
@@ -0,0 +1,49 @@
...
+export default function checkAcceptedCookies(): boolean {
+    const gdpr = hasGdprCookie();
+
+    if (gdpr) {
+        setGDPRCookie();
+        return true;
+    }
+
+    showGDPRPrompt();
+    return false;
```
sergivillar, line 48: not sure about this logic, if you have GDPRcookie, why set it again?

and, also, if you have it but rejected (because you chose that), why show the banner again? if you rejected the cookie, this shouldn't be shown each time you enter the page, right?

yceballost:
is it possible to use mística style(for the background you can use `backgroundOverlay`) and button in this banner? 😶

image

https://ibit.ly/-436

.storybook/cookie-policy.ts
brtbrt: why this?

.storybook/cookie-policy.ts
brtbrt: @pladaria 's idea to make the cookie last forever. there's no reject button: in case you don't accept you'll have the banner untill you accept it (let's keep this simple)

.storybook/cookie-policy.ts
brtbrt: I mean, no problem at all, but "follow coding style" is not so good to get the grasp of your mindset :D

.storybook/cookie-policy.ts
sergivillar: well, if we do this for the GRPD law, we need to add the features that the law said, event to reject the cookies once accepted (https://gdpr.eu/cookies/)

brtbrt: and we'd need a popup explaining what the cookies  are, what the functionality is for, etc. I don't know, handling the way you say it means another kind of task: no problem, but let's decide this once and for all! 🤔😛

.storybook/cookie-policy.ts
atabel: there should be a good reason to not follow the coding style of the rest of the repo

brtbrt:
let's close the PR in order to clarify the privacy's functionality issues that has arisen: how to manage GDPR properly

PR #577, opened 2022-11-10
fix(Inputs): Review readonly input props

SergioMorchon:
This comes from https://github.com/Telefonica/gct-comms-tool-cms/pull/854, where we have a `ReadonlyArray` of steps passed to the `Stepper`, that requires a mutable `Array` on its `props`.


src/theme.tsx
```diff
@@ -210,38 +210,38 @@ export type EventFormat = 'universal-analytics' | 'google-analytics-4';
...
+    }>;
+    platformOverrides?: Readonly<{
         platform?: 'ios' | 'android' | 'desktop';
         insideNovumNativeApp?: boolean;
         userAgent?: string;
-    };
-    texts?: Partial<ThemeTexts>;
-    analytics?: {
+    }>;
+    texts?: Partial<Readonly<ThemeTexts>>;
```
atabel, line 225: Instead of this, type `ThemeTexts` as `Readonly`

src/theme.tsx
```diff
@@ -210,38 +210,38 @@ export type EventFormat = 'universal-analytics' | 'google-analytics-4';
...
+    analytics?: Readonly<{
         logEvent: (trackingEvent: TrackingEvent) => Promise<void>;
         eventFormat?: EventFormat;
-    };
-    dimensions?: {headerMobileHeight: number | 'mistica'};
+    }>;
+    dimensions?: Readonly<{headerMobileHeight: number | 'mistica'}>;
```
pladaria, line 230: This will require changes in webapp:
image

src/theme.tsx
pladaria: webapp:
image

src/theme.tsx
SergioMorchon: That is strange, because it tells the caller that Mística requires your object to be mutable (that means that you should expect Mística manipulating your objects).
Is there any other way we can pass the initial theme config to _pages_ without changing it afterwards that way?

src/theme.tsx
```diff
@@ -6,7 +6,7 @@ import type {Skin, Colors, SkinName, TextPresetsConfig} from './skins/types';
 import type {TrackingEvent} from './utils/types';
 import type {MediaQueries} from './utils/media-queries';
 
-export type ThemeTexts = typeof TEXTS_ES;
+export type ThemeTexts = Readonly<typeof TEXTS_ES>;
```
pladaria, line 9: Please verify the result en webapp:
image

What I do is:
1. build mistica: `yarn build`
2. `cp -r dist dist-es ../webapp/node_modules/@telefonica/mistica/`
3. in webapp: run `yarn flow` (sometimes I must `killall flow` first)

src/theme.tsx
pladaria: Let me change the question. Do we *really* need a readonly here?
To me the important type that mistica users shouldn't be allowed to change is `Theme` (line 247).

src/theme.tsx
SergioMorchon: Sure, done here: https://github.com/Telefonica/webapp/pull/2303

SergioMorchon: Done on the CMS, with no errors from this PR (I've detected this: https://github.com/Telefonica/mistica-web/pull/528#discussion_r1018979437)

src/theme.tsx
SergioMorchon: That is what I did here: https://github.com/Telefonica/webapp/pull/2303

No, it is not really needed until it is (the same reasoning we already had and still have with components like the `Stepper.steps`).

pladaria (approved):

atabel (approved):

PR #581, opened 2022-11-14
feat(vanilla-extract): migrate ButtonGroup, ButtonLayout, Button, Touchable, Image and Video

sergivillar:


src/button-layout.tsx
```diff
@@ -204,18 +139,52 @@ const ButtonLayout: React.FC<ButtonLayoutProps> = ({
 
     const needsLinkAlignment = !isTabletOrSmaller && align === 'left';
 
+    const getJustifyContent = () => {
+        if (align === 'center' || (align === 'full-width' && isTabletOrSmaller)) {
```
atabel, line 143: Can why try to use media queries instead of `useScreenSize`?

By using media queries we avoid hydration issues (`useScreenSize` does not work in SSR) or style flashes (first render to match server, second client render with the correct mq values).

Now that we're refactoring the styles, it's a good moment to rethink this kind of things

src/image.tsx
```diff
@@ -156,7 +133,11 @@ const Image = React.forwardRef<HTMLImageElement, ImageProps>(
                 }}
                 ref={ref}
                 src={src}
-                className={classes.image}
+                className={classnames(
+                    styles.image,
+                    ratio !== 0 ? sprinkles({position: 'absolute'}) : sprinkles({position: 'static'}),
+                    noBorderSetting ? styles.noBorder : sprinkles({borderRadius: 8})
```
atabel, line 139:
```suggestion
                    sprinkles({
                        position: ratio !== 0 ? 'absolute' :  'static',
                        borderRadius: noBorderSetting ? undefined : 8,
                    })
```

src/sprinkles.css.ts
```diff
@@ -13,7 +17,7 @@ const colorProperties = defineProperties({
     },
 });
 
-const sizes = [0, 2, 4, 8, 12, 16, 24, 32, 40, 48, 56, 64, 72, 80] as const;
+const sizes = [0, 2, 4, 6, 8, 12, 16, 24, 32, 40, 48, 56, 64, 72, 80] as const;
```
atabel, line 20: hmmm... where do you need 6px?

src/video.css.ts
```diff
@@ -0,0 +1,12 @@
+import {style} from '@vanilla-extract/css';
+import {sprinkles} from './sprinkles.css';
+
+export const video = style([
+    sprinkles({display: 'block', width: '100%', height: '100%'}),
+    {
+        background: 'transparent',
+        objectFit: 'cover',
+        maxWidth: '100%',
+        maxHeight: '100%',
```
atabel, line 10: all these could be moved to sprinkles

src/button-layout.css.ts
```diff
@@ -0,0 +1,89 @@
...
+export const link = style([
+    sprinkles({display: 'flex', width: '100%'}),
+    {
+        margin: buttonLayoutSpacing / 2,
+        justifyContent: 'inherit',
+    },
+]);
+
+export const linkAlignment = style({
+    marginLeft: buttonLayoutSpacing / 2 - 6,
```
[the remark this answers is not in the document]
sergivillar: maybe... could this be `PADDING_Y_LINK`?

src/sprinkles.css.ts
sergivillar: paddingY: PADDING_Y_LINK,

in link stlyes

pladaria (approved):

src/sprinkles.css.ts
sergivillar: what about the 12 value? I added this because I saw that 12 (that maybe we should remove it)

atabel (approved):

PR #584, opened 2022-11-15
WEB-701 restore readonly type

pladaria:


.yarnrc.yml
```diff
@@ -1,7 +1,7 @@
 nodeLinker: node-modules
 
 plugins:
-    - path: .yarn/plugins/@yarnpkg/plugin-interactive-tools.cjs
-      spec: '@yarnpkg/plugin-interactive-tools'
+  - path: .yarn/plugins/@yarnpkg/plugin-interactive-tools.cjs
+    spec: "@yarnpkg/plugin-interactive-tools"
```
SergioMorchon, line 5: If you run `prettier`, will change this file?

SergioMorchon (approved):

atabel (approved):

.yarnrc.yml
pladaria: image

hmm this should be formatted by the precommit hook

pladaria: looks like the hook isn't working (at least on my machine)

Thanks noticing it

PR #589, opened 2022-11-17
feat(vanilla-extract): migrate carousel

atabel:
- Migrated `Carousel` to Vanilla Extract
- Removed `ContainerTypeContext` and replaced by css classes in grid layout columns
  - Finally replaced by a new context that only takes into account the desktop case. Not used for the moment, but may be needed in a near future


src/carousel.css.ts
```diff
@@ -0,0 +1,334 @@
...
+        transition: 'transform 0.3s ease-in-out, background-color 0.3s ease-in-out',
+        zIndex: 2, // needed because images has zIndex 1, otherwise this component won't be shown
+
+        '@media': {
+            [mq.tabletOrSmaller]: {
+                width: 4,
+                height: 4,
+            },
+        },
+    },
```
sergivillar, line 25: I dont know why, but in my PRs im using always mobile/tablet as default, and isDesktop to media queries, I dont know either if this has any sense xD

src/carousel.css.ts
```diff
@@ -0,0 +1,334 @@
...
+
+            ':first-child': {
+                width: `calc(1 / ${itemsPerPageMobile} * 100% - ${gap} * (${itemsPerPageMobile} - 1) / ${itemsPerPageMobile} * 1px)`,
+            },
+
+            scrollSnapAlign: 'start',
+            scrollMargin: `calc(${mobilePageOffsetPrev} * 1px)`,
+
+            selectors: {
+                [`${carouselWithScroll}:not(${centeredCarousel}) &`]: {
```
sergivillar, line 169: I had some issues with this approach, if `centeredCarousel` in the future has also a sprinkles (just because it changes) this selector wont work, because will be something like this `:not(.style_blabla .sprinkles_xxx)` that does not work on CSS. 
I found safer create empty clases `const notSelector = style({})`to cover this usages, and avoid broke it without notice in the future...
WDYT?

src/utils/types.tsx
```diff
@@ -10,10 +10,3 @@ export type IconProps = {
     className?: string;
     style?: React.CSSProperties;
 };
-
-export type ContainerType =
-    | 'desktop-small-column'
-    | 'desktop-medium-column'
-    | 'desktop-wide-column'
-    | 'tablet-column'
-    | 'mobile-column';
```
wyunreal, line 19: As confirmed with design (@MontseTelefonica ) we should keep support for container types in JS code (not only CSS), because next week we'll require the carousel's itemsPerPage prop to behave differently (to have different values) depending on the containers, including desktop small, medium and wide columns, mobile column and tablet column.

wyunreal (changes requested):

src/carousel.tsx
```diff
@@ -323,20 +161,11 @@ const BaseCarousel: React.FC<BaseCarouselProps> = ({
     onPageChange,
     dataAttributes,
 }) => {
-    const {texts} = useTheme();
-    const containerType = useContainerType();
+    const {texts, platformOverrides} = useTheme();
     const itemsPerPageConfig = normalizeItemsPerPage(itemsPerPage);
     const mobilePageOffsetConfig = normalizeMobilePageOffset(mobilePageOffset);
-    const {isDesktopOrBigger, isLargeDesktop} = useScreenSize();
+    const {isDesktopOrBigger} = useScreenSize();
```
sergivillar, line 167: this cant be removed?

src/carousel.css.ts
atabel: are you sure about that? I'd say it'd still work, but I'll re-check it

src/carousel.tsx
atabel: no, because we need it for the itemsperpage config

src/carousel.css.ts
sergivillar: yep, had this issue this Chip component, in this case this works because `centeredCarousel` only has a style (no sprinkles, no base class) so only one string class is generated `.centeredCarousel_xxx`, but if you add a base case a sprinkles, that this generate more than one class and this :not() will fail

atabel: checked and it works without problems, VE does the magic and only include one class

atabel: image

src/utils/types.tsx
atabel: ok, added a context

.yarnrc.yml
```diff
@@ -1,7 +1,7 @@
 nodeLinker: node-modules
 
 plugins:
-  - path: .yarn/plugins/@yarnpkg/plugin-interactive-tools.cjs
-    spec: "@yarnpkg/plugin-interactive-tools"
+    - path: .yarn/plugins/@yarnpkg/plugin-interactive-tools.cjs
+      spec: '@yarnpkg/plugin-interactive-tools'
```
pladaria, line 5: your pre-commit doesn't seem to work

sergivillar (approved):

src/screen-size-context-provider.tsx
```diff
@@ -93,10 +93,10 @@ const ScreenSizeContextProvider: React.FC<Props> = ({children}) => {
             isTablet,
             isTabletOrBigger,
```
pladaria, line 94: not related to your changes, but there's a typo in `mediaQueries.largueDesktop` (largue)

src/desktop-container-type-context.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react';
+
+export type ContainerType = 'small' | 'medium' | 'large';
+
+const DesktopContainerTypeContext = React.createContext<ContainerType | null>(null);
+
+export const useDesktopContainerType = (): ContainerType | null =>
+    React.useContext(DesktopContainerTypeContext);
```
pladaria, line 8: The hook and context should be exposed in the index

atabel: no, it's an implementation detail and will only be used inside mistica for the moment.
If we need to expose it in the future, will do it

src/screen-size-context-provider.tsx
atabel: I'll fix it

atabel: done

flow-defs/desktop-container-type-context.js.flow
```diff
@@ -0,0 +1,7 @@
+// @flow
+
+import * as React from "react";
+export type ContainerType = "small" | "medium" | "large";
```
wyunreal, line 4: This only refers to desktop, right ?

wyunreal: If yes, I would rename to DesktopContainerType

src/desktop-container-type-context.tsx
wyunreal: agree, its internal by now

wyunreal (approved):

flow-defs/desktop-container-type-context.js.flow
atabel: yes, you're right, I'll rename it

atabel: done

pladaria (approved):

PR #590, opened 2022-11-17
feat(vanilla-extract): Dialog, Divider, DoubleField, EmptyState, EmptyStateCard

sergivillar:
EmptyState component has been completely rewritten, IMO changes in the 2 broken screenshots are expected (As internal structure has been simplified, 1px has been moved around), if you are okey with this, I'll update that screenshot.

Dialog screenshot has been updated, now is more accurate. This is because previously the icon size was calculated used useScreenSize hook and `size` props from icon, now we have a wrapper around icon and styled with a media query (now size is 64px, previously 67, that's the difference)


src/empty-state.css.ts
```diff
@@ -0,0 +1,93 @@
...
+export const contentVariants = styleVariants({
+    default: {
+        marginTop: 0,
+        '@media': {
+            [mq.desktopOrBigger]: {
+                margin: 64,
+            },
+        },
+    },
+    largeImage: {
```
yceballost, line 42: I found differences in image behavior in mobile

production: [example 1](https://ibit.ly/lbCm) | [example 2 (1:1)](https://ibit.ly/5ibK)
this branch: [example 1](https://ibit.ly/JbUR) | [example 2 (1:1)](https://ibit.ly/Ezj_)

maybe missing `width 100%`?

I think the correct behavior should not show a deformed image, just should fit the photo to the longest side like the image component → https://ibit.ly/FHan

In desktop sizes seem to behave well

yceballost (changes requested):

src/empty-state.css.ts
sergivillar: @yceballost changes image behaviour test again in mobile and desktop (to be sure all still works as expected:) )

yceballost (approved):

src/double-field.css.ts
```diff
@@ -0,0 +1,33 @@
+import {styleVariants} from '@vanilla-extract/css';
+import * as mq from './media-queries.css';
+import {sprinkles} from './sprinkles.css';
+import {DEFAULT_WIDTH} from './text-field-components';
```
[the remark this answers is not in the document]
sergivillar: text-field-components is not migrated yet

src/empty-state.css.ts
```diff
@@ -0,0 +1,109 @@
...
+        [mq.desktopOrBigger]: {
+            position: 'absolute',
+            height: '100%',
+        },
+    },
+});
+
+export const largeImage = style([
+    {
+        height: '100%',
```
pladaria, line 63: sprinkles

src/empty-state.tsx
```diff
@@ -91,6 +48,14 @@ interface IconProps extends BaseProps {
     largeImageUrl?: undefined;
 }
 
+const getBorderStyle = (isInverse: boolean) => {
+    if (isInverse) {
+        return styles.inverseBorder;
+    }
+
+    return sprinkles({border: 'regular'});
```
pladaria, line 56: I think this should be a style defined in the css

atabel (approved):

pladaria (approved):

PR #592, opened 2022-11-17
feat(vanilla-extract): Migrate FadeIn, Feedbacks, Form, HighlightedCard, LoadingBar

sergivillar:


src/highlighted-card.tsx
```diff
@@ -104,50 +67,56 @@ const Content: React.FC<Props> = (props) => {
...
-                    </>
-                )}
-            </div>
-            {imageUrl && (
+        <div
+            style={{
+                ...assignInlineVars({[styles.vars.width]: props.width ? `${props.width}px` : '100%'}),
+                display: 'flex',
+                width: '100%',
+            }}
```
sergivillar, line 78: new wrapper needed to set css var, any other idea?

src/portal.tsx
```diff
@@ -34,12 +34,12 @@ export const Portal: React.FC<Props> = ({children, className}) => {
...
     React.useEffect(() => {
         if (container && className) {
-            container.classList.add(className);
+            container.classList.add(...className.split(' '));
         }
 
         return () => {
             if (container && className) {
-                container.classList.remove(className);
+                container.classList.remove(...className.split(' '));
```
sergivillar, line 42: check this out, now we need this, VE could pass more than one class here (LoadingBar needs this)

src/highlighted-card.tsx
atabel: you can set the var to the nodes that use it instead

src/portal.tsx
atabel: good catch

atabel (approved):

src/boxed.tsx
```diff
@@ -17,6 +17,8 @@ type Props = {
     /** "data-" prefix is automatically added. For example, use "testid" instead of "data-testid" */
     dataAttributes?: DataAttributes;
     'aria-label'?: string;
+    width?: number | string;
+    height?: number | string;
```
atabel, line 21: please, can you review MediaCard, DataCard and SnapCard to use these new props instead of a className to set the width and height?

atabel (approved):

pladaria (approved):

PR #593, opened 2022-11-18
feat(vanilla-extract): Migrate TextLink

SergioMorchon:


src/text-link.css.ts
```diff
@@ -0,0 +1,58 @@
...
+            },
+        },
+    },
+    selectors: {
+        '&[disabled]': {
+            opacity: 0.5,
+            textDecoration: 'none',
+        },
+    },
+});
```
sergivillar, line 27: we have something called `[sprinkles](https://vanilla-extract.style/documentation/packages/sprinkles/)` for styles that are widely used in mistica, so here you must combine sprinkles with styles, something like 

```
style([
    sprinkles({width: 'auto', display: 'inline', color: vares.colors.textLink}),
   { ...rest of styles...}
])
```


you can check sprinkles.css.ts to check what styles do we have here



also apply this change for the rest of the styles

SergioMorchon: Thank you! I was aiming to learn about `vanilla-extract`, so all your feedback is very welcome.
I'll look into that!

src/text-link.tsx
```diff
@@ -107,9 +76,10 @@ const TextLink: React.FC<TextLinkProps> = ({children, className = '', small, dis
                 props.trackingEvent ?? (props.trackEvent ? createDefaultTrackingEvent() : undefined)
             }
             disabled={disabled || formStatus === 'sending'}
-            className={classnames(classes.textLink, className, {
-                [classes.small]: small,
-                [classes.inverse]: isInverse,
+            className={classnames(styles.textLink, className, {
+                [styles.small]: small,
+                [styles.inverse]: isInverse,
+                [styles.inverseDark]: isInverse && isDarkMode,
```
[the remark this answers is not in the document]
SergioMorchon: Gotcha, thanks for the quick meeting to explain me the runtime differences vs JSS and why this matters!
I'll change it

sergivillar (approved):

atabel (approved):

pladaria (approved):

PR #597, opened 2022-11-21
feat(vanilla-extract): Migrate Tabs

SergioMorchon:


pladaria (approved):

src/tabs.css.ts
```diff
@@ -0,0 +1,146 @@
...
+        },
+    },
+});
+
+const tabSelectedBae = sprinkles({
+    color: vars.colors.textPrimary,
+});
+
+export const tabSelectionVariants = styleVariants({
+    noSelected: [
```
sergivillar, line 115: maybe instead of `noSelected` -> `default`?

sergivillar (approved):

src/tabs.css.ts
SergioMorchon: I considered it, but finally though that `noSelected` vs `selected(Animating)` would be clearer

PR #600, opened 2022-11-22
feat(ResponsiveLayout): allow background color

atabel:
https://github.com/Telefonica/mistica-design/discussions/885


src/responsive-layout.tsx
```diff
@@ -2,24 +2,40 @@ import * as React from 'react';
...
+}) => {
+    const isInverseOutside = useIsInverseVariant();
     return (
-        <div
-            className={classnames(styles.container, className)}
-            {...getPrefixedDataAttributes(dataAttributes)}
-        >
-            <div className={fullWidth ? styles.fullWidth : styles.responsiveLayout}>{children}</div>
-        </div>
+        <ThemeVariant isInverse={isInverse || isInverseOutside}>
```
sergivillar, line 28: if `isInverseOutside` is already true, is this ThemeVariant needed?

sergivillar (approved):

yceballost (approved):
✨

src/responsive-layout.tsx
atabel: not really, but it's simpler this way that rendering the `ThemeVariant` conditionally

pladaria (approved):

PR #602, opened 2022-11-23
fix(styles): some fixes in components migrated to VE

atabel:


src/theme-context-provider.tsx
```diff
@@ -33,6 +33,22 @@ import type {Theme, ThemeConfig} from './theme';
  */
 const misticaCss = '<MISTICA_CSS>';
 
+const MisticaCss = () => {
```
atabel, line 36: This moves the VE styles to a `<style>` tag in the `<head>`. This fixes style precedence issues with jss in webapp.

This approach doesn't work with SSR, because the `<style>` is only injected in client side. This solution is transitional until we remove JSS from mistica, but meanwhile if a web needs SSR and mistica, it can load the styles using the `css/mistica.css` file and calling passing the `experimental_disableCssInJs` prop to `ThemeContextProvider` to avoid injecting the `<style>` tag in the client

src/carousel.css.ts
```diff
@@ -137,8 +137,10 @@ export const carousel = style([
     },
 ]);
 
+const responsiveLayoutSideMargin = fallbackVar(responsiveLayoutVars.sideMargin, '0px');
```
atabel, line 140: this is needed because there is a story in webapp that loads a carousel outside a ResponsiveLayout, and in that case the sideMargin css var is not defined. This fallbacks to 0px when not defined and makes the carousel component more resilient

src/highlighted-card.css.ts
```diff
@@ -24,6 +24,7 @@ export const touchableContainer = sprinkles({
     border: 'none',
     padding: 0,
     width: '100%',
+    backgroundColor: 'transparent',
```
atabel, line 27: this was not being reset because we use the BaseTouchable

src/list.tsx
```diff
@@ -493,14 +493,7 @@ const RowContent = React.forwardRef<TouchableElement, RowContentProps>((props, r
...
     return (
-        <Box
-            paddingX={16}
-            className={classNames(styles.rowContent, {
-                [styles.hoverBackground]: !(disabled || isInverse),
-                [styles.pointer]: !disabled,
-            })}
-            role={role}
-        >
+        <Box paddingX={16} className={styles.rowContent} role={role}>
```
atabel, line 496: not touchable Rows should not have the hover effect

src/text-link.css.ts
```diff
@@ -73,7 +73,3 @@ export const variants = styleVariants({
         },
     ],
 });
-
-export const small = style({
-    fontSize: 14,
```
atabel, line 78: moved to inline styles because the `BaseTouchable`  overrides this to `'inherit'`

src/text.css.ts
```diff
@@ -37,6 +37,7 @@ export const withoutWordBreak = style({
 });
 
 export const text = style({
+    margin: 0, // Needed to reset the default browser margin that adds to p, h1, h2... elements.
```
atabel, line 40: moved here from inline styles, to allow webapp to override it with classNames

src/text.tsx
```diff
@@ -42,6 +36,7 @@ export interface TextPresetProps {
     truncate?: boolean | number;
     wordBreak?: boolean;
     hyphens?: 'auto' | 'manual' | 'none';
+    textAlign?: 'center';
```
atabel, line 39: new prop for `Text`, used in `Tabs`

flow-defs/theme-context-provider.js.flow
```diff
@@ -18,6 +18,7 @@ declare type Props = {
   providerId?: string,
   as?: string,
   children?: React.Node,
+  experimental_disableCssInJs?: boolean,
```
pladaria, line 21: :O

pladaria (approved):

src/box.tsx
```diff
@@ -43,19 +43,25 @@ const Box = React.forwardRef<HTMLDivElement, Props>(
         },
         ref
     ) => {
+        const paddingStyles = {paddingTop, paddingBottom, paddingLeft, paddingRight};
+        let paddingClasses = '';
+        try {
+            paddingClasses = sprinkles(paddingStyles);
```
sergivillar, line 49: do we really need this? I know that we have a 24px in webapp, but I think that value should be changed to 24 or 32, if not this change could also be applied to Stack and other components

sergivillar (approved):

src/box.tsx
atabel: I prefer to be safe here for the moment

PR #604, opened 2022-11-25
fix(ButtonLayout): unstability in screenshot tests

atabel:


src/button-layout.tsx
```diff
@@ -139,6 +139,7 @@ const ButtonLayout: React.FC<ButtonLayoutProps> = ({
         <div
             ref={wrapperElRef}
             className={classnames(
+                styles.baseContainer,
```
atabel, line 142: while the layout is in the "measuring state", we also need the `display: flex`, `margin`, `flexWrap` styles

sergivillar (approved):

PR #610, opened 2022-12-09
feat(Carousel): extend items per page for tablet and desktop container types

wyunreal:


yceballost:
i can see different behaviors between storybook and playroom, is it expected?
(Check the size of the next item)
image
image

src/carousel.css.ts
```diff
@@ -170,26 +172,26 @@ export const carouselItem = style({
     },
     '@media': {
         [mq.tabletOrSmaller]: {
```
SergioMorchon, line 174: I think this should be `mq.tablet`, to avoid overlapping with `mq.mobile`

src/carousel.tsx
```diff
@@ -62,22 +63,54 @@ export const PageBullets: React.FC<PageBulletsProps> = ({currentIndex, numPages,
     );
 };
 
-type ItemsPerPageProp = {mobile?: number; desktop?: number} | number;
+type DesktopItemsPerPage = {small?: number; medium?: number; large?: number} | number;
+type ItemsPerPageProp = {mobile?: number; tablet?: number; desktop?: DesktopItemsPerPage} | number;
```
atabel, line 67: Do you think we'll need the different values for desktop? I'd say we'll always want `itemsPerPage.desktop.small` to be equal to `itemsPerPage.mobile` or `itemsPerPage.desktop.medium` to be equal to `itemsPerPage.tablet`.

If we asume that, we can simplify the api surface of the component

atabel (commented):
Please review the failing screenshot tests

src/carousel.css.ts
```diff
@@ -198,6 +200,25 @@ export const carouselItem = style({
                 },
             },
         },
+        [mq.mobile]: {
```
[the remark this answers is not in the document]
wyunreal: Great, I'll do that ;)

src/carousel.tsx
wyunreal: As talked with design, they want different values :)

yceballost: I can't remember talk about that @wyunreal 🤔 

In a sidebar on desktop, you can use mobile itemsPerPage by default, I can't find any case against this. Do you know any example?

Maybe we need to include this detail in specs to avoid confussions, @aweell ?

wyunreal: image

wyunreal: Talk is not a talk, its a figma
Also, I think its a really bad idea to think that a small column on desktop will be always the same than  a mobile device, same with mid size desktop column and tablet, its a bad idea to couple these different concepts.

(and we already have a requirement to have this different cases).

wyunreal: figma: https://www.figma.com/file/xNUwhH29z3Fre4foNltgtX/CPS_Specs?node-id=2670%3A59287&t=oRYoD9xihBs6ewJB-0

wyunreal: > In a sidebar on desktop, you can use mobile itemsPerPage by default, I can't find any case against this. Do you know any example?

By default, we are using this as you say.

atabel: > image

`2.2` ? we shouldn't use decimals here. Appart from that, I see the equivalence between mobile and small column and tablet and medium column

wyunreal: I dont see the equivalence, are totally different concepts. but I habe a requirement on this figma, if you think its better to wait, I suggest @yceballost to talk to @MontseTelefonica  in order to clarify this. If designers thinks we can just have mobile / desktop / tablet, I'm ok, but please think it twise, a small column on desktop is not a mobile device and viceversa ( think that there are more than one small column sizes because it depends on the grid template the page is using, same with medium column sizes).

wyunreal: @pladaria wdyt ? do you think a mobile or tablet is the same, conceptually, than a small / medium desktop column ? do you think its semantic to say: mobile: 2 and expects to have 2 also in small desktop column ? TBH, I think its a bad API design buttt, if you think its a good API design, I will follow your recommendation for sure :)

atabel: With "I see the equivalence" I meant in that concrete definition (except for the decimal part):
- mobile: 1 -> desktop-small-column: 1
- tablet 2.2 -> desktop-medium-column: 2

This isn't an API design discussion, it should be a design needs discussion. If Design doesn't need the extra configurability I wouldn't include it in the component. If it's really needed, then I'm ok

src/carousel.css.ts
wyunreal: this comment is obsolete as the solution changed ;)

src/carousel.tsx
wyunreal: Its needed as we have a Figma requiring it ;)

wyunreal: So can you please approve the PR ? or at least please move the question to design :)

atabel: Ok, talked in a call with Design. This is a requirement, so the api makes sense

src/carousel.css.ts
```diff
@@ -94,14 +94,18 @@ export const carouselContainer = sprinkles({
...
+const itemsPerPage = createVar();
 const itemsPerPageMobile = createVar();
+const itemsPerPageTablet = createVar();
 const itemsPerPageDesktop = createVar();
 const mobilePageOffsetPrev = createVar();
 const mobilePageOffsetNext = createVar();
 const gap = createVar();
 
 export const vars = {
+    itemsPerPage,
```
atabel, line 106: you don´t need to export this var because it isn´t used outside

src/__stories__/carousel-on-different-container-sizes-story.tsx
```diff
@@ -20,29 +20,29 @@ export default {
...
+    withBullets,
     cardsTitlePrefix,
-    intemsPerPage,
+    itemsPerPage,
 }: {
+    numItems: number;
+    withBullets: boolean;
     cardsTitlePrefix: number;
-    intemsPerPage: number;
+    itemsPerPage: {mobile: number; tablet: number; desktop: {small: number; medium: number; large: number}};
```
pladaria, line 39: alternative (and backwards compatible) API:
```ts
itemsPerPage: 
    | number 
    | {mobile: number: tablet: number, desktop: number | {small: number; medium: number; large: number} }
```

So all these options would be valid:
```ts
itemsPerPage={5} // applies for all sizes
itemsPerPage={mobile: 3, tablet: 5, desktop: 8}
itemsPerPage={mobile: 3, tablet: 5, desktop: {small: 3, medium: 5: large: 8}}}
```

wyunreal: I prefer to support the same than in the carousel component (at least the more similar)

pladaria (approved):

wyunreal:
> i can see different behaviors between storybook and playroom, is it expected? (Check the size of the next item) image image

This was caused by the different margin sizes of responsive layout in mobile and tablet, now fixed by another branch.

src/carousel.tsx
```diff
@@ -66,22 +67,54 @@ export const PageBullets: React.FC<PageBulletsProps> = ({currentIndex, numPages,
...
+    desktopContainerType: DesktopContainerType,
+    itemsPerPage?: ItemsPerPageProp
+): {
+    mobile: number;
+    tablet: number;
+    desktop: number;
+} => {
+    const defaultItemsPerPage = {mobile: 1, tablet: 1, desktop: {small: 1, medium: 2, large: 3}};
+    if (!itemsPerPage) {
+        return {...defaultItemsPerPage, desktop: defaultItemsPerPage.desktop.large};
```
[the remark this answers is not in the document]
wyunreal: yes, you are right, changed !!

src/carousel.tsx
```diff
@@ -62,22 +63,54 @@ export const PageBullets: React.FC<PageBulletsProps> = ({currentIndex, numPages,
...
+
+const normalizeItemsPerPage = (
+    desktopContainerType: DesktopContainerType,
+    itemsPerPage?: ItemsPerPageProp
+): {
+    mobile: number;
+    tablet: number;
+    desktop: number;
+} => {
+    const defaultItemsPerPage = {mobile: 1, tablet: 1, desktop: {small: 1, medium: 2, large: 3}};
```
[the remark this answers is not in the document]
wyunreal: changed tablet default to 2, now these are the defaults design wants ;)

atabel (approved):

SergioMorchon (approved):

yceballost (approved):

PR #614, opened 2022-12-14
feat(css): don't inline mistica css in js

atabel:
BREAKING CHANGE: mistica no longer inlines css in js, you must import the css/mistica.css in your app


sergivillar (approved):

brtbrt (approved):

pladaria:
I'm missing some doc update

brtbrt (approved):

PR #622, opened 2022-12-16
feat(DataAttributes): expose component name as data attribute

pladaria:


.github/actions/audit-accessibility/main.js
```diff
@@ -201,17 +206,24 @@ const main = async () => {
...
-            })
-        );
-    }
+                resolve();
+            });
+        });
+    };
+
+    const pool = new PromisePool(job, os.cpus().length - 1);
+    await pool.start();
```
pladaria, line 226: small improvement, instead of resolving by chunks, where the next chunk is not initiated until the previous one has finished, this has a pool and when a promise resolves, a new job is created

bnombela (approved):
lgtm!

.github/actions/audit-accessibility/main.js
atabel: have you measured the improvement?

src/text.tsx
```diff
@@ -126,7 +126,7 @@ export const Text: React.FC<TextProps> = ({
             id,
             role,
             'aria-level': ariaLevel,
-            ...getPrefixedDataAttributes(dataAttributes),
+            ...getPrefixedDataAttributes(dataAttributes, 'Text'),
```
atabel, line 129: this component is too low level, I think the data-component-name won't give any value in TextX components

atabel (approved):

sergivillar (approved):

.github/actions/audit-accessibility/main.js
pladaria: no, just intuition, but let me see

PR #638, opened 2023-01-09
fix(menu): ssr hydration missmatch

### LINKED ISSUE #632: Hydration error when rendering Menu component NextJS
**Describe the bug**
I have a 'use client' component in NextJS 13 app dir which renders a Menu component. I can see a console error related to style prop of a div

**Screenshots**
Captura de pantalla 2023-01-03 a las 13 14 12

atabel:
fixes #632


src/menu.tsx
```diff
@@ -153,13 +153,15 @@ const Menu: React.FC<MenuProps> = ({renderTarget, renderMenu, width, position =
             style={{
                 ...assignInlineVars({
                     [styles.vars.width]: width ? `${width}px` : '100%',
-                    [styles.vars.top]: itemsComputedProps.top ?? '',
```
atabel, line 156: in SSR, the `''` is removed from `style`, but not in client side.
I haven't been able to reproduce this with our ssr tests, because we are using `renderToString` api, that seems to respect the original `''`. I managed to replicate the issue in next.js using the beta `app/` directory, but next.js is probably using `renderToReadableStream` or `renderToPipeableStream` instead of `renderToString` in SSR

pladaria: an alternative could be to set the style to "initial", "none", 0..

pladaria (approved):

sergivillar (approved):

PR #641, opened 2023-01-10
feat(react): support for react 18

atabel:


sergivillar:
is react 18 bigger @atabel ? lib overhead has increased

atabel:
> is react 18 bigger @atabel ? lib overhead has increased

hmmm... it seems so, after recalc the baseline, it seems like lib overhead has not changed too much, so the change you were seeing before seems to be caused by react update, yes

src/password-field.tsx
```diff
@@ -68,8 +68,10 @@ const PasswordField: React.FC<PasswordFieldProps> = ({
     React.useEffect(() => {
         const input = inputRef.current;
         if (input) {
-            input.selectionStart = caretPositionRef.current;
-            input.selectionEnd = caretPositionRef.current;
+            requestAnimationFrame(() => {
+                input.selectionStart = caretPositionRef.current;
+                input.selectionEnd = caretPositionRef.current;
+            });
         }
```
pladaria, line 75: cancel raf on unmount?

src/test-utils/index.tsx
```diff
@@ -189,3 +189,31 @@ export const openSSRPage = ({
 
 export const setRootFontSize = (px: number): Promise<void> =>
     page.$eval('html', (e, px) => e.setAttribute('style', `font-size: ${px}px;`), px);
+
+export const waitFor = <T,>(
```
pladaria, line 193: Why don't you use the waitFor from RTL?

https://testing-library.com/docs/dom-testing-library/api-async/#waitfor

pladaria (approved):

PR #665, opened 2023-02-14
WEB-766 hero component

sergivillar:


yceballost (changes requested):

playroom/snippets.tsx
```diff
@@ -1502,6 +1502,68 @@ const skeletonSnippets = [
...
+            }
+            headline={<Tag type="active">Headline</Tag>}
+            pretitle="Pretitle"
+            title="Title"
+            description="This is a long description with a long text to see how this works"
+            extra={<Placeholder />}
+            button={<ButtonPrimary fake>Action</ButtonPrimary>}
+            buttonLink={<ButtonLink href="#">Link</ButtonLink>}
+            dataAttributes={{testid: 'hero'}}
+            desktopMediaPosition="right"
```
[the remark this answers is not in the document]
sergivillar: well, that is not a valid value and TS will complain about that

playroom/snippets.tsx
```diff
@@ -1502,6 +1502,68 @@ const skeletonSnippets = [
...
+          withBullets
+          items={Array.from({length: 3}).map(() => (
+              <Box paddingBottom={32}>
+                <Hero
+                  background="default"
+                  media={
+                      <Image
+                          src="https://api.lorem.space/image/furniture?w=1500&h=1500&hash=8B7BCDC2"
+                          aspectRatio="16:9"
+                          noBorderRadius
```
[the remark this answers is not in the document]
sergivillar: well images in Slidshow have an special behaviour that comes from the Slideshow, is defined in mistica as is

src/hero.tsx
```diff
@@ -0,0 +1,188 @@
...
+                    {extra && <div>{extra}</div>}
+                </div>
+                {(button || buttonLink) && <ButtonGroup primaryButton={button} link={buttonLink} />}
+            </Stack>
+        </section>
+    );
+};
+
+type HeroProps = {
+    height?: string;
```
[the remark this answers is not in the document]
sergivillar: You have a Box defined in this example!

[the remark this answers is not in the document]
yceballost: ouch! true

yceballost (approved):

aweell (approved):

AnaMontes11 (approved):
Amazing job @sergivillar !

src/hero.tsx
```diff
@@ -0,0 +1,192 @@
...
+            >
+                <ResponsiveLayout
+                    isInverse={isInverse}
+                    dataAttributes={{'component-name': 'Hero', ...dataAttributes}}
+                >
+                    <GridLayout
+                        template="6+6"
+                        left={
+                            <Box
+                                paddingY={isInsideSlideShow ? 80 : 56}
```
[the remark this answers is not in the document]
sergivillar: this 80px is a fix for let the Slideshow bullets work with Hero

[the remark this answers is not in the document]
aweell: They still work when the padding is 56 in desktop due to image / actions position, we should keep that measure instead of 80px since there's no possible collision.

Images with 56px of padding:

Screenshot 2023-02-22 at 09 46 01
Screenshot 2023-02-22 at 09 45 56

src/__stories__/hero-story.tsx
```diff
@@ -0,0 +1,111 @@
...
+    title: 'Components/Hero/Component',
+};
+
+type HeroArgs = {
+    media: 'video' | 'image';
+    headlineType: TagType;
+    headline: string;
+    pretitle: string;
+    title: string;
+    subtitle: string;
```
atabel, line 16: not used

playroom/snippets.tsx
```diff
@@ -1439,6 +1439,113 @@ const skeletonSnippets = [
...
+            inverseBullets={false}
+            items={Array.from({ length: 3 }, (_, idx) => (
+              <Hero
+                background="default"
+                media={
+                  <Image
+                    src={
+                      [
+                        "https://api.lorem.space/image/furniture?w=1500&h=1500&hash=8B7BCDC2",
+                        "https://api.lorem.space/image/furniture?w=1500&h=1500&hash=8B7BCDC4",
```
atabel, line 1526: you only have 2 urls here, but the slideshow has 3 items

src/__stories__/hero-slideshow-story.tsx
```diff
@@ -0,0 +1,41 @@
...
+    );
+};
+
+export const Default: StoryComponent = () => (
+    <Slideshow
+        withBullets
+        inverseBullets={false}
+        dataAttributes={{testid: 'hero'}}
+        items={Array.from({length: 3}).map(() => (
+            <HeroComponent />
```
atabel, line 36: I'd use different Hero setups here (for example, some without extra or description) to see how the actions align to bottom in that case.

src/__stories__/hero-story.tsx
```diff
@@ -0,0 +1,111 @@
...
+    return (
+        <Hero
+            background={background}
+            media={mediaComponent}
+            headline={headline ? <Tag type={headlineType}>{headline}</Tag> : undefined}
+            pretitle={pretitle}
+            title={title}
+            description={description}
+            extra={withExtra ? <Placeholder /> : undefined}
+            button={button}
```
atabel, line 63: @yceballost 
buttonprimary + buttonsecondary isn't supported by spec?

src/hero.css.ts
```diff
@@ -0,0 +1,66 @@
...
+export const containerMobile = sprinkles({flexDirection: 'column'});
+
+export const containerDesktop = sprinkles({alignItems: 'center', height: '100%'});
+
+export const actions = style([
+    sprinkles({display: 'flex', flex: 1, alignItems: 'flex-end'}),
+    {marginTop: 24},
+]);
+
+export const layout = style([
```
atabel, line 38: I'd include a comment here explaining why we needed to duplicate this styles from ResponsiveLayout

src/hero.css.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export const container = style([
+    sprinkles({display: 'flex'}),
+    {
+        minHeight: 400,
+        height: fallbackVar(height, '100%'),
+        '@media': {
+            [mq.desktopOrBigger]: {
+                minHeight: 460,
+                height: fallbackVar(height, '100%'),
```
atabel, line 23: isn't this redundant?

src/hero.tsx
```diff
@@ -0,0 +1,213 @@
...
+const Hero = React.forwardRef<HTMLDivElement, HeroProps>(
+    (
+        {height, background = 'default', media, desktopMediaPosition = 'left', dataAttributes, ...rest},
+        ref
+    ) => {
+        const {isTabletOrSmaller} = useScreenSize();
+        const isInsideSlideShow = useIsInsideSlideshowContext();
+        const isInverse = background === 'brand' || background === 'brand-secondary';
+
+        if (isTabletOrSmaller) {
```
atabel, line 128: have you tried to avoid using this hook and use mediaqueries instead?

atabel (approved):

src/hero.tsx
sergivillar: we have two things that make this not possible, first the value of `MediaBorderRadiusProvider` that depends on window size.
Also GridLayout needed in desktop will suppose more code duplication to implement it in Hero
do you have any other idea?

atabel: ok, let it this way for the moment

atabel: if we want to improve this in the future, I think we could:

- GridLayout should fallback to 1 col in mobile, so not a big deal, I think
- For BorderRadius we could use a css class + mediaquery to disable/enable border-radius in images depending on this ancestor class. Maybe too convoluted, not sure if worth it

src/__stories__/hero-slideshow-story.tsx
```diff
@@ -32,8 +35,8 @@ export const Default: StoryComponent = () => (
         withBullets
         inverseBullets={false}
         dataAttributes={{testid: 'hero'}}
-        items={Array.from({length: 3}).map(() => (
-            <HeroComponent />
+        items={Array.from({length: 3}).map((_, idx) => (
+            <HeroComponent idx={idx} />
```
atabel, line 39: I think you need a key here

src/hero.css.ts
```diff
@@ -16,11 +16,11 @@ export const container = style([
     sprinkles({display: 'flex'}),
     {
         minHeight: 400,
-        height: fallbackVar(height, '100%'),
+        height,
         '@media': {
             [mq.desktopOrBigger]: {
                 minHeight: 460,
-                height: fallbackVar(height, '100%'),
+                height,
```
atabel, line 23: I meant, if you have this already defined in base styles, there is no point in repeating it inside media query

atabel (approved):

src/hero.css.ts
atabel: Yes, the fallback was redundant too

PR #671, opened 2023-02-20
feat(DisplayDataCard, DisplayMediaCard): new components

atabel:


src/boxed.tsx
```diff
@@ -21,6 +21,10 @@ type Props = {
     height?: number | string;
 };
 
+type InternalProps = {
+    borderRadius: 8 | 16;
+};
```
atabel, line 26: Allow Boxed with 16 border radius, but only internally, not exposed outside mistica

src/text.tsx
```diff
@@ -42,6 +42,7 @@ export interface TextPresetProps {
     role?: string;
     'aria-level'?: number;
     dataAttributes?: DataAttributes;
+    forceMobileSizes?: boolean;
```
atabel, line 45: used to force the same text sizes in mobile and desktop (as requested by DisplayCards)

playroom/snippets.tsx
```diff
@@ -681,6 +681,60 @@ const cardSnippets: Array<Snippet> = [
             subtitle="Subtitle"
         />`,
     },
+    {
+        group: 'Cards',
+        name: 'DisplayDataCard',
+        code: `
+        <DisplayDataCard
+          icon={
+            <Circle size={40} backgroundColor={skinVars.colors.brandLow}>
```
sergivillar, line 690: does skinVars works in playroom?

src/__stories__/display-media-card-story.tsx
```diff
@@ -0,0 +1,165 @@
...
+    Text2,
+    Inline,
+} from '..';
+
+export default {
+    title: 'Components/Cards/Display media card',
+};
+
+const BACKGROUND_SRC =
+    'https://images.unsplash.com/photo-1622819584099-e04ccb14e8a7?ixlib=rb-4.0.3&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=1932&q=80';
```
sergivillar, line 24: isn't it better to serve this new assets from storybook? to avoid the issue we had last week when this went down

src/boxed.tsx
sergivillar: why no public? if you want some custom component and lives near to a DisplayCard and want the same border radius...

src/card.tsx
```diff
@@ -72,16 +82,11 @@ const CardContent: React.FC<CardContentProps> = ({
...
-                                            as="div"
-                                            regular
-                                            transform="uppercase"
-                                        >
+                                        <Text1 truncate={pretitleLinesMax} as="div" regular>
                                             {pretitle}
                                         </Text1>
                                     )}
-                                    <Text4 truncate={titleLinesMax} as="h1" regular>
+                                    <Text4 truncate={titleLinesMax} as="h3" regular>
```
sergivillar, line 89: why h3?

src/card.tsx
```diff
@@ -393,3 +406,252 @@ export const SnapCard = React.forwardRef<HTMLDivElement, SnapCardProps>(
...
+    isInverse?: boolean;
+};
+
+const CardActionsGroup = ({actions, isInverse}: CardActionsGroupProps): JSX.Element => {
+    return (
+        <Inline space={0}>
+            {actions.map(({onPress, label, Icon}, index) => (
+                <IconButton
+                    size={48}
+                    key={index}
```
sergivillar, line 427: wouldn't it be better to use `useId` instead of index? as this could come form api or cache maybe it changes the order or something else and this could be broken

src/card.tsx
```diff
@@ -393,3 +406,252 @@ export const SnapCard = React.forwardRef<HTMLDivElement, SnapCardProps>(
...
+                    onPress={onPress}
+                    aria-label={label}
+                    className={styles.cardActionIconButton}
+                    style={{display: 'flex'}}
```
sergivillar, line 431: couldn't this flex be move to `cardActionIconButton`?

src/card.tsx
```diff
@@ -393,3 +406,252 @@ export const SnapCard = React.forwardRef<HTMLDivElement, SnapCardProps>(
...
+     */
+    icon?: React.ReactElement;
+    actions?: Array<CardAction>;
+    onClose?: () => void;
+    dataAttributes?: DataAttributes;
+    headline?: React.ReactComponentElement<typeof Tag>;
+    pretitle?: string;
+    pretitleLinesMax?: number;
+    title: string;
+    titleLinesMax?: number;
```
sergivillar, line 495: we don't allow this `xLinexMax`in Hero component becuase design told us this is an a11y issue...  @yceballost ?

src/card.tsx
```diff
@@ -393,3 +406,252 @@ export const SnapCard = React.forwardRef<HTMLDivElement, SnapCardProps>(
...
+                        ) : (
+                            <Box paddingBottom={actions?.length || onClose ? 64 : 0} />
+                        )}
+                        <Box
+                            paddingX={24}
+                            paddingBottom={24}
+                            className={backgroundImage ? styles.displayCardGradient : undefined}
+                        >
+                            <Stack space={24}>
+                                <Stack space={0}>
```
sergivillar, line 579: why an Stack instead of a `div`?

src/text.tsx
sergivillar: wouldnt be weird to have a DisplayCard and a DataCard in desktop with different Text sizes? given the fact that this cards are pretty similar

src/boxed.tsx
atabel: we dont want that (for the moment)
And Design is working on tokenizing border radius to be configurable by brand, so making this public will make the tokenization more dificult

src/text.tsx
atabel: that's a question for Design Core

playroom/snippets.tsx
atabel: hmmm... let me see

src/__stories__/display-media-card-story.tsx
atabel: we have a task for that

src/card.tsx
atabel: it seems no browser support h1 inside article/section to create a new hierarchy. So, maybe h3 is a good choice for cards

src/card.tsx
atabel: I don't see how this can break

src/card.tsx
atabel: hmmm... maybe, I'll check it

src/card.tsx
atabel: but we need this for some Cards configured by CMS

src/card.tsx
sergivillar: for instance, you read from persistence cards 1,2,3
then you fetch from server, and you get cards 2,1,3 (re ordered from cms), react wont update this list

playroom/snippets.tsx
atabel: yes, it works, but we can use just `colors`, so I'll change it

src/card.tsx
atabel: hmmm... I don't see why

src/card.tsx
atabel: hmmm... no, because there is already a class in IconButton adding `display: inline-block` :(

src/card.tsx
atabel: ok, changed

src/card.tsx
yceballost: CMS need this control . Ideally, this prop wouldn't have to exist, at least, in the title. In hero we didn't add it to avoid these AA problems, it is possible in the future we need to include this lines limitation in hero, but we will wait for this need.

sergivillar: why "CMS need" this? if this is really an accessibility issue, this should be avoid and CMS operator be aware of this, dont know why this is needed because of the CMS

kydorn:
We will need support for onPress in the future, as we will both have tappeable cards with action icons on the top-right corner. However, this is not an immediate need, so at least for the digital hub mvp, we are okay with this.

Maybe we could add support for literal '9:10' aspect ratio, but as long as the decimal approach is supported we are also okay with it.

kydorn (approved):

src/card.tsx
sergivillar: well, someone can change the order in the CMS, or if the content come in partial responses... dont know, I think it is safer to use useId than index as a key IMO

src/card.tsx
```diff
@@ -393,3 +403,296 @@ export const SnapCard = React.forwardRef<HTMLDivElement, SnapCardProps>(
...
+
+type AspectRatio = '1:1' | '16:9' | '7:10' | '4:3' | 'auto';
+
+const CSS_ASPECT_RATIO = {
+    '1:1': '1',
+    '16:9': '16 / 9',
+    '7:10': '7 / 10',
+    '4:3': '4 / 3',
+    auto: 'auto',
+} as const;
```
sergivillar, line 447: could these values be taken from image/video files? or this are a differents images?

src/card.tsx
```diff
@@ -393,3 +403,296 @@ export const SnapCard = React.forwardRef<HTMLDivElement, SnapCardProps>(
...
+                                        {(headline || pretitle || title) && (
+                                            <header>
+                                                <Stack space={16}>
+                                                    {headline}
+                                                    <Stack space={4}>
+                                                        {pretitle && (
+                                                            <Text2
+                                                                forceMobileSizes
+                                                                truncate={pretitleLinesMax}
+                                                                as="div"
```
sergivillar, line 628: if this is a pretitle, should this be read before the title? h2? or after the title h4?

src/text.tsx
sergivillar: @yceballost ?

sergivillar (approved):

src/card.tsx
atabel: @yceballost, @aweell, there is a list of supported aspect ratios, or at least recomended ones? I've copied these ones from Image/Video component, and I think they were specified in the specs, but I don't know if we want to recommend the same aspectRatios for DisplayMediaCards, a different set, or nothing at all

src/card.tsx
atabel: I wouldn't give it an h category

src/card.tsx
aweell: Yes, there's a different set of recommended values:

* 9:10
* 1:1
* 16:9
* Custom value

We should use these, instead of the Image defaults.

Screenshot 2023-02-24 at 10 28 22

src/card.tsx
aweell: Would't be too risky to make that assumption? A card can be the only element in a main section without a title (h2), or inside a subsection (h4), since the semantic hierarchy of an element only can be defined based on the layout that element is placed, maybe we shoud expose the title hierarchy? Or...avoid using a heading for the moment?

src/text.tsx
yceballost: @HectorSancho

HectorSancho: In the tests we did, the desktop presets grew excessively and the text at maximum broke the component or was disproportionate when they appeared in a carousel and compared to other components. 

We believe that it scales better and also works better visually by limiting the scaling of the text-presets in desktop resolutions.

aweell (approved):

src/card.tsx
atabel: yes, it won't be the perfect h in all cases, but I think it would be worse not having an h at all

PR #675, opened 2023-02-22
Now the header layout has optional header

wyunreal:
This use case is required for mobile apps that uses a native title bar and a webview, the header layout can delegate the header to the native titlebar.


src/__stories__/header-story.tsx
```diff
@@ -55,6 +55,35 @@ export const Default: StoryComponent = () => {
...
+                            title="Bills"
+                            breadcrumbs={[{title: 'Account', url: '/consumptions'}]}
+                        />
+                    }
+                    extra={<Placeholder />}
+                />
+            </div>
+            <ResponsiveLayout>
+                <Stack space={16}>{extraSideBySideCheckbox}</Stack>
+            </ResponsiveLayout>
```
atabel, line 80: I'd use story args instead, see other stories, like card stories for example

atabel (approved):

src/header.tsx
```diff
@@ -155,7 +155,11 @@ export const MainSectionHeader: React.FC<MainSectionHeaderProps> = ({title, desc
 type HeaderLayoutProps = {
     isInverse?: boolean;
     breadcrumbs?: RendersNullableElement<typeof NavigationBreadcrumbs>;
-    header: React.ReactNode; // intentionally not forced to RendersElement<typeof Header> to allow skeletons for example
+    /**
+     * Intentionally not forced to RendersElement<typeof Header> to allow skeletons for example
+     * The header is optional in order to allow webviews to delegate the header visualization to the surrounding native app.
```
atabel, line 160: Please, include this same explanation in the story. Stories should work as documentation, so it's a good place to explain this case

src/__stories__/header-story.tsx
wyunreal: Yes, but as the story is placed in the same file, I prefer to migrate it completely, not only this new story and it can be done in a maintenance PR if possible ;)

Winde (approved):

src/header.tsx
wyunreal: done ;)

src/__stories__/header-story.tsx
atabel: fair

src/header.tsx
atabel: I meant in the html, for the message to appear in the story

src/__stories__/header-story.tsx
```diff
@@ -55,6 +55,39 @@ export const Default: StoryComponent = () => {
...
+        </Stack>
+    );
+};
+
+NoHeader.storyName = 'Header layout with no header';
```
yceballost, line 85: why not `header layout`?

wyunreal: I think My title is more clear to understand ;)

src/__stories__/header-story.tsx
```diff
@@ -76,18 +89,22 @@ export const NoHeader: StoryComponent = () => {
                 />
             </div>
             <ResponsiveLayout>
-                <Stack space={16}>{extraSideBySideCheckbox}</Stack>
+                <Stack space={16}>
+                    <Text2 medium>Documentation:</Text2>
+                    <Text1 medium={false}>
```
atabel, line 94: medium false?
```suggestion
                    <Text1 regular>
```

atabel (approved):

src/__stories__/header-story.tsx
atabel: fixed

PR #676, opened 2023-02-23
feat(colors): new backgroundBrandSecondary

atabel:


src/skins/telefonica.tsx
```diff
@@ -181,7 +182,7 @@ export const getTelefonicaSkin: GetKnownSkin = () => {
             appBarBackground: palette.darkModeGrey,
             background: palette.darkModeBlack,
             backgroundContainer: palette.darkModeGrey,
-            backgroundBrand: palette.darkModeBlack,
```
aweell, line 184: Should be backgroundBrand kept instead of being replaced?

atabel: yes!

aweell (approved):

PR #677, opened 2023-02-23
Menu component using a portal to render the dropdown

wyunreal:


src/__screenshot_tests__/menu-screenshot-test.tsx
```diff
@@ -0,0 +1,35 @@
+import {openStoryPage, screen} from '../test-utils';
```
Winde, line 1: Worth it to test a menu inside one of the components that were interacting with the visibility: hidden? So that any future changes to menu or data-card,etc.. will not have similar wrong interaction?

Winde (approved):

src/menu.tsx
```diff
@@ -1,15 +1,15 @@
 import * as React from 'react';
 import classnames from 'classnames';
-import {assignInlineVars} from '@vanilla-extract/dynamic';
 import {ESC, TAB} from './utils/key-codes';
 import {cancelEvent, getPrefixedDataAttributes} from './utils/dom';
 import Overlay from './overlay';
 import * as styles from './menu.css';
-import {sprinkles} from './sprinkles.css';
+import {Portal, useWindowSize} from '.';
```
sergivillar, line 7: inside mistica I think that we always import things from the component's file `./Portal` and not from the index

src/menu.tsx
```diff
@@ -54,42 +62,50 @@ const Menu: React.FC<MenuProps> = ({renderTarget, renderMenu, width, position =
...
+        const availableSpaceOnBottom = windowSize.height - bottomTarget - MARGIN_THRESHOLD;
+        const availableSpaceOnTop = topTarget - MARGIN_THRESHOLD;
+        const menuFitsOnBottom = availableSpaceOnBottom > heightMenu;
+        const menuFitsOnTop = availableSpaceOnTop > heightMenu;
+
+        if (menuFitsOnBottom) {
             setItemsComputedProps({
-                top: '100%',
+                left: leftDirection,
+                top: `${bottomTarget}px`,
```
sergivillar, line 80: unneeded pc units, this could be just numbers

src/menu.tsx
```diff
@@ -54,42 +62,50 @@ const Menu: React.FC<MenuProps> = ({renderTarget, renderMenu, width, position =
...
+
+        const leftDirection =
+            position === 'left' ? `${leftTarget}px` : `${leftTarget + widthTarget - width}px`;
+
+        const availableSpaceOnBottom = windowSize.height - bottomTarget - MARGIN_THRESHOLD;
+        const availableSpaceOnTop = topTarget - MARGIN_THRESHOLD;
+        const menuFitsOnBottom = availableSpaceOnBottom > heightMenu;
+        const menuFitsOnTop = availableSpaceOnTop > heightMenu;
+
+        if (menuFitsOnBottom) {
```
sergivillar, line 77: previously we had 3 branches to draw the menu, now we have 4, is that new branch (if-else) needed?

src/menu.tsx
atabel: yes please, this avoids circular dependencies

src/menu.tsx
```diff
@@ -1,15 +1,15 @@
...
 import {ESC, TAB} from './utils/key-codes';
 import {cancelEvent, getPrefixedDataAttributes} from './utils/dom';
 import Overlay from './overlay';
 import * as styles from './menu.css';
-import {sprinkles} from './sprinkles.css';
+import {Portal, useWindowSize} from '.';
 
 import type {DataAttributes} from './utils/types';
 
-const MAX_HEIGHT_DEFAULT = 416;
```
atabel, line 12: I think the Menu should have a max height to avoid edge cases where there are to much items in the menu to fit in the screen (in small screens) or showing a huge list (in large screens). In the select, for example, we have a max height

atabel (approved):

src/__screenshot_tests__/menu-screenshot-test.tsx
wyunreal: Update: added a story and a test case for this.

atabel: It's a good idea to test it here, yes

atabel (approved):

src/menu.tsx
```diff
@@ -148,35 +166,39 @@ const Menu: React.FC<MenuProps> = ({renderTarget, renderMenu, width, position =
...
+                                ...assignInlineVars({
+                                    ...(itemsComputedProps
+                                        ? {
+                                              [styles.vars.top]: itemsComputedProps.top,
+                                              [styles.vars.bottom]: itemsComputedProps.bottom,
+                                              [styles.vars.left]: `${itemsComputedProps.left}px`,
+                                              [styles.vars.transformOrigin]:
+                                                  itemsComputedProps.transformOrigin,
+                                              [styles.vars.maxHeight]: `${itemsComputedProps.maxHeight}px`,
+                                              [styles.vars.width]: `${width}px`,
```
sergivillar, line 191: some units with `px` and others as numbers

src/menu.css.ts
```diff
@@ -1,40 +1,41 @@
...
 const width = createVar();
-const transformOrigin = createVar();
 const maxHeight = createVar();
+const transformOrigin = createVar();
 
 export const vars = {
     top,
     bottom,
-    right,
+    left,
```
sergivillar, line 15: is this change needed?

src/menu.css.ts
```diff
@@ -1,40 +1,41 @@
...
         bottom,
-        right,
+        left,
         width,
-        transformOrigin,
         maxHeight,
+        transformOrigin,
+        display: 'flex',
+        flexDirection: 'column',
+        position: 'absolute',
```
sergivillar, line 36: move to sprinkles (position was already there before), is display flex needed?

src/menu.tsx
wyunreal: Yes, and its intentional, I prefer to have a number when I want to restrict the type only to numbers, because I dont need "auto" or "100%".

src/menu.css.ts
wyunreal: Yes, because its easier to understand by anyone a left oriented coordinate system.

src/menu.css.ts
wyunreal: Display flex is needed to make the menu items adopt all the horizontal space available for the menu.

wyunreal: Moving these props to sprinkles ;)

sergivillar (approved):

PR #679, opened 2023-02-27
remove 12:5 aspect ratio

yceballost:
Removed 12:5 aspect ratio from default values


src/video.tsx
```diff
@@ -8,12 +8,11 @@ import * as styles from './video.css';
 
 import type {DataAttributes} from './utils/types';
 
-export type AspectRatio = '1:1' | '16:9' | '12:5' | '4:3';
+export type AspectRatio = '1:1' | '16:9' | '4:3';
```
kydorn, line 11: can we use the same type we use in Image?

kydorn (approved):

src/video.tsx
yceballost: mmm, I was thinking about if 21:9 has to be only in video, but I agree to keep it in both. Do you think the video component has to include 7:10, maybe be quite weird for video?

atabel (approved):

PR #683, opened 2023-03-02
Fix select issue in SSR

sergivillar:
wait till #684 to merge it and get the new common hook


src/select.tsx
```diff
@@ -259,6 +263,14 @@ const Select: React.FC<SelectProps> = ({
         }
     }, [autoFocus]);
 
+    React.useEffect(() => {
+        // We use this Ref to always use the native variant in the first render, this way we avoid rehydration issues when using SSR
```
atabel, line 267:
```suggestion
        // We use this Ref to always use the native variant in the first render, this way we avoid hydration issues when using SSR
```

atabel (approved):

Marcosld (approved):

PR #685, opened 2023-03-06
feat: [PTIECBC03-3872] POC Brazil: Adding bleed prop to Header component

geremiascorrea-tlf:
This PR includes the changes in Vivo alteration requests in Header component. In this case, It was only necessary to make an addition of a prop bleed. The developer can change dynamically this prop. Therefore, it was also added the prop into Storybook file. The tests were considered, but it was difficult to perform the process and it was also concluded that they may not be necessary for this scenario.

The changes are according to the changes requested at this link: https://www.figma.com/file/Xm3r8nRNlbAiZNlkaH5YX1/%F0%9F%94%B8-Header-Component-Specs?node-id=0%3A1&t=MOfw03jc7Nh9qZWk-0


src/header.tsx
```diff
@@ -135,6 +135,7 @@ type HeaderLayoutProps = {
     sideBySideExtraOnDesktop?: boolean;
     children?: void;
     dataAttributes?: DataAttributes;
+    bleedValue?: number;
 };
```
yceballost, line 140: Hi Geremias! Thanks for this contribution, I really appreciate it!

I'm a designer, so, my comment will be focused on design specs, not on the tech side

what I have tested, bleed is not wrapping the content, this cause the following problem
image

As specs define, bleed has to wrap the content inside (avoiding extra prop)
image
If the content grows, the color zone will grows too
image

Your solution can cause problems with dynamic content or when users change the size of the text by accessibility.

The value of bleed is closed in the design specs but I like the idea to have the possibility to set a custom height(by default is defined in 40px) for bleed prop. I will add to the specs the possibility to have a custom height for this prop.

yceballost (changes requested):

src/header.tsx
geremiascorrea-tlf: Hello! First, thanks for the feedback. Despite Figma being mentioned, there were some changes in strategies according to what was established, jointly with those responsible for Vivo. Initially, it was imagined that it would be almost necessary to create a new component, well fixed in the layout proposed there. However, discussing with those responsible for Mística, it was discovered that practically everything (with the exception of bleed) could be done with the current format of what the Header/HeaderLayout allowed. As this slot issue is possible with the HeaderLayout called "extra", the possibilities of the internal content are extremely wide, making it maybe impossible to clearly control the size/height of the bleed, in a fixed way as initially proposed. With that in mind, we thought of allowing us to set the desired height when creating/declaring the component (or at least no making sense), even though it might have some limitations with variations in screen formats. As a POC, we have space for error and testing and these definitions can again be changed, of course. However, we did internal alignments and it was accepted in the following format. Do you believe it is possible/permissible to leave it like that or is it necessary to re-analyze it more faithfully to what was initially thought?

geremiascorrea-tlf: Hello, I made some changes based on previous feeback. Looking back, it made sense to ignore the bleed behavior in the header's upper scope. With that, I considered the same from the beginning of the extra scope/prop, which is aligned with what was requested in Figma. Thinking about scalability, for now, I left the bleed with two props. The first, "bleed", just triggers its use or not. If activated, its default value is 40. However, with the "bleedValue" prop, it is possible to change its value to a more suitable context.

I considered making use of the bleedValue with using classNames, but the limitation of values by the sprinkles made it that way, it seemed more interesting, because I don't know if it would be a case of changing this file.

Note: this is the first PR on our side for Mística, so we don't have a perfect idea of the patterns, the changes that should and shouldn't be made, etc. If you have tips, it's valid as PR feedback too.

src/header.tsx
```diff
@@ -135,6 +135,8 @@ type HeaderLayoutProps = {
     sideBySideExtraOnDesktop?: boolean;
     children?: void;
     dataAttributes?: DataAttributes;
+    bleed?: boolean;
+    bleedValue?: number;
 };
```
yceballost, line 141: I found a bug when you use a card inside extra prop + bleed prop
image

yceballost (changes requested):

src/header.tsx
geremiascorrea-tlf: Good afternoon and thank you again for your feedback,

Regarding the desktop issue, yes, I ended up missing the fact that I changed it only for the context of tablet screens or smaller. I already quickly corrected this point. With the conflicts I will soon do the same. I'll upload the changes soon.

Regarding the issue of the card mixing with the background that was mentioned, is this perhaps more a responsibility to the component being applied to the extra? In case some type of border is applied to it according to preference (which in a given case, it may be preferable to leave it like that, for example). I say this because I was thinking about the generic component itself, opening up implementations.

src/header.tsx
```diff
@@ -164,6 +164,8 @@ type HeaderLayoutProps = {
     sideBySideExtraOnDesktop?: boolean;
     children?: void;
     dataAttributes?: DataAttributes;
+    bleed?: boolean;
+    bleedValue?: number;
```
atabel, line 168: Do we want this to be customizable? I don't think so. Design spec says it should be always 40px (@yceballost ?)

atabel (commented):
This is broken:
image

src/header.tsx
atabel: I'd make the `HeaderLayout` component to disable the inverse variant for the `extra` when `bleed` is enabled

src/header.tsx
geremiascorrea-tlf: Hello, I apologize for the delay and not having noticed this error (I didn't put content below in the playroom and I didn't test it in the storybook on the last commit). I'm working on this problem today and I haven't come up with a reasonable solution for this bleed that doesn't influence the size of the external div/layout as a whole, so I haven't commented on anything. 

About Figma, only the fixed part of 40 px was really requested. The idea was to allow, in addition to the request (by default), configurable in case you wanted to use another distance. However, even by request, it is possible to allow only the fixed 40 px and, if requested in the future, do it dynamically.

Note: if you have an idea of a way to apply this bleed exactly to the point 40 px below the beginning of the extra's content, I appreciate it. It's a context that doesn't seem so trivial to me, because the background includes header + extra and both don't have fixed sizes, so it's hard to predict. The way I had thought of it, using the height property at that point unfortunately has side effects.

geremiascorrea-tlf:
I rebuilt the solution, thinking about separating the header and extra contents into different divs, thus controlling the case with isInverse, as indicated before.

However, as it is also requested to increase the background by 40 pixels, behind the content of the add-on (which is not so trivial). I thought of using translate to go up the content of the extra, compensating with the padding-bottom. The case of "sideBySideExtraOnDesktop" was also considered, which does not fit in a case with bleed and therefore does not enter the bleed application solution.

I noticed that when applying the bleed, as I uploaded the extra's content, I couldn't compensate for the decrease in the extra's div below. My idea was to try to use the height to force it to decrease 40 pixels, but without success. I went up anyway because that way maybe you have more time to analyze and bring a constructive answer on our side, given the knowledge you have. Tomorrow I will enter later and it is more feasible to have that extra time of possible return. Other than that, I think the solution is now better on track, avoiding greater side effects.

atabel:
Hi, @geremiascorrea-tlf, I think this could be a simpler implementation approach:

We need to apply bleed when `bleed` prop is provided, `extra` prop is provided and `isTabletOrSmaller || !!sideBySideExtraOnDesktop`

When we need to apply bleed, instead of rendering the `extra` inside the `<stack>` next to `{header}`, we can render it below the `</ResponsiveLayout>` as a separate `<div>` (let's call it the bleed div). Inside the bleed `<div/>` we can have 2 elements:

- A `position: absolute` `<div>` with a `backgroundBrand` background and a width of 40px (the bleed background)
    - Note that we need the bleed div to be `position: relative` for this to work, and we'll probably need to tweak the zIndex of the bleed background div.
- A `<ResponsiveLayout>{extra}</ResponsiveLayout`

Something like this:
```js
</ResponsiveLayout>

<div className={styles.bleed}>
    <div className={styles.bleedBackground}/>
    <ResponsiveLayout>{extra}</ResponsiveLayout>
<div>
```
image

geremiascorrea-tlf:
> Hi, @geremiascorrea-tlf, I think this could be a simpler implementation approach:
> 
> We need to apply bleed when `bleed` prop is provided, `extra` prop is provided and `isTabletOrSmaller || !!sideBySideExtraOnDesktop`
> 
> When we need to apply bleed, instead of rendering the `extra` inside the `<stack>` next to `{header}`, we can render it below the `</ResponsiveLayout>` as a separate `<div>` (let's call it the bleed div). Inside the bleed `<div/>` we can have 2 elements:
> 
> * A `position: absolute` `<div>` with a `backgroundBrand` background and a width of 40px (the bleed background)
>   
>   * Note that we need the bleed div to be `position: relative` for this to work, and we'll probably need to tweak the zIndex of the bleed background div.
> * A `<ResponsiveLayout>{extra}</ResponsiveLayout`
> 
> Something like this:
> 
> ```js
> </ResponsiveLayout>
> 
> <div className={styles.bleed}>
>     <div className={styles.bleedBackground}/>
>     <ResponsiveLayout>{extra}</ResponsiveLayout>
> <div>
> ```
> 
> image

Hi, @atabel, seems to be a solution as well. But just to confirm, this last collaboration (from yesterday) release is very similar. Not a viable option? In the meantime, I will design the one mentioned by you as well.

yceballost (changes requested):

src/__stories__/header-story.tsx
```diff
@@ -29,11 +29,13 @@ export const Default: StoryComponent = () => {
         true
     );
     const [withBreadcrumbs, breadcrumbsCheckbox] = useCheckbox('With breadcrumbs (desktop only)', true);
+    const [bleed, bleedCheckbox] = useCheckbox('Bleed', false);
     return (
         <Stack space={16}>
             <div data-testid="header-layout">
                 <HeaderLayout
                     isInverse={isInverse}
```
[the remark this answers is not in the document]
geremiascorrea-tlf: Yes, there really was a logical flaw there, I've fixed it!

src/header.tsx
```diff
@@ -164,6 +164,7 @@ type HeaderLayoutProps = {
     sideBySideExtraOnDesktop?: boolean;
     children?: void;
     dataAttributes?: DataAttributes;
+    bleed?: boolean;
 };
```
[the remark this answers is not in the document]
geremiascorrea-tlf: Exactly, I already did some tests and came up with a solution to not generate extra space anymore. Also resolved!

geremiascorrea-tlf:
Good afternoon! I ask you to retest this last commit (ded81d4). At first I corrected the two problems found and I think it is now perfectly functional! I will still try to develop the solution indicated there (for safety), in case you really think it is necessary. Please provide feedback on this.

Also, sorry for the multiple versions submitted and multiple returns required.

src/header.tsx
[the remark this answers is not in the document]
geremiascorrea-tlf: Wow, thank you so much for the feedback (and agility). I was ignoring tests with this prop until now, my fault. Not fully knowing the component we are making changes to is tricky. But with each commit it becomes clearer all the possible features. I believe I have fixed the problem and maybe there is no more. Please retest when possible (commit 1d47bf3).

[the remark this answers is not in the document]
geremiascorrea-tlf: Hey, I created another solution, based on what was indicated above and in a simpler way than this branch. I separated header and extra and applied the bleed with a linear-gradient fixed by 40 pixels. As far as I've checked, it's not generating problems with validations. If you want to check the solution, it's in the "vivo-header-bleed-alt" branch (https://github.com/Telefonica/mistica-web/tree/vivo-header-bleed-alt).

It's for if you consider it valid, I appreciate any feedback.

src/header.tsx
```diff
@@ -173,47 +174,69 @@ export const HeaderLayout: React.FC<HeaderLayoutProps> = ({
...
+                isInverse={isInverse}
+                dataAttributes={{'component-name': 'HeaderLayout', ...dataAttributes}}
+            >
+                <div style={{paddingBottom: bleed && isInverse && extra ? 40 : 0}}>
+                <OverscrollColor />
+                {isTabletOrSmaller ? (
+                    <div>
+                        <Box paddingTop={header ? 32 : 0} paddingBottom={24}>
+                            <Stack space={24}>
+                                {header}
```
atabel, line 193: the extra should be inside this stack when `bleed` is false

src/header.tsx
```diff
@@ -173,47 +174,69 @@ export const HeaderLayout: React.FC<HeaderLayoutProps> = ({
...
-                    </Stack>
-                </Box>
+                    </Box>  
+                )}
+                </div>
+            </ResponsiveLayout>
+            {extra && (isTabletOrSmaller || !sideBySideExtraOnDesktop) && (
+                <ResponsiveLayout 
+                    isInverse={!!(isInverse && !bleed)}
+                    dataAttributes={{'component-name': 'HeaderLayout', ...dataAttributes}}
```
atabel, line 225:
```suggestion

```
Remove this dataAttributes here

src/header.tsx
```diff
@@ -173,47 +174,69 @@ export const HeaderLayout: React.FC<HeaderLayoutProps> = ({
...
+                >
+                    <div style={{
+                        transform: bleed && isInverse ? `translateY(-40px)` : `translate(0, 0)`,
+                        marginBottom: bleed && isInverse ? '-40px' : '0px',
```
atabel, line 229: I'd prefer a solution without transform translate nor negative margins, like the one I proposed

geremiascorrea-tlf: I understand. If possible, look at commit edits on this branch (https://github.com/Telefonica/mistica-web/tree/vivo-header-bleed-alt). In case it seems more interesting, I can open the PR of this one. It is similar to the process you indicated. I believe it is much cleaner and without changing as many code properties as desired. The other two changes mentioned now, I can also apply there.

geremiascorrea-tlf: In the case of the aforementioned branch (vivo-header-bleed-alt), I can make some changes so that if the bleed is false, I can automatically get the extra in the first ResponsiveLayout. If the prop bleed is activated, it will pull the code snippet with the second ResponsiveLayout.

yceballost (approved):
Everything works as expected from the design side! thanks a lot for contributing to Mística <3

yceballost:
I found buggy behavior with gradient backgrounds
https://rebrand.ly/ic6056c

This is not important at the moment because O2 Classic skin will be removed soon

geremiascorrea-tlf:
> I found buggy behavior with gradient backgrounds https://rebrand.ly/ic6056c
> 
> This is not important at the moment because O2 Classic skin will be removed soon

I had thought this might be an issue with gradient backgrounds, but I didn't know any of the possible backgrounds would be like that. Separating into two divs to deal with this issue (which are the most reasonable and were recommended), generates this problem for these cases. 

As I understand it, we should ignore it for now, but it's good knowledge of these cases.

src/header.tsx
```diff
@@ -173,47 +174,63 @@ export const HeaderLayout: React.FC<HeaderLayoutProps> = ({
...
-                        {header}
-                        {extra}
-                    </Stack>
-                </Box>
-            ) : sideBySideExtraOnDesktop ? (
-                <Box paddingTop={breadcrumbs ? 16 : 48} paddingBottom={48}>
-                    <GridLayout
-                        template="6+6"
-                        left={
+        <div>
```
atabel, line 182: I think you can replace this `<div>` with a fragment (`<>`)

src/header.tsx
```diff
@@ -173,47 +174,63 @@ export const HeaderLayout: React.FC<HeaderLayoutProps> = ({
...
+            {bleed && extra && (isTabletOrSmaller || !sideBySideExtraOnDesktop) && (
+                <ResponsiveLayout
+                    backgroundColor={
+                        isInverse
+                            ? `linear-gradient(to bottom, ${vars.colors.backgroundBrand} 40px, ${vars.colors.background} 0%)`
+                            : vars.colors.background
+                    }
+                >
+                    <Box paddingBottom={isTabletOrSmaller ? 24 : 48}>
+                        <Stack space={isTabletOrSmaller ? 24 : 32}>{extra}</Stack>
```
atabel, line 229: this `<Stack/>` is not needed

src/header.tsx
```diff
@@ -173,47 +174,63 @@ export const HeaderLayout: React.FC<HeaderLayoutProps> = ({
...
+            </ResponsiveLayout>
+            {bleed && extra && (isTabletOrSmaller || !sideBySideExtraOnDesktop) && (
+                <ResponsiveLayout
+                    backgroundColor={
+                        isInverse
+                            ? `linear-gradient(to bottom, ${vars.colors.backgroundBrand} 40px, ${vars.colors.background} 0%)`
+                            : vars.colors.background
+                    }
+                >
+                    <Box paddingBottom={isTabletOrSmaller ? 24 : 48}>
```
atabel, line 228: @yceballost, it feels weird to have a `paddingBottom` when `bleed` is enabled, what do you think?
image

atabel:
> I found buggy behavior with gradient backgrounds https://rebrand.ly/ic6056c
> 
> This is not important at the moment because O2 Classic skin will be removed soon

This could be fixed by using the `backgroundFeedbackBottom` color instead of `backgroundBrand` for the bleed part. But I think we can ignore the o2-classic case, as we are going to deprecate it

src/header.tsx
yceballost: In the bleed specs this padding doesn't exists. I think is better remove this padding to avoid mix spacing with vertical rhythm

src/header.tsx
geremiascorrea-tlf: Fixed!

geremiascorrea-tlf:
@atabel, I noticed there was a comment about the paddingBottom. Just explaining that the paddingBottom I used in the bleed part was to match the original behavior, which asked for 24 px in the case of isTabletOrSmaller or 48 in other cases. Creating the space between the end of the extra (and HeaderLayout, as a whole) and the next component applied below.

I was in doubt if I accidentally "resolve", because I had a small crash on the screen and I may have miss clicked in an attempt to unlock.

Other than that, if you can reevaluate again, I appreciate it! Thanks for the feedback so far!

src/header.tsx
atabel: ok, please @geremiascorrea-tlf, remove this `<Box>`

atabel:
> @atabel, I noticed there was a comment about the paddingBottom. Just explaining that the paddingBottom I used in the bleed part was to match the original behavior, which asked for 24 px in the case of isTabletOrSmaller or 48 in other cases. Creating the space between the end of the extra (and HeaderLayout, as a whole) and the next component applied below.
> 
> I was in doubt if I accidentally "resolve", because I had a small crash on the screen and I may have miss clicked in an attempt to unlock.
> 
> Other than that, if you can reevaluate again, I appreciate it! Thanks for the feedback so far!

yes, as @yceballost says here https://github.com/Telefonica/mistica-web/pull/685#discussion_r1137053450, that spacing isn't in the specs, so I think we should remove it

src/__stories__/header-story.tsx
atabel, line 1: @geremiascorrea-tlf, would you mind to add a screenshot test to cover the bleed case?

You can add it in a different PR if you prefer ;)

src/header.tsx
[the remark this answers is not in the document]
atabel: what I mean, is the component should render the same with or without bleed if isInverse is false.
In the video you can see the bottom padding changes when I switch the bleed prop and isInverse is false

src/__stories__/header-story.tsx
atabel: ok, it's fine to me, you can do it in a different PR.

To generate a screenshot, you need to write a new test in `header-screenshot-test.tsx`. That test will:
- open the header story
- check the bleed checkbox in the story
- take a screenshot like this:

```ts
const image = await story.screenshot();
```
Then you can run the test with:

```sh
yarn test-acceptance header
```

The first time the test runs, the screenshot file (.png) will be generated. Just review the screenshot to see if it's the expected result and commit/push it.

The following runs of the test will check the screenshot with the one stored in the repo, if both match, the test will pass, if there is any difference, the test will fail.

src/header.tsx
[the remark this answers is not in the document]
geremiascorrea-tlf: Oh ok, it was the padding I noticed. In 1 minute I'll upload the change about it. Thanks for the feedback about that, I made the previous change and don't notice it.

src/header.tsx
```diff
@@ -178,6 +178,10 @@ export const HeaderLayout: React.FC<HeaderLayoutProps> = ({
 }) => {
     const {isTabletOrSmaller} = useScreenSize();
 
+    const isBleedActivated = (): boolean => {
+        return bleed && isInverse;
+    };
```
atabel, line 183: this can be a const, no need a function.

geremiascorrea-tlf: Sure, fixed!

PR #688, opened 2023-03-10
WEB-796 fix safari and media card border radius

sergivillar:


src/card.css.ts
```diff
@@ -9,7 +9,12 @@ export const actions = style([
     {marginTop: 16},
 ]);
 
-export const boxed = sprinkles({display: 'flex'});
+export const boxed = style([
+    sprinkles({display: 'flex'}),
+    {
+        isolation: 'isolate', // Needed to preserve border-radius with Video component and Safari
```
sergivillar, line 15: this create a new stacking context, similar to the old hack of `translateZ(0)`

sergivillar:
> **Screenshot tests report**
> 
> ❌ carousel / carousel-mobile

I think this test could be updated, WDYT @atabel ?

src/card.css.ts
atabel: oh, and Safari support this? from which version?

atabel (approved):

atabel:
> > **Screenshot tests report**
> > x carousel / carousel-mobile
> 
> I think this test could be updated, WDYT @atabel ?

to be fair, I don't see the difference, so yes

src/card.css.ts
sergivillar: since safari 8 :O

PR #693, opened 2023-03-20
chore(HeaderLayout): add screenshot test to cover the bleed case

geremiascorrea-tlf:
Creation and generation of screenshot tests for the prop bleed case, referring to the HeaderLayout component.

A new test was generically made for mobile device and desktop context (although this one, due to the extra prop on the right side would not be applied), which generated two new screenshots. Another two screenshots needed to be regenerated due to the interference of the bleed prop value.

Remembering that the component itself is already up in the master branch, but it was commented that the creation should be carried out in another PR, which I am requesting to upload at this moment.


src/__stories__/header-story.tsx
```diff
@@ -29,7 +29,7 @@ export const Default: StoryComponent = () => {
         true
     );
     const [withBreadcrumbs, breadcrumbsCheckbox] = useCheckbox('With breadcrumbs (desktop only)', true);
-    const [bleed, bleedCheckbox] = useCheckbox('Bleed', false);
+    const [bleed, bleedCheckbox] = useCheckbox('Bleed', true);
```
atabel, line 32: Instead of changing the default value to true, you should check the checkbox in the bleed test.

The previous HeaderLayout screenshot tests should not change

atabel (changes requested):

src/__stories__/header-story.tsx
```diff
@@ -81,6 +81,7 @@ export const NoHeader: StoryComponent = () => {
             <div data-testid="header-layout">
                 <HeaderLayout
                     isInverse
+                    bleed
```
atabel, line 84: please remove this

src/__screenshot_tests__/__image_snapshots__/header-screenshot-test-tsx-header-in-mobile-ios-is-inverse-true-is-error-amount-true-1-snap.png
yceballost, line 1: is this change expected?

geremiascorrea-tlf: I didn't know it wasn't ideal to modify previous snapshots. I've done it in the format that I believe is expected.

src/__screenshot_tests__/__image_snapshots__/header-screenshot-test-tsx-header-with-bleed-1-snap.png
atabel, line 1: I don't see the bleed here :/

src/__stories__/header-story.tsx
```diff
@@ -18,7 +18,11 @@ export default {
     },
 };
 
-export const Default: StoryComponent = () => {
+type Args = {
+    bleedValue: boolean;
+};
```
atabel, line 23: Please don't use Args in this story.

It's fine (event preferred) to use `Args` in stories, but this story is using inline controls (`useTextField`, `useCheckbox`, etc) and it isn't a good idea to mix both (args + inline controls). I suggest you to just check the bleed checkbox in your test.

In the future we can migrate this story to use `Args` instead, but that would be a different task

geremiascorrea-tlf: Oh, okay. I hadn't thought of it that way. I will check how to perform correctly and even today I upload the changes. Thanks for the feedback and explanation.

src/__screenshot_tests__/header-screenshot-test.tsx
```diff
@@ -76,3 +76,18 @@ test('Header with truncation and color overrides', async () => {
...
 });
+
+test.each(DEVICES)('Header with bleed', async (device) => {
+    await openStoryPage({
+        id: 'components-headers-header--default',
+        device,
+    });
+
+    const bleedCheckbox = await screen.findByLabelText('Bleed');
+    await bleedCheckbox.click();
```
atabel, line 87: I'd also click on the "Extra content placed on the right in desktop" checkbox to be able to screenshot the `bleed` case in desktop

geremiascorrea-tlf: Of course, I wasn't thinking about that. Fixed!

yceballost (approved):

PR #700, opened 2023-03-27
feat(Inline): allow wrap and negative space

atabel:
fixes WEB-782 and WEB-793


src/__stories__/tag-story.tsx
```diff
@@ -42,61 +42,63 @@ export const Default: StoryComponent<Args> = ({label: labelFromArgs}) => {
...
-                    {getLabel('Success')}
-                </Tag>
-                <Tag Icon={IconAlertRegular} type="warning">
-                    {getLabel('Warning')}
-                </Tag>
-                <Tag Icon={IconCloseRegular} type="error">
-                    {getLabel('Error')}
-                </Tag>
-            </Container>
+        <div data-testid="tags" style={{display: 'inline-flex'}}>
```
atabel, line 45: Added this div to make the screenshot pass. Otherwise, puppeteer takes a screenshot including the `<Inline/>` negative margins

src/inline.css.ts
```diff
@@ -16,15 +18,24 @@ export const fullWidth = style([
...
+
 export const noFullWidth = style([
     inline,
     {
         display: ['inline-flex', 'inline-grid'],
     },
 ]);
 
-globalStyle(`${inline} > div:not(:empty) ~ div:not(:empty)`, {
+globalStyle(`${inline} > div`, {
```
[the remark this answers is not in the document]
atabel: I need the first element to have left margin too, it's simpler than the previous solution, and I think it doesn't break any case

sergivillar (approved):

anabelengp (approved):

src/inline.tsx
```diff
@@ -8,14 +8,35 @@ import * as styles from './inline.css';
...
+        | 'around'
+        | 'evenly';
     alignItems?: 'flex-start' | 'flex-end' | 'center' | 'stretch' | 'baseline';
     children: React.ReactNode;
     className?: string;
     role?: string;
     'aria-labelledby'?: string;
     fullWidth?: boolean;
     dataAttributes?: DataAttributes;
+    wrap?: boolean;
```
yceballost, line 39: should the inline component wrap the content by default?

yceballost (approved):

src/inline.tsx
atabel: I don't think so. And it would be a breaking change

PR #705, opened 2023-03-30
Vivo avatar update

BrunoGuimaraes103:
Adding the border control on the component and running prettier command

figma link: https://www.figma.com/file/jWVdZ8WvB3URODZiKoENMn/Avatar?node-id=20%3A3870&t=d7wx85zHgmqt6Yfi-1

### DIFF
--- a/src/__stories__/avatar-story.tsx
+++ b/src/__stories__/avatar-story.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import {Avatar, IconBrainRegular, IconFireRegular, IconStarFilled, ThemeVariant, skinVars} from '..';
+import { Avatar, IconBrainRegular, IconFireRegular, IconStarFilled, ThemeVariant, skinVars } from '..';
 
 const badgeOptions = ['true', 'false', 'undefined', '0', '1', '5', '10'];
 
@@ -7,16 +7,26 @@ export default {
     title: 'Components/Avatar',
     argTypes: {
         size: {
-            control: {type: 'range', min: 24, max: 128, step: 4},
+            control: { type: 'range', min: 24, max: 128, step: 4 },
         },
         badge: {
             options: badgeOptions,
-            control: {type: 'select'},
+            control: { type: 'select' },
         },
         icon: {
             options: ['undefined', 'IconStarFilled', 'IconFireRegular', 'IconBrainRegular'],
-            control: {type: 'select'},
+            control: { type: 'select' },
         },
+        borderSize: {
+            control: { type: 'range', min: 1, max: 5, step: 0.05 },
+        },
+        borderColor: {
+            defaultValue: 'borderLow',
+            options: ['border', 'borderLow', 'borderHigh', 'borderSelected', ''],
+            mapping: skinVars.colors,
+            control: { type: 'select' }
+        },
+
     },
     parameters: {
         fullScreen: true,
@@ -33,6 +43,9 @@ type Args = {
     badge: string;
     inverse: boolean;
     ariaLabel: string;
+    borderAvatar: boolean;
+    borderColor: 'border' | 'borderLow' | 'borderHigh' | 'borderSelected';
+    borderSize: number;
 };
 
 export const Default: StoryComponent<Args> = ({
@@ -45,10 +58,13 @@ export const Default: StoryComponent<Args> = ({
     hideImage,
     hideInitials,
     ariaLabel,
+    borderAvatar,
+    borderColor = 'borderLow',
+    borderSize,
 }) => {
     // eslint-disable-next-line no-eval
     const badgeValue = badgeOptions.includes(badge) ? eval(badge) : undefined;
-    const Icon = {IconStarFilled, IconFireRegular, IconBrainRegular}[icon];
+    const Icon = { IconStarFilled, IconFireRegular, IconBrainRegular }[icon];
 
     return (
         <ThemeVariant isInverse={inverse}>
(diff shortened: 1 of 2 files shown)


atabel:
I can't open the linked figma :(

BrunoGuimaraes103:
> Adicionando o controle de borda no componente e executando o comando mais bonito
> 
> link figma: https://www.figma.com/file/U4ipIXOk64bdM5tSvaqPKS/%F0%9F%94%B8-Avatar-Specs?node-id=2-61&t=C55keUKC24YRIiwj-0


the link has been corrected

src/avatar.tsx
```diff
@@ -34,6 +34,9 @@ type AvatarProps = {
     badge?: boolean | number;
     'aria-label'?: string;
     dataAttributes?: DataAttributes;
+    borderAvatar?: boolean;
+    borderColor: 'border' | 'borderLow' | 'borderHigh' | 'borderSelected';
+    borderSize?: number;
```
yceballost, line 39: specs say that the border width is 1px. I prefer to limit this value to the moment. If in the future is need it, we could evolve the component to cover this need. Making the possibility open can cause unnecessary inconsistencies through different products in an ecosystem

In the future, we can cover this need through tokens if any brand is required

yceballost (changes requested):

src/__stories__/avatar-story.tsx
```diff
@@ -33,6 +38,8 @@ type Args = {
     badge: string;
     inverse: boolean;
     ariaLabel: string;
+    borderAvatar: boolean;
```
yceballost, line 41: I think this prop can be named as `border`, wdyt @atabel?

yceballost (approved):

src/__stories__/avatar-story.tsx
```diff
@@ -17,6 +17,11 @@ export default {
             options: ['undefined', 'IconStarFilled', 'IconFireRegular', 'IconBrainRegular'],
             control: {type: 'select'},
         },
+        borderColor: {
+            defaultValue: 'borderLow',
```
yceballost, line 21: I think that is not showing defaultValue by default 🤔

https://rebrand.ly/yr48gsu

yceballost (changes requested):

src/avatar.tsx
```diff
@@ -34,6 +34,8 @@ type AvatarProps = {
     badge?: boolean | number;
     'aria-label'?: string;
     dataAttributes?: DataAttributes;
+    border?: boolean;
+    borderColor?: 'border' | 'borderLow' | 'borderHigh' | 'borderSelected';
```
yceballost, line 38: these colors are not defined in the specs, why did you decide to use these colors?

I think is better to let it open, because could have some cases that you need to use the same color of the background

yceballost (changes requested):

src/avatar.tsx
yceballost: yes, borderLow could works in most of cases

BrunoBiagioniNeto: Yaio, nós estamos utilizando os Tokens de Borda nas bordas conforme padrões Mística. Existem Tokens para Bordas e seguimos as definições, podemos abrir se vocês acharem que é válido, mas aí será um uso muito amplo e despadronizado.

Vocês querem deixar aberto? Se sim nós o faremos, mas ao meu ver é perigoso.

yceballost: It is ok, to keep aligned with Avatar border specs that Vivo design team defined, the best option is remove the possibility of change border color. If in the future is need it, we will evolve to allow that.

atabel: yes, please, dont make this configurable.
We only want a boolean prop called `border`, when provided, the component will apply a border of `1px solid ${skinVars.colors.borderLow}`

src/avatar.tsx
```diff
@@ -86,7 +90,13 @@ const Avatar = ({
                 className={classes.avatar}
                 role="img"
                 aria-label={ariaLabel ?? initials}
-                style={{width: size, height: size, color: textColor, background: backgroundColor}}
+                style={{
+                    width: size,
+                    height: size,
+                    color: textColor,
+                    background: backgroundColor,
+                    boxShadow: border,
```
atabel, line 98: why box `shadow` instead of `border`?

atabel (changes requested):

src/avatar.tsx
atabel: hmmm... please, use border, I don't see a good reason for not using it

AlexJCelmer: If using border we need to decide the box-sizing of content-box or border-box to not influence on the size of component. I sugest to use box-sizing: border-box; to not make the component increse its size from 64 to 65 when applying the border.

We will be doing this change right now.

src/avatar.tsx
```diff
@@ -69,11 +69,10 @@ const Avatar = ({
         props.backgroundColor ?? (isInverse ? vars.colors.brandHigh : vars.colors.brandLow);
     const textColor = props.textColor ?? (isInverse ? vars.colors.textPrimaryInverse : vars.colors.brand);
     const [imgLoadError, setImgLoadError] = React.useState(false);
-
     React.useEffect(() => {
         setImgLoadError(false); // reset error state when url changes
     }, [src]);
-
+    const borderColor = props.border ? `1px solid ${vars.colors.borderLow}` : 'none';
```
atabel, line 75: please, rename to `border`, you aren't changing the border color only.
```suggestion
    const border = props.border ? `1px solid ${vars.colors.borderLow}` : 'none';
```
alternatively, you can just inline this in the style object below

src/avatar.tsx
```diff
@@ -90,7 +90,7 @@ const Avatar = ({
                     height: size,
                     color: textColor,
                     background: backgroundColor,
-                    border: borderColor,
+                    border: border,
```
atabel, line 93: you have a linting error here

yceballost (approved):

PR #706, opened 2023-03-31
WEB-775 test upload images

sergivillar:


atabel (commented):
please, try to make both jobs fail (test-ssr and test-acceptance) I want to see if the diff of both jobs are uploaded or not

atabel (approved):

PR #708, opened 2023-04-03
fix(Carousel): only call onPageChange when a true page change has heppen

wyunreal:
Objectives:
- Disallow the onPageChange to be called before the user interacts with the carousel.
- Disallow consecutive calls of the callback due to re renders.
- Disallow more than one callback to be called per each page change.


src/carousel.tsx
```diff
@@ -334,22 +334,38 @@ const BaseCarousel: React.FC<BaseCarouselProps> = ({
...
+                        shownItemIndexes.unshift(idx);
+                    }
                 }
+
+                onPageChange({pageIndex: currentPageIndex, shownItemIndexes});
             }
-            onPageChange({pageIndex: currentPageIndex, shownItemIndexes});
+
+            pageChangeControl.current = {
+                canEmit: !initialActiveItem || pageChangeControl.current.lastPageIndex !== currentPageIndex,
```
Winde, line 364: This seems weird:
canEmit seems to have similar logic to line 346, only adding a condition of !initialActiveItem.

Isn't this redundant?

atabel: Yes, I don't get this logic. Also, why `!initialActiveItem` ?

wyunreal: when initialActiveItem is used, the carousel performs a scroll move to the initial item, this is not caused by the user swiping the carousel, so we don't want to fire a page changed in this case, so thats why I've used the !initialActiveItem.

wyunreal: Its not redundant, as we need to avoid calling the onPageChange for this first automatic transition when initialActiveItem is set and different than 0.

Winde: Let's say: initialActiveItem is A
Let's say: pageChangeControl.current.lastPageIndex !== currentPageIndex is B

canEmit is only used in line 345
canEmit = !A || B

and the condition is:
canEmit && B

so:
(!A || B) && B

so, that's the same as saying B by doing pageChangeControl.current = {...}

would it not make more sense to be explicit with this logic instead of ofuscating it with canEmit?

let's keep a ref like:
noPaginationHasOcurred = true

let's modify goToPage to treat the inital positioning differently
```
const goToPage = (page, animate, initialPositioning) =>  {
....
if (!initialPositioning){
  noPaginationHasOcurred = false;
}
...
}
```

Now the condition for when changing page can be more explicit
```
if (noPaginationHasOcurred) {
   return; // We don't emit events prior to first pagination
}

if (pageChangeControl.current.lastPageIndex !== currentPageIndex){
  ...we do stuff here...
}

```

wyunreal: This doesn't works in the case the initial item index is 0 or undefined, be cause the goToPage function is called only when initial item index is set, different than 0, or the user clicks on an arrow.. (it doesnt get called on user swipe).

wyunreal (commented):
Code refactored, WDYT now ?

Winde (approved):

atabel (approved):

src/carousel.tsx
```diff
@@ -347,9 +350,18 @@ const BaseCarousel: React.FC<BaseCarouselProps> = ({
                     shownItemIndexes.unshift(idx);
                 }
             }
-            onPageChange({pageIndex: currentPageIndex, shownItemIndexes});
+
+            if (!pageInitialized.current) {
+                pageInitialized.current =
+                    lastPageIndex.current !== currentPageIndex &&
+                    shownItemIndexes.includes(initialActiveItem || 0);
```
wyunreal, line 357: This condition changed due to a bug causing a test to fail: the carousel should be considered initialized if a page transition ends with the initial item being contained on the current page.

atabel (approved):

PR #709, opened 2023-04-05
fix(Button): change minWidth

geremiascorrea-tlf:
Modifying min button width size and fixes snapshots based on this modification, requested by Vivo.

Figma link: https://www.figma.com/file/koROdh3HpEPG2O8jG52Emh/%F0%9F%94%B8-Buttons-Component-Specs?node-id=0-1&t=eJRUzlQqlKF9xRBL-0

Important note: Changing this min-width value, as it is proposed by default behavior, changes several already committed snapshots. Initially we noticed the 4 around the button in a direct way, in which it was indicated to change. But opening the PR and retesting completely locally, another 70 snapshots showed errors, due to the button layout change. I would like you to know the most appropriate thing to do in this case. Would it change all the other snapshots and commit to this PR? Is there any indication for this case?


yceballost (approved):

geremiascorrea-tlf:
Just a notice of the last commit. To test the pipe, since locally it was taking timeout (due to the Carousel component, it seems to me), I committed with the corrected snapshots. The only flaw that remains is this one from Carousel, but since the test execution locally is taking timeout for the entire component, I couldn't replace the snapshot.

If you have any tips, thanks!

geremiascorrea-tlf:
> > Just a notice of the last commit. To test the pipe, since locally it was taking timeout (due to the Carousel component, it seems to me), I committed with the corrected snapshots. The only flaw that remains is this one from Carousel, but since the test execution locally is taking timeout for the entire component, I couldn't replace the snapshot.
> > If you have any tips, thanks!
> 
> I had troubles like this, I solved closing all programs that I had opened, Figma was eating the entire RAM, and this caused timeout in the snapshot process 😬
> 
> You can also try running the command only for this test
> 
> `yarn test-acceptance XXXXXXX-screenshot-test -u`
> 
> where "XXXXXXX" is the name of the screenshot test

This is the command I've been running. I've tried closing everything, believing it to be the RAM memory, but it keeps giving timeout, only in Carousel component. But I will continue with new attempts at it.

Supposedly, managing to generate the image of the new Carousel component that is returning an error, we will have all the snapshots updated according to the modification of the button component.

yceballost:
> > > Just a notice of the last commit. To test the pipe, since locally it was taking timeout (due to the Carousel component, it seems to me), I committed with the corrected snapshots. The only flaw that remains is this one from Carousel, but since the test execution locally is taking timeout for the entire component, I couldn't replace the snapshot.
> > > If you have any tips, thanks!
> > 
> > 
> > I had troubles like this, I solved closing all programs that I had opened, Figma was eating the entire RAM, and this caused timeout in the snapshot process 😬
> > You can also try running the command only for this test
> > `yarn test-acceptance XXXXXXX-screenshot-test -u`
> > where "XXXXXXX" is the name of the screenshot test
> 
> This is the command I've been running. I've tried closing everything, believing it to be the RAM memory, but it keeps giving timeout, only in Carousel component. But I will continue with new attempts at it.
> 
> Supposedly, managing to generate the image of the new Carousel component that is returning an error, we will have all the snapshots updated according to the modification of the button component.

I ran the command locally in your branch and this is the result
image

Screenshot testing is bugged and Abel is trying to fix it

atabel:
> 

yes, it should be fixed now. Please update this branch with master

atabel (approved):

PR #711, opened 2023-04-11
chore(storybook): add controls to snackbar story

marcoskolodny:
image


src/__stories__/snackbar-story.tsx
```diff
@@ -1,26 +1,49 @@
...
+Default.args = {
+    buttonText: 'Action',
+    message: 'Some message',
+    type: 'INFORMATIVE',
+    duration: 'Default',
+};
+
+Default.argTypes = {
+    duration: {
+        options: ['Default', '5000', '10000'],
```
atabel, line 42: I'd include more values, 15000 for example. Because 5 and 10s are already the default values (when action is provided or not)

atabel (approved):

brtbrt (approved):

PR #715, opened 2023-04-12
chore(Playroom): add Loader component

marcoskolodny:
Created [Loader component](https://jira.tid.es/browse/WEB-749) in playroom.

- behaviour when `loader` is a string (i.e. an url): [Video](https://user-images.githubusercontent.com/25785151/231468314-1c6a3272-35ca-426c-90cb-0c0052e211fc.webm)

- behaviour when `loader` is a function: [Video](https://user-images.githubusercontent.com/25785151/231468425-1419ae42-2abe-4efb-a2c3-456e7ff77b58.webm)

In case of errors inside `loader`, `render` or `renderLoading` functions, `renderError` is executed.

For the videos, I used a fake API that I created locally. It returns the message `"Hello World"` after waiting for two seconds.


playroom/components/loader.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+type Props = {
+    loader: string | (() => Promise<any>);
+    render: (data: any) => React.ReactElement;
+    renderLoading: () => React.ReactElement;
+    renderError: () => React.ReactElement;
+};
+
+export const Loader: React.FC<Props> = ({loader, render, renderLoading, renderError}) => {
+    const [content, setContent] = React.useState<React.ReactElement>(<></>);
```
atabel, line 11: it isn't a good idea to store react elements (jsx) inside a react state. I'd just store the api result and/or error in react state, and call `render` `renderLoading` or `renderError` at render time

playroom/components/loader.tsx
```diff
@@ -0,0 +1,46 @@
+import * as React from 'react';
+
+type Props = {
+    loader: string | (() => Promise<any>);
+    render: (data: any) => React.ReactElement;
+    renderLoading: () => React.ReactElement;
+    renderError: () => React.ReactElement;
+};
+
+export const Loader: React.FC<Props> = ({loader, render, renderLoading, renderError}) => {
```
atabel, line 10: for files that exports a single component, we preffer a default export.
```ts
const Loader = ...

export default Loader;
```

playroom/components/loader.tsx
```diff
@@ -0,0 +1,46 @@
+import * as React from 'react';
+
+type Props = {
+    loader: string | (() => Promise<any>);
+    render: (data: any) => React.ReactElement;
+    renderLoading: () => React.ReactElement;
+    renderError: () => React.ReactElement;
```
atabel, line 7: I'd made these 2 functions optional and render null if they aren't provided

playroom/components/loader.tsx
```diff
@@ -0,0 +1,46 @@
+import * as React from 'react';
+
+type Props = {
+    loader: string | (() => Promise<any>);
```
atabel, line 4: I'd call this prop `load`. `<Loader loader` sounds weird to me

playroom/components/loader.tsx
```diff
@@ -0,0 +1,64 @@
...
+type Props = {
+    load: string | (() => Promise<any>);
+    render: (data: any) => React.ReactElement;
+    renderLoading?: () => React.ReactElement;
+    renderError?: () => React.ReactElement;
+};
+
+const Loader: React.FC<Props> = ({load, render, renderLoading, renderError}) => {
+    const [loaderData, setLoaderData] = React.useState(null);
+    const [isLoading, setIsLoading] = React.useState(false);
```
brtbrt, line 12: wouldn't be better a `true` when it's first loaded? 🙃

playroom/components/loader.tsx
```diff
@@ -0,0 +1,64 @@
...
+    }, [load, render, renderLoading, renderError]);
+
+    let content;
+
+    if (isLoading) {
+        content = renderLoading ? renderLoading() : null;
+    } else if (hasError) {
+        content = renderError ? renderError() : null;
+    } else {
+        content = loaderData ? render(loaderData) : null;
```
atabel, line 58: I think I prefer early returns in this case:
```suggestion
    if (isLoading) {
        return renderLoading();
    } else if (hasError) {
        return renderError();
    } 
    return render(loaderData);
```

playroom/components/loader.tsx
```diff
@@ -0,0 +1,68 @@
...
+type LoaderState = 'loading' | 'error' | 'success';
+
+const Loader: React.FC<Props> = ({load, render, renderLoading = () => null, renderError = () => null}) => {
+    const [loaderData, setLoaderData] = React.useState(null);
+    const [loaderStatus, setLoaderStatus] = React.useState<LoaderState>('loading');
+
+    React.useEffect(() => {
+        let isRenderActive = true;
+
+        if (isRenderActive) setLoaderStatus('loading');
```
atabel, line 19: this is always true

playroom/components/loader.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+        if (typeof load === 'string') {
+            fetch(load)
```
atabel, line 22: to avoid repeating some logic, you can create a `load` function when a string is provided, this way you only have a single code path:
```tsx
const loadFn = typeof load === 'function' ?  load : () => {
    fetch(load).then((response) => {
        if (!response.ok) {
            throw new Error('fetch failed');
        }
        return response.json();
    });
}

// from here you have a single code path instead of two (one for string and one for function)
```

atabel (approved):

playroom/snippets.tsx
```diff
@@ -1615,6 +1615,55 @@ const heroSnippets = [
     },
 ];
 
+const loaderSnippets = [
+    {
+        name: 'Loader with url',
```
yceballost, line 1620: maybe include `Loader with url (internal use)` or something like that to avoid confusions?

PR #717, opened 2023-04-13
WEB-658 add link props to navbar logo

brtbrt:
to be discussed: do we need to add some demo purposes on playroom, or improve the storybook? I didn't think so, but please do tell


src/navigation-bar.tsx
```diff
@@ -200,15 +202,80 @@ type MainNavigationBarProps = {
...
+    trackingEvent?: TrackingEvent | ReadonlyArray<TrackingEvent>;
+    onPress: () => void;
+};
+
+type MainNavigationBarProps = ExclusifyUnion<
+    | MainNavigationBarPropsBase
+    | MainNavigationBarToProps
+    | MainNavigationHrefProps
+    | MainNavigationOnPressProps
+>;
```
brtbrt, line 228: are we ok with these props? check them carefully 🙏🙃

brtbrt:
it will be fixed by https://github.com/Telefonica/mistica-web/pull/721

PR #720, opened 2023-04-14
fix(Carousel): initial flash rendering as tablet while screenSize provider is not initialized

wyunreal:


src/carousel.tsx
```diff
@@ -198,8 +198,8 @@ const BaseCarousel: React.FC<BaseCarouselProps> = ({
     const desktopContainerType = useDesktopContainerType();
     const itemsPerPageConfig = normalizeItemsPerPage(desktopContainerType || 'large', itemsPerPage);
 
-    const {isDesktopOrBigger, isMobile} = useScreenSize();
-    const mobileOrTabletItemsPerPage = isMobile ? itemsPerPageConfig.mobile : itemsPerPageConfig.tablet;
+    const {isDesktopOrBigger, isTablet} = useScreenSize();
+    const mobileOrTabletItemsPerPage = isTablet ? itemsPerPageConfig.tablet : itemsPerPageConfig.mobile;
```
wyunreal, line 202: This is required to improve a bad behaviour happened in renders when the provider is not initialized, that was causing the carousel behave like in tablet, its better to behave like mobile in those cases (as the usage of tablets is lot less than mobile).

src/carousel.tsx
```diff
@@ -352,9 +352,7 @@ const BaseCarousel: React.FC<BaseCarouselProps> = ({
             }
 
             if (!pageInitialized.current) {
-                pageInitialized.current =
-                    lastPageIndex.current !== currentPageIndex &&
```
wyunreal, line 356: Also, this line is not required anymore.

atabel (approved):

PR #722, opened 2023-04-19
feat(ThemeVariant): support alternative theme variant

atabel:
jira: https://jira.tid.es/browse/WEB-818


package.json
```diff
@@ -81,7 +81,7 @@
         "@swc/core": "^1.3.11",
         "@swc/jest": "^0.2.23",
         "@telefonica/acceptance-testing": "2.13.0",
-        "@telefonica/eslint-config": "^1.0.7",
+        "@telefonica/eslint-config": "^1.3.0",
```
atabel, line 84: unrelated with this task, but I've updated the eslint-config and fixed some type imports

src/theme-variant-context.tsx
```diff
@@ -1,14 +1,20 @@
 import * as React from 'react';
 
-const IsInverseThemeVariant = React.createContext<boolean>(false);
+export type Variant = 'inverse' | 'default' | 'alternative';
```
brtbrt, line 3: my "OCD" says: move the `default` as the first type 😂

src/tag.tsx
```diff
@@ -33,7 +34,11 @@ const Tag: React.FC<TagProps> = ({Icon, children, dataAttributes, type = 'promo'
     const tagTypeToColors = {
         promo: [colors.promoHigh, colors.promoHighInverse, colors.promoLow],
```
brtbrt, line 35: side-side note: I think there's room for more semantics some way: let me explain this better.
I've had to expand the sourcecode in order to understand what position was for (the `[textColor, inverseTextColor, backgroundColor]` below). Although it's not such a big problem, I'm wondering if there's a better way to make this variable more semantic (i.e. an object with properties) where it's defined, and not where it's used 🙃

src/skins/telefonica.tsx
```diff
@@ -182,13 +185,15 @@ export const getTelefonicaSkin: GetKnownSkin = () => {
             appBarBackground: palette.darkModeGrey,
             background: palette.darkModeBlack,
             backgroundContainer: palette.darkModeGrey,
+            backgroundContainerBrand: palette.darkModeGrey,
```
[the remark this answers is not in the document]
brtbrt: errata corrige: it was the dark mode 😅 my fault

src/chip.tsx
```diff
@@ -76,9 +80,12 @@ const Chip: React.FC<ChipProps> = ({Icon, children, id, dataAttributes, active,
...
             <Box
-                className={classnames(styles.chipVariants[active ? 'active' : 'default'], {
-                    [styles.chipInteractiveVariants[isDarkMode ? 'dark' : 'light']]: isInteractive,
-                })}
+                className={classnames(
+                    styles.chipVariants[active ? 'active' : overAlternative ? 'overAlternative' : 'default'],
+                    {
+                        [styles.chipInteractiveVariants[isDarkMode ? 'dark' : 'light']]: isInteractive,
+                    }
+                )}
```
brtbrt, line 88: just for my own info: why does not the chip have the inverse?

src/boxed.tsx
```diff
@@ -71,7 +71,9 @@ export const InternalBoxed = React.forwardRef<HTMLDivElement, Props & InternalPr
                         overflow: 'hidden',
                         background: !background
                             ? isInverseInside && !isDarkMode
-                                ? vars.colors.backgroundBrand
+                                ? isInverseOutside
+                                    ? vars.colors.brandHigh
+                                    : vars.colors.backgroundContainerBrand
```
brtbrt, line 76: did the lint allow you this? :D

brtbrt (approved):

src/tag.tsx
atabel: yes, we could have that, but the resulting code would be quite bigger because we have to repeat the same object keys for every tag.
Maybe a comment above would be enough, wdyt?

src/chip.tsx
atabel: the chips have the same styles over inverse/default

src/boxed.tsx
brtbrt: oh ❤️ LET THE FUN BEGIN

aweell (approved):

marcoskolodny (approved):

PR #726, opened 2023-04-24
feat(Button): show loading state when onPress returns a promise until it resolves

marcoskolodny:
Issue: https://jira.tid.es/browse/WEB-824


src/__stories__/button-story.tsx
```diff
@@ -257,6 +257,11 @@ export const LoadingState: StoryComponent = () => {
             setIsLoading(false);
         }, 2000);
     };
+
+    const handlePressWithPromise = async () => {
+        await new Promise((resolve) => setTimeout(resolve, 3000));
+    };
```
atabel, line 263: You don't need the `async`/`await`
```suggestion
    const handlePressWithPromise = () =>
        new Promise((resolve) => setTimeout(resolve, 3000));
```

src/__stories__/button-story.tsx
```diff
@@ -258,9 +258,8 @@ export const LoadingState: StoryComponent = () => {
         }, 2000);
     };
 
-    const handlePressWithPromise = async () => {
-        await new Promise((resolve) => setTimeout(resolve, 3000));
-    };
+    const handlePressWithPromise: () => Promise<void> = () =>
+        new Promise((resolve) => setTimeout(resolve, 3000));
```
atabel, line 262:
```suggestion
    const handlePressWithPromise = (): Promise<void> =>
        new Promise((resolve) => setTimeout(resolve, 3000));
```

src/button.tsx
```diff
@@ -291,7 +292,18 @@ const Button = React.forwardRef<TouchableElement, ButtonProps & {type: ButtonTyp
     }
 
     if (props.onPress) {
-        return <BaseTouchable {...commonProps} onPress={props.onPress} />;
+        return (
+            <BaseTouchable
+                {...commonProps}
+                onPress={(e) => {
+                    const result = props.onPress(e);
+                    if (result) {
```
kydorn, line 300: if (result?.then)

You don't know at this point if the onPress is returning undefined, null, a value or a promise. To check if it is a promise you can check if it has the 'then' property. I don't know if there is a better way tbh

src/button.tsx
```diff
@@ -291,7 +292,18 @@ const Button = React.forwardRef<TouchableElement, ButtonProps & {type: ButtonTyp
     }
 
     if (props.onPress) {
-        return <BaseTouchable {...commonProps} onPress={props.onPress} />;
+        return (
+            <BaseTouchable
+                {...commonProps}
+                onPress={(e) => {
```
kydorn, line 298: should this be a useCallback?

src/button.tsx
```diff
@@ -291,7 +292,18 @@ const Button = React.forwardRef<TouchableElement, ButtonProps & {type: ButtonTyp
...
+                    const result = props.onPress(e);
+                    if (result) {
+                        setIsOnPressPromiseResolving(true);
+                        result.then(() => setIsOnPressPromiseResolving(false));
```
kydorn, line 302: if the promise fails you need to also remove the spinner. why not use .finally?

src/button.tsx
kydorn: discussed offline, If result has a value, it will always be a promise, so this is not a promise.

src/button.tsx
```diff
@@ -151,8 +151,22 @@ const Button = React.forwardRef<TouchableElement, ButtonProps & {type: ButtonTyp
...
+        (e: React.MouseEvent<HTMLElement, MouseEvent>) => {
+            if (props.onPress) {
+                const result = props.onPress(e);
+                if (result) {
+                    setIsOnPressPromiseResolving(true);
+                    result.finally(() => setIsOnPressPromiseResolving(false));
+                }
+            }
+        },
+        [props]
```
atabel, line 166: I think this would be enough:
```suggestion
        [props.onPress]
```

atabel (approved):

PR #728, opened 2023-04-26
fix(TextField): label not shrinking when the field was autofilled by browser autocomplete

brtbrt:
fix: when autofill was performed, the label didn't use to collapse.


src/__screenshot_tests__/form-fields-screenshot-test.tsx
```diff
@@ -27,6 +27,27 @@ test.each(TESTABLE_DEVICES)('Default textfield appears properly on %s', async (d
...
+        id: 'components-input-fields--variants',
+        device: 'DESKTOP',
+    });
+
+    const fieldWrapper = await screen.findByTestId('normal-field');
+    const image = await fieldWrapper.screenshot();
+
+    expect(image).toMatchImageSnapshot();
+
+    await page.evaluate(() => {
```
brtbrt, line 41: since autofill can't be performed on puppeteer, I simulate it this way: in case the fix isn't applied, this simulation makes the test fail

src/__screenshot_tests__/form-fields-screenshot-test.tsx
```diff
@@ -27,6 +27,27 @@ test.each(TESTABLE_DEVICES)('Default textfield appears properly on %s', async (d
     expect(image).toMatchImageSnapshot();
 });
 
+test('Textfield collapses its label when autofill fills out the form', async () => {
```
brtbrt, line 30: Even though the fix is applied on the `TextFieldBase` I'd test just a component for this corner case. WDYT?

src/text-field-base.tsx
```diff
@@ -300,6 +300,11 @@ export const TextFieldBase = React.forwardRef<any, TextFieldBaseProps>(
                                 // Workaround for systems where maxlength prop is applied onBlur (https://caniuse.com/#feat=maxlength)
                                 if (maxLength === undefined || event.target.value.length <= maxLength) {
                                     setCharacterCount(event.target.value.length);
+
+                                    if (event.target.value.length > 0 && inputState !== 'focused') {
```
atabel, line 304: please add a comment here explaining why this is needed

src/__screenshot_tests__/form-fields-screenshot-test.tsx
atabel: please, include a comment here explaining we are simulating an autofill

src/__screenshot_tests__/form-fields-screenshot-test.tsx
```diff
@@ -27,6 +27,27 @@ test.each(TESTABLE_DEVICES)('Default textfield appears properly on %s', async (d
...
+        device: 'DESKTOP',
+    });
+
+    const fieldWrapper = await screen.findByTestId('normal-field');
+    const image = await fieldWrapper.screenshot();
+
+    expect(image).toMatchImageSnapshot();
+
+    await page.evaluate(() => {
+        const input = document.querySelector('input#aria-id-hook-1');
```
atabel, line 42: I'd search the element by data-testid instead

atabel (approved):

marcoskolodny (approved):

PR #731, opened 2023-04-26
chore(serve-static-storybook): fix video load

atabel:


scripts/serve-static-storybook.js
```diff
@@ -1,13 +1,16 @@
-const StaticServer = require('static-server');
+const handler = require('serve-handler');
```
atabel, line 1: replaced `static-server` package by `serve-handler`. `static-serve` was not serving videos well, and `serve-handler` seems to work fine

marcoskolodny (approved):

PR #733, opened 2023-04-27
chore(tokens): workflow to import design tokens from mistica-design repo

atabel:
https://confluence.tid.es/display/CTO/PID%3A+Mistica+Token+Generator


packages/generate-design-tokens/index.js
```diff
@@ -0,0 +1,107 @@
...
+    KNOWN_SKINS.forEach((skinName) => {
+        console.log('Generating tokens for skin', skinName);
+
+        if (!fs.existsSync(`${skinName}.json`)) {
+            console.error(`Missing ${skinName}.json file`);
+            return;
+        }
+
+        const skinSrc = generateSkinSrc(skinName);
+        fs.writeFileSync(path.join(__dirname, '..', '..', 'src', 'skins', `${skinName}.tsx`), skinSrc);
```
brtbrt, line 103: I'd define the folder like `DESIGN_TOKENS_FOLDER` for a clearer semantics :)
Actually, thinking about future evolution, I'm not that sure that hardcoding source and destination folders are the best, but it's a minor concern in this case!  Just thinking aloud

brtbrt (approved):

brtbrt (approved):

packages/generate-design-tokens/index.js
```diff
@@ -0,0 +1,108 @@
...
+    if (colorWithAlphaMatches) {
+        const colorName = colorWithAlphaMatches[1];
+        const alpha = colorWithAlphaMatches[2];
+        return `applyAlpha(${colorName}, ${alpha})`;
+    }
+
+    throw new Error(`Unknown color format: ${colorDescription.value}`);
+};
+
+const buildRadius = (radiusDescription) => {
```
marcoskolodny, line 25: Is it safe to assume we will never get something like `1em`?

marcoskolodny (approved):

packages/generate-design-tokens/index.js
atabel: yes, it is. If we need this in the future, we'll need to review this too

PR #557, opened 2022-10-14
WEB-662 fixed footer layout as fixed only when enough height available

brtbrt:
The change has been currently applied only on the `FixedFooterLayout`. If we're all ok to this, I'll apply it to the `ButtonFixedLayout` 😉.

- [x] FixedFooterLayout
- [x] ButtonFixedLayout

Behavior: in case we're on mobile/tablet, and the viewport's height is smaller than the footer itself, the footer switches from fixed to the normal flow. And vice versa.

https://user-images.githubusercontent.com/61736059/195844815-30e77c82-f66a-459b-bc91-d77e0c38229a.mp4

PR #735, opened 2023-04-28
Mbertamini/web 704 iframe footer fixed

brtbrt:


src/fixed-footer-layout.tsx
```diff
@@ -54,10 +55,13 @@ const FixedFooterLayout: React.FC<Props> = ({
...
     const {height: realFooterHeight, ref} = useElementDimensions();
-    const windowHeight = useWindowHeight();
+    const isWithinIFrame = useIsWithinIFrame();
+    const windowHeight = useWindowHeight(true);
     const screenHeight = useScreenHeight();
 
-    const hasContentEnoughVSpace = windowHeight - realFooterHeight > screenHeight / FOOTER_CANVAS_RATIO;
+    const hasContentEnoughVSpace =
+        windowHeight - realFooterHeight >
+        (isWithinIFrame ? windowHeight : screenHeight) / FOOTER_CANVAS_RATIO;
```
atabel, line 64: why don't we always use window height?

src/hooks.tsx
```diff
@@ -116,40 +116,49 @@ export const useAriaId = (id?: string): string => {
     }
 };
 
-export const useWindowSize = (): {
+const useSpecificWindowObject = (useInnerFrame: boolean): Window | undefined =>
```
atabel, line 119: I'd avoid calling `useXXX` to functions that are not hooks. This function isn't a hook because it doesn't use any other hook (useState, useEffect, etc)

src/hooks.tsx
```diff
@@ -116,40 +116,49 @@ export const useAriaId = (id?: string): string => {
...
+    typeof window !== 'undefined' ? (useInnerFrame ? window?.self : window) : undefined;
```
atabel, line 120: I don't understand this logic, isn't `window === window.self` ?

src/fixed-footer-layout.tsx
brtbrt: it came from the way we decided to tackle the fixed footer layout issued we used to have with the keyboard open: 
Screenshot 2023-05-06 at 16 43 35

and the PR:
https://github.com/Telefonica/mistica-web/pull/557

if I remember well, screen is more a more stable way of calculating what we need

src/hooks.tsx
brtbrt: 🤔I see, I'd better leave it within the fixed footer layout itself, and call it `getSpecificWindowObject`. WDYT?

src/hooks.tsx
brtbrt: of course you don't understand, it was absolutely wrong and unnecessary! My fault, thanks for pointing out.

src/hooks.tsx
```diff
@@ -137,6 +137,9 @@ export const useWindowSize = (): {
     );
 
     React.useEffect(() => {
+        if (typeof window === 'undefined') {
+            return;
+        }
```
atabel, line 142: in which case window will be undefined? If you're thinking in SSR, useEffects don't run in SSR

src/fixed-footer-layout.tsx
atabel: If I remember right, the virtual keyboard + fixed footer  was an issue in Android, not in iOS.
image
According to this message. If the `availHeight` doesn't change when we open/close the virtual keyboard but the `innerHeight` does change, shouldn't we using the `innerHeight` (`useWindowHeight` in this case)?

src/hooks.tsx
brtbrt: oh, excellent. thanks

src/fixed-footer-layout.tsx
brtbrt: in case we're into an iframe, the `innerHeight` does not change :): it will be the iframe's height itself
Screenshot_1683555051
Screenshot_1683555065

I see it convenient using it for this edge case

atabel (approved):

marcoskolodny (approved):

PR #743, opened 2023-05-04
feat(StackingGroup): new component

BrunoGuimaraes103:
New component, Stacking Group along with changing the Image component to include a border as specified in the story:
https://jira.telefonica.com.br/browse/PTIECBC03-3987

Figma component specs:
https://www.figma.com/file/3ME86VKaAQ2BoeqoQuHMYx/%F0%9F%94%B8-%5BREADY%5D-StackingGroup-Specs


BrunoGuimaraes103:
One of the errors in the test, build-acceptance(2), occurs due to the difference in colors between the current screenshot of the error case of the image component with the inverse context where it should not have any transparency, but in the code this part is untouched with transparency

The reason I didn't touch this part of the code is because the resolution of this color token has already been resolved in this case:
https://github.com/Telefonica/mistica-design/pull/1226

src/stacking-group.tsx
```diff
@@ -0,0 +1,155 @@
...
+    const isInverse = useIsInverseVariant();
+    const backgroundColor = isInverse ? vars.colors.brandHigh : vars.colors.brandLow;
+    const textColor = isInverse ? vars.colors.textPrimaryInverse : vars.colors.brand;
+
+    let maxItems = 0;
+    if (stacked) {
+        maxItems = 4;
+    } else {
+        maxItems = 3;
+    }
```
atabel, line 34: this shouldn't be harcoded here, it should be a `maxItems` prop and be `Infinity` by default

src/stacking-group.tsx
```diff
@@ -0,0 +1,155 @@
+import * as React from 'react';
+import Avatar from './avatar';
+import Image from './image';
+import * as styles from './stacking-group.css';
+import {useIsInverseVariant} from './theme-variant-context';
+import {vars} from './skins/skin-contract.css';
+import classNames from 'classnames';
+
+type dataArrayType = {
```
atabel, line 9: Types always start with uppercase
Don't include the word Array or Type in a type

You can call it just `type Data =`. But in this case I think it doesn't make sense to create a new type, I'd just inline it in the Props definition

src/stacking-group.tsx
```diff
@@ -0,0 +1,155 @@
...
+    src?: string;
+    initials?: string;
+};
+
+type StackingGroupProps = {
```
atabel, line 14: If there is only one component in the file, just call this type Props:
```suggestion
type Props = {
```

src/stacking-group.tsx
```diff
@@ -0,0 +1,155 @@
...
+    size: 24 | 32 | 64;
+    data: dataArrayType[];
```
atabel, line 16: use generic array syntax `Array<>`

src/stacking-group.tsx
```diff
@@ -0,0 +1,155 @@
...
+    let maxItems = 0;
+    if (stacked) {
+        maxItems = 4;
+    } else {
+        maxItems = 3;
+    }
+
+    const plusCase = data.length - maxItems;
+
+    const verify = type === 'avatar';
```
atabel, line 38: verify what?

BrunoGuimaraes103:
@atabel, I have now linked figma in the first comment below the story link

src/stacking-group.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+    return (
+        <div className={styles.inline}>
```
atabel, line 41: why don't use the Inline component?

playroom/snippets.tsx
```diff
@@ -1727,6 +1727,46 @@ const logoSnippets = [
     },
 ];
 
+const stackingGroupSnippets = [
+    {
+        name: 'Stacking Group',
+        code: `
+        <StackingGroup
```
atabel, line 1734: I don't like this api, this component should be used like this:
```ts
<StackingGroup stacked={false} maxItems={5} moreItemsStyle={{type: 'circle', size: 64}}>
    <Avatar size={64} src="https://source.unsplash.com/600x600/?face" />
    <Avatar size={64} src="https://source.unsplash.com/600x600/?face" />
    <Avatar size={64} src="https://source.unsplash.com/600x600/?face" />
    <Avatar size={64} src="https://source.unsplash.com/600x600/?face" />
    <Avatar size={64} src="https://source.unsplash.com/600x600/?face" />
    <Avatar size={64} src="https://source.unsplash.com/600x600/?face" />
</StackingGroup>
```

```ts
type Props = {
    stacked?: boolean;
    maxItems?: number;
    moreItemsStyle: {
        type: 'circle' | 'square';
        size: 64 | 32 | 24;
    };
    children: React.ReactNode;
}
```

src/__screenshot_tests__/stacking-group-screenshot-test.tsx
```diff
@@ -0,0 +1,24 @@
...
+test.each`
+    size    | type        | stacked  | inverse
+    ${'64'} | ${'image'}  | ${false} | ${false}
+    ${'64'} | ${'avatar'} | ${false} | ${false}
+    ${'64'} | ${'image'}  | ${true}  | ${false}
+    ${'64'} | ${'avatar'} | ${true}  | ${false}
+    ${'64'} | ${'image'}  | ${false} | ${true}
+    ${'64'} | ${'avatar'} | ${false} | ${true}
+    ${'64'} | ${'image'}  | ${true}  | ${true}
+    ${'64'} | ${'avatar'} | ${true}  | ${true}
```
atabel, line 12: why is 64 a param in the test? it's always 64

src/__stories__/stacking-group-story.tsx
```diff
@@ -0,0 +1,81 @@
...
+    argTypes: {
+        size: {
+            options: [24, 32, 64],
+            control: {type: 'select'},
+        },
+        type: {
+            options: ['avatar', 'image'],
+            control: {type: 'select'},
+        },
+    },
```
atabel, line 17: move this to `Default.argTypes`

src/__stories__/stacking-group-story.tsx
```diff
@@ -0,0 +1,81 @@
...
+                <StackingGroup size={size} type={type} stacked={stacked} data={data} />
+            </div>
+        </ThemeVariant>
+    );
+};
+
+Default.storyName = 'Stacking Group';
+Default.args = {
+    size: 24,
+    data: [
```
atabel, line 52: Instead of harcoding here an array, I'd include a slider with the number of items to include.
Also include an arg for `maxItems` prop, please

atabel (changes requested):

src/__screenshot_tests__/__image_snapshots__/stacking-group-screenshot-test-tsx-stacking-group-5-snap.png
yceballost, line 1: why in this case placeholder image hasn't a border? (in the other screenshots placeholder in image has border)

src/stacking-group.tsx
yceballost, line 1: We need to include the possibility of setup the maxItem

For example
image

or 

image

yceballost (changes requested):

src/stacking-group.tsx
BrunoGuimaraes103: According to history, there is a maximum limit of 5 if the stacked is true and 4 if it is false
Captura de tela 2023-05-08 132759

yceballost: This maximun is just for the Vivo Novo app, but not in terms of mística component. Mística component provide more flexibility because have to be built as a global component. Later, when a product (in this case novo app) want to use this component, can provide its logic for a specific use.

playroom/snippets.tsx
yceballost: The size of moreItemsStyle could be open because you can use any size in the children component (always following the multiplies of 8)

src/stacking-group.tsx
BrunoGuimaraes103: The Verify const was to verify which element would be mounted in the Stacking Group being Avatar or Image but now it has been renamed to styleCircle where its function is to determine if the element that shows the excess number of elements will be a circle or a square.

src/__screenshot_tests__/stacking-group-screenshot-test.tsx
BrunoGuimaraes103: In the Screenshot tests, the default was 64 to avoid repeating the prints for each of the screen sizes (24, 32 and 64) with variations but due to the switch to accept any component now the only size change available is that of the element that shows the surplus number, which in the story was configured so that the same value is used in the image component.

src/__screenshot_tests__/__image_snapshots__/stacking-group-screenshot-test-tsx-stacking-group-3-snap.png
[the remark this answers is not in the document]
BrunoGuimaraes103: The reason for the misalignment in the screenshot was due to an error in the image not having the src (pos was the same Array used to assemble the Avatars) and that is why this misalignment ended up occurring.

src/__screenshot_tests__/__image_snapshots__/stacking-group-screenshot-test-tsx-stacking-group-5-snap.png
BrunoGuimaraes103: The Placeholder of the Stacking Group screenshots, had no border due to being stacked false, and in this case it was defined that it would not have the border, but now with the elements being passed to the Stacking Group, you would have to put these props in place will or will not have the border on the component itself when moving to Stacking.

BrunoGuimaraes103:
According to the requests given, the code was modified to adapt where the logic was changed to instead of taking data from an Array and assembling the Avatar and Image components inside the Stacking Group, now you can pass any element in the Stacking Group, I did before had a misconception about it being just these two components and not any one, so the logic was switched.

Also, as the component is passed into the stacking group, the error case border was removed from the image, leaving only the option to use the border for the normal case.

src/__stories__/stacking-group-story.tsx
```diff
@@ -0,0 +1,75 @@
...
+        control: {type: 'range', min: 1, max: 8, step: 1},
+    },
+};
+
+Default.args = {
+    itemsToInclude: 5,
+    type: 'circle',
+    size: 64,
+    maxItems: 4,
+    stacked: false,
```
yceballost, line 73: should be `true` by default? @atabel

src/stacking-group.tsx
yceballost, line 1: recently we include skin tokens for border radius, can you use `container` for the square maxItem element? I updated specs to include this information in the Token artboard.

https://www.figma.com/file/3ME86VKaAQ2BoeqoQuHMYx/%F0%9F%94%B8-%5BREADY%5D-StackingGroup-Specs?type=design&node-id=2060-532&t=C1M07ggxQX7tNO1B-4

yceballost (changes requested):

src/stacking-group.tsx
yceballost, line 1: I made this example to test the component, it is possible that I made some mistakes in my code (remember I am designer 😬)

https://tinyurl.com/2pzyw5q7

But I found some possible problems (assuming that my code is fine):
- Mixing avatars with image and non-image cause weird sizes
	- image
- Placeholder image doesn't have border, include it
	- image
- MaxItem element don't match with the children height, try it moving the size slider
- Don't use border when stacked is false
	- image

yceballost (changes requested):

src/stacking-group.tsx
BrunoGuimaraes103: Good afternoon @yceballost, I will respond according to the cases raised:

- It's happens only in the playroom itself, I tested this error in the storybook and the result was different as in the print below:
screenshot case 1

It turns out that in the playroom there are some differences from the storybook, for example, in it you can not pass obrigatory props, so I believe that this may be one of those cases of peculiarity of the playroom.

- I just included so that the border in the image is also in case of an error in the image, I still haven't uploaded the code post I wanted your answer to the last question in this comment.

- I just made the adjustment.

- It turns out that it was as specified in the history of not having a border when stacked false, but I can make the border optional if you want
screenshot history border

yceballost: Hi Bruno! thanks! 

I fix my code https://tinyurl.com/2oxwvku9

I can see little details to fix
1. the way of the images placeholder place is wrong. You can see hoy the 3 item is under the second one
image
2. In storybook, the last item is smaller than the rest (in my example, the last item is bigger than the rest. Anyway, in both cases the size is wrong)
image
image

BrunoGuimaraes103: Hi @yceballost, I solved these cases raised the reason for the misalignment was because the size of the border was messing with the size of the div of the maxItem, to avoid this I added the boxSizing: 'border-box'.

In case the error image falls behind, I added zIndex 1 to the div that surrounds the error image thus correcting this problem.

src/__stories__/stacking-group-story.tsx
yceballost, line 1: Thanks for the fixs, Bruno!

I think the component works visually. To finish my review, can you adapt the storybook example with the logic of this example? https://tinyurl.com/2oxwvku9

For example: 
- If stacked is false, avatar or image don't apply border inside
- Use avatar when type prop is "circle" and use image when type prop is square. To avoid bad usage in storybook examples about using square with avatar for example.

yceballost (approved):

src/__stories__/stacking-group-story.tsx
BrunoGuimaraes103: Done, I also took the opportunity to make the maxItem slider so as not to put negative numbers in it

src/stacking-group.tsx
```diff
@@ -0,0 +1,68 @@
...
+                                position: 'absolute',
+                                width: size,
+                                height: size,
+                                justifyContent: 'center',
+                                alignItems: 'center',
+                                color: textColor,
+                            }}
+                            aria-label={'+' + plusCase.toString()}
+                        >
+                            {'+' + plusCase}
```
yceballost, line 59: sorry for opening a review again, I didn't see this detail. Can you use `<Text4>` for the text inside more item element?

This information wasn't in the specs, but all the text should use a text preset. I add it to specs
image

BrunoGuimaraes103: No problem, I already put the change

src/stacking-group.tsx
```diff
@@ -0,0 +1,69 @@
...
+    children: React.ReactNode;
+};
+
+const StackingGroup: React.FC<Props> = ({moreItemsStyle, stacked = false, maxItems = Infinity, children}) => {
+    const isInverse = useIsInverseVariant();
+    const backgroundColor = isInverse ? vars.colors.brandHigh : vars.colors.brandLow;
+    const textColor = isInverse ? vars.colors.textPrimaryInverse : vars.colors.brand;
+    const border = stacked ? `1px solid ${vars.colors.borderLow}` : 'none';
+
+    const styleCircle = moreItemsStyle.type === 'circle';
```
atabel, line 23: inline this condition where you are using it, or if you prefer to declare a variable, do it near the line where you are using it

src/stacking-group.tsx
```diff
@@ -0,0 +1,69 @@
...
+    const space = stacked ? -8 : 8;
+    const countChildren = React.Children.count(children);
+    const plusCase = countChildren - maxItems;
+
+    const styleBorderRadius = styleCircle ? vars.borderRadii.avatar : vars.borderRadii.container;
```
atabel, line 28: just `borderRadius`, you can even inline it in the div `style`

src/stacking-group.tsx
```diff
@@ -24,7 +24,7 @@ const StackingGroup: React.FC<Props> = ({moreItemsStyle, stacked = false, maxIte
     const countChildren = React.Children.count(children);
     const plusCase = countChildren - maxItems;
 
-    const styleBorderRadius = styleCircle ? 64 : vars.borderRadii.container;
+    const styleBorderRadius = styleCircle ? vars.borderRadii.avatar : vars.borderRadii.container;
```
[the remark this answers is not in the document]
atabel: yes pelase

src/stacking-group.tsx
```diff
@@ -0,0 +1,69 @@
...
+                            display: 'flex',
+                            position: 'absolute',
+                            width: size,
+                            height: size,
+                            justifyContent: 'center',
+                            alignItems: 'center',
+                            color: textColor,
+                        }}
+                        aria-label={'+' + plusCase.toString()}
+                    >
```
atabel, line 58: You don't need the inner div. The aria-label is not needed, screen readers will read the text inside Text4
Please use classNames when posible, extract non-dynamic styles to a class in `stacking-group.css.ts`

```ts
<div
    className={styles.moreItems} // place non-dynamic styles in stacking-group.css.ts
    style={{width: size, height: size, borderRadius, border}
>
```

src/dialog.tsx
```diff
@@ -390,7 +390,7 @@ export default class DialogRoot extends React.Component<DialogRootProps, DialogR
     componentDidMount(): void {
         dialogRootInstances++;
         if (dialogRootInstances === 1) {
-            // eslint-disable-next-line @typescript-eslint/no-this-alias
+
```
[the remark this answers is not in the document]
BrunoGuimaraes103: This part of the code came from a master update

src/image.tsx
```diff
@@ -206,7 +210,17 @@ export const ImageContent = React.forwardRef<HTMLImageElement, ImageProps>(
...
                 {isError && withErrorFallback && (
-                    <div style={{position: 'absolute', width: '100%', height: '100%'}}>
+                    <div
+                        style={{
+                            position: 'absolute',
+                            width: '100%',
+                            height: '100%',
+                            boxSizing: 'border-box',
+                            border,
+                            borderRadius: vars.borderRadii.container,
```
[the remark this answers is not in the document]
BrunoGuimaraes103: I made the change to include noBorderSetting in this new version of the code

src/stacking-group.tsx
```diff
@@ -0,0 +1,69 @@
...
+    const space = stacked ? -8 : 8;
+    const countChildren = React.Children.count(children);
+    const plusCase = countChildren - maxItems;
+
+    const styleBorderRadius = styleCircle ? vars.borderRadii.avatar : vars.borderRadii.container;
+    const size = moreItemsStyle.size;
+
+    return (
+        <Inline space={space}>
+            {React.Children.map(children, (child, index: number) => index < maxItems && <>{child}</>)}
```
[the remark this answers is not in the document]
BrunoGuimaraes103: The quantity of items was requested to be infinity by default where the limit could be defined by the programmer, in order to be able to support more items.

src/dialog.tsx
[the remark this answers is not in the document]
atabel: then you did a bad merge, leave the code as it is in master:
https://github.com/Telefonica/mistica-web/blob/master/src/dialog.tsx#L393-L394

src/stacking-group.tsx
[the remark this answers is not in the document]
atabel: yes, but what I mean is that if maxItems is 5, you should show 4 items  + the "more items" item, not 5 + 1. Please see the examples in the specs

src/stacking-group.tsx
```diff
@@ -14,52 +15,30 @@ type Props = {
...
-                        position: 'relative',
                         width: size,
                         height: size,
-                        backgroundColor,
-                        borderRadius: styleBorderRadius,
-                        boxSizing: 'border-box',
-                        border,
-                        zIndex: 1,
+                        backgroundColor: isInverse ? vars.colors.brandHigh : vars.colors.brandLow,
+                        borderRadius: moreItemsStyle.type === 'circle' ? 64 : vars.borderRadii.container,
```
atabel, line 35:
```suggestion
                        borderRadius: moreItemsStyle.type === 'circle' ? '50%' : vars.borderRadii.container,
```

src/stacking-group.tsx
```diff
@@ -14,52 +15,30 @@ type Props = {
...
-    const textColor = isInverse ? vars.colors.textPrimaryInverse : vars.colors.brand;
-    const border = stacked ? `1px solid ${vars.colors.borderLow}` : 'none';
-
-    const styleCircle = moreItemsStyle.type === 'circle';
-    const space = stacked ? -8 : 8;
     const countChildren = React.Children.count(children);
-    const plusCase = countChildren - maxItems;
-
-    const styleBorderRadius = styleCircle ? vars.borderRadii.avatar : vars.borderRadii.container;
+    const moreItemsCount = countChildren - maxItems;
```
atabel, line 21:
```suggestion
    const moreItemsCount = countChildren - maxItems + 1;
```
image

playroom/snippets.tsx
```diff
@@ -1763,6 +1763,23 @@ const logoSnippets = [
     },
 ];
 
+const stackingGroupSnippets = [
+    {
+        name: 'Stacking Group',
+        code: `
+        <StackingGroup stacked={false} maxItems={5} moreItemsStyle={{type: 'circle', size: 64}}>
```
atabel, line 1770: I'd remove the `stacked={false}` in this snippet

src/__stories__/stacking-group-story.tsx
```diff
@@ -0,0 +1,93 @@
...
+                    background: inverse ? skinVars.colors.backgroundBrand : skinVars.colors.background,
+                    lineHeight: 0,
+                }}
+                data-testid="stacking-group"
+            >
+                <StackingGroup stacked={stacked} maxItems={maxItems} moreItemsStyle={{type, size}}>
+                    {Array.from({length: itemsToInclude}, (_, idx) =>
+                        type === 'circle' ? (
+                            <Avatar
+                                border={stacked}
```
atabel, line 49: you have a React warning here. Please add a key to Avatar and Image elements

src/__stories__/stacking-group-story.tsx
```diff
@@ -0,0 +1,93 @@
...
+        <ThemeVariant isInverse={inverse}>
+            <div
+                style={{
+                    padding: 16,
+                    width: 'fit-content',
+                    background: inverse ? skinVars.colors.backgroundBrand : skinVars.colors.background,
+                    lineHeight: 0,
+                }}
+                data-testid="stacking-group"
+            >
```
atabel, line 44: instead of applying the background color and the padding manually, you could use a `<ResponsiveLayout isInverse>` + `<Box>` here:
```js
<ResponsiveLayout isInverse={inverse}>
    <Box paddingY={24}>
```

src/stacking-group.tsx
```diff
@@ -0,0 +1,48 @@
...
+    const isInverse = useIsInverseVariant();
+    const countChildren = React.Children.count(children);
+    const moreItemsCount = countChildren - maxItems + 1;
+    const space = stacked ? -8 : 8;
+    const size = moreItemsStyle.size;
+
+    return (
+        <Inline space={space}>
+            {React.Children.toArray(children).slice(0, maxItems - 1)}
+            {countChildren >= maxItems && (
```
atabel, line 28: I think this logic is still wrong.
If there are 4 children and maxItems is 4, I'd expect the 4 children to be shown, no 3 + 1:
image
The +1 case doesn't make sense. It should be at least +2

I'd do something like this:
```ts
const itemsToShow = countChildren <= maxItems ? countChildren : maxItems - 1;
const moreItemsCount = countChildren = itemsToShow

return (
    <Inline space={space}>
        {React.Children.toArray(children).slice(0, itemsToShow)}
        {moreItemsCount > 1 && (
            <div
```

src/__stories__/stacking-group-story.tsx
```diff
@@ -0,0 +1,88 @@
...
+export const Default: StoryComponent<Args> = ({
+    stacked,
+    size,
+    type,
+    maxItems,
+    itemsToInclude,
+    inverse = false,
+}) => {
+    return (
+        <div data-testid="stacking-group" style={{width: 'max-content'}}>
```
atabel, line 34: you don't need this div, and why do you add the width: 'max-content' ? it's not needed

If you want the data-testid for the screenshot test, you can use the `dataAttributes` prop in ResponsiveLayout or Box if you want:

```ts
<ResponsiveLayout isInverse={inverse} dataAttributes={{testid: 'stacking-group'}}>
```

src/__stories__/stacking-group-story.tsx
```diff
@@ -31,8 +31,8 @@ export const Default: StoryComponent<Args> = ({
     inverse = false,
 }) => {
     return (
-        <ResponsiveLayout isInverse={inverse} dataAttributes={{testid: 'stacking-group'}}>
-            <Box paddingY={24}>
+        <ResponsiveLayout isInverse={inverse}>
+            <Box padding={16} width="max-content" dataAttributes={{testid: 'stacking-group'}}>
```
atabel, line 35: just set paddingY, the horizontal padding is already managed by ResponsiveLayout

atabel (approved):

atabel:
@BrunoGuimaraes103, please resolve the conflicts with master and I think we can merge this PR
Thank you  ;)

aweell:
The current border radius definition for the border when the item type is an image should be reviewed:

Both have 8px causing this gap:

Screenshot 2023-05-26 at 12 48 54

To solve this maybe it would be interesting to calculate the outer border-radius as the sum of the border width + the item border radius.

PR #745, opened 2023-05-08
feat(HorizontalScroll): new component

atabel:


src/horizontal-scroll.css.ts
yceballost, line 1: is it possible to show a horizontal scroll bar in desktop breakpoints? It is quite difficult to use this component on desktop without feedback

yceballost (changes requested):

src/horizontal-scroll.css.ts
brtbrt: AFAIK on mac os that cannot be accomplished natively, unless told explicitly on the preference pane:
scrolltip

I'll leave a solution I used to apply on an external project: it allows to show the scrollbar when hovering the area. it works for any OS, any browser (or it's supposed to)

https://codepen.io/chriscoyier/pen/RqKJdO
https://s3-us-west-2.amazonaws.com/s.cdpn.io/3/simple-scrollbar.js

atabel: I think it's better to honor the OS settings in this case, so I won't hide the horizontal scroll bar in desktop. @yceballost, please can you add a note in the specs?

yceballost: that's a good way, thanks!

brtbrt (approved):

src/horizontal-scroll.css.ts
marcoskolodny: Why not allowing scrollbar in touch devices?

atabel: It's quite rare to have scrollbars in mobile apps

marcoskolodny (approved):

PR #746, opened 2023-05-08
chore(tokens): generate colors types too

atabel:
It's common that mistica-design adds/changes color names. This PR generates the `Colors` type too


src/skins/types/index.tsx
atabel, line 1: moved `skin/types.tsx` -> `skin/types/index.tsx`
extracted `type Colors` to `skin/types/colors.tsx` file
The `colors.tsx` will be autogenerated

brtbrt (approved):

packages/generate-design-tokens/index.js
```diff
@@ -111,7 +125,13 @@ const generateSkinFiles = () => {
 
         const skinSrc = generateSkinSrc(skinName);
         fs.writeFileSync(path.join(SKINS_FOLDER, `${skinName}.tsx`), skinSrc);
+        anyGeneratedSkin = skinName;
     });
+
+    if (anyGeneratedSkin) {
```
marcoskolodny, line 131: why using this if statement? As far as I can see,  it will always be true unless `KNOWN_SKINS` is empty

marcoskolodny (approved):

PR #748, opened 2023-05-09
fix(Video): logic for video container

marcoskolodny:


src/video.css.ts
```diff
@@ -13,4 +13,17 @@ export const video = sprinkles({
...
+export const videoElement = sprinkles({
+    width: '100%',
+    height: '100%',
+    position: 'absolute',
+});
+
+export const posterElement = sprinkles({
+    width: '100%',
+    height: '100%',
+    position: 'relative',
```
atabel, line 28:
```suggestion
    position: 'absolute',
```

atabel (dismissed):

aweell (approved):
LGTM!

src/video.css.ts
marcoskolodny: fixed

src/video.tsx
```diff
@@ -193,6 +199,10 @@ const Video = React.forwardRef<HTMLVideoElement, VideoProps>(
                 style={{
                     // For some reason adding this style with classnames doesn't add the border radius in safari
                     borderRadius: !borderRadiusContext ? 0 : vars.borderRadii.container,
+                    visibility: showPoster ? 'hidden' : 'visible',
+                    position: showPoster || ratio !== 0 ? 'absolute' : 'static',
+                    width: showPoster ? posterWidth : '100%',
+                    height: showPoster ? posterHeight : '100%',
```
atabel, line 205: In the case of `aspectRatio={0}`, if poster image and video have different aspect ratios, there will be a weird jump. But I think there is no way to avoid that

src/video.tsx
```diff
@@ -162,6 +162,9 @@ const Video = React.forwardRef<HTMLVideoElement, VideoProps>(
             }
         });
 
+        const showPoster = videoStatus === 'error' || videoStatus === 'loading' || videoStatus === 'loaded';
+        const {ref: posterRef, width: posterWidth, height: posterHeight} = useElementDimensions();
```
atabel, line 166: you didn't find a way of doing this with css?

atabel (approved):

src/video.tsx
marcoskolodny: This is expected behaviour, and it's how video with poster works in html

PR #749, opened 2023-05-09
chore(ci): limit concurrency of CI workflows

atabel:
This will avoid race conditions of old jobs finishing later than the jobs of newer commits


marcoskolodny (approved):
Nice one!

PR #758, opened 2023-05-16
fix(Form): export formValues type in form

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1394)


src/form.tsx
```diff
@@ -16,7 +16,7 @@ if (
     import('scroll-behavior-polyfill').finally(() => {});
 }
 
-type FormValues = {[name: string]: any};
+export type FormValues = {[name: string]: any};
```
atabel, line 19: you need to re-export it in `index.tsx` to make it public

marcoskolodny: fixed

atabel (approved):

PR #621, opened 2022-12-16
feat(DataCard, MediaCard): remove extra top padding

atabel:

PR #759, opened 2023-05-16
feat(Chip): Add ability to click Chips (#621)

Winde:
Requested through CMS-3021. Chips should now be clickable, not compatible with closable


vdedios (approved):

yceballost (approved):

src/chip.tsx
```diff
@@ -76,24 +101,59 @@ const Chip: React.FC<ChipProps> = ({Icon, children, id, dataAttributes, active,
...
-                className={classnames(
-                    styles.chipVariants[active ? 'active' : overAlternative ? 'overAlternative' : 'default'],
-                    {
-                        [styles.chipInteractiveVariants[isDarkMode ? 'dark' : 'light']]: isInteractive,
-                    }
-                )}
-                paddingLeft={paddingLeft}
-                paddingRight={12}
-                {...getPrefixedDataAttributes(dataAttributes, 'Chip')}
+            <BaseTouchable trackingEvent={props.trackingEvent} onPress={props.onPress} as="a">
```
atabel, line 125: why `as="a"` ?

src/chip.tsx
```diff
@@ -76,24 +101,59 @@ const Chip: React.FC<ChipProps> = ({Icon, children, id, dataAttributes, active,
...
         );
     }
+
+    if (props.href) {
+        return (
+            <BaseTouchable
+                trackingEvent={props.trackingEvent}
+                href={props.href}
+                newTab={props.newTab}
+                dataAttributes={dataAttributes}
```
atabel, line 149: why do you pass `dataAttributes` here? they are already applied in `renderContent`

elenagarcia5 (approved):

src/chip.tsx
Winde: will remove

src/chip.tsx
Winde: If I remove the as="a"

The styles change due to BaseTouchable:
image

Seems like button is styled

atabel: hmmm, I see. I think I'd apply some styles to reset the button builtin styles, but I prefer this to be a `button`, not an `a`.
Something like this should work:
```js
sprinkles({
    border: 'none',
    background: 'transparent',
    padding: 0,
})
```

Winde: applied css reset and reverted to button

PR #762, opened 2023-05-17
feat(Grid): new component

atabel:
jira: WEB-1398


src/__stories__/grid-story.tsx
```diff
@@ -0,0 +1,262 @@
...
+            </Grid>
+        </Box>
+    </ResponsiveLayout>
+);
+
+AutoColumns.storyName = 'Auto columns';
+AutoColumns.args = {
+    gap: 16,
+    columnsMinSize: 100,
+    numItems: 10,
```
marcoskolodny, line 261: Maybe we should also allow setting width/height in controls?

src/grid.tsx
```diff
@@ -0,0 +1,209 @@
...
+                        ? typeof columns === 'number'
+                            ? styles.gridTemplateColumns[columns]
+                            : styles.gridTemplateColumnsAutoRepeat
+                        : '',
+                    rows
+                        ? typeof rows === 'number'
+                            ? styles.gridTemplateRows[rows]
+                            : styles.gridTemplateRowsAutoRepeat
+                        : '',
+                    flow ? styles.gridAutoFlow[flow] : '',
```
marcoskolodny, line 122: There is no need for the undefined check in here. If `flow` is undefined, the resulting value will be undefined, and `classnames` works fine when that happens.

aweell (approved):

src/grid.tsx
```diff
@@ -0,0 +1,209 @@
+import {assignInlineVars} from '@vanilla-extract/dynamic';
+import classNames from 'classnames';
+import * as React from 'react';
+import * as styles from './grid.css';
+import {getPrefixedDataAttributes} from './utils/dom';
+
+import type {DataAttributes} from './utils/types';
+
+type RowsColumns = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12;
+type Gap = 0 | 2 | 4 | 8 | 12 | 16 | 24 | 32 | 40 | 48 | 56 | 64 | 72 | 80;
```
marcoskolodny, line 10: According to CSS docs, gap can be also a percentage ([Link](https://developer.mozilla.org/en-US/docs/Web/CSS/gap))

src/grid.tsx
```diff
@@ -0,0 +1,209 @@
...
+        | 'center'
+        | 'stretch'
+        | 'space-around'
+        | 'space-between'
+        | 'space-evenly';
+    alignContent?: 'start' | 'end' | 'center' | 'stretch' | 'space-around' | 'space-between' | 'space-evenly';
+    width?: string | number;
+    height?: string | number;
+    minWidth?: string | number;
+    minHeight?: string | number;
```
marcoskolodny, line 76: why not including max width/height in props?

src/grid.tsx
atabel: yes, and according to CSS docs it can be 3 or 5 too :P but I think we don't want to support those values in mistica. But @aweell, if you find supporting % useful for mistica I can add support

src/grid.tsx
atabel: I need the check, otherwise TS will error:
image

src/grid.tsx
marcoskolodny: Why not supporting %? I think it's not rare for an user to use gap with percentages.

aweell: I see gap here equivalent to the `space` prop in the Inline or the Stack components, if percentage values weren't needed there I wouldn't add support until the need arises.

src/__stories__/grid-story.tsx
atabel: I don't feel it would be very useful in this story

src/grid.tsx
marcoskolodny: Discussed offline. We will remove width/minWidth

marcoskolodny (approved):

PR #767, opened 2023-05-29
fix(DateField): input type for onBlur/onFocus prop

marcoskolodny:
Discussion: [Link](https://github.com/Telefonica/webapp/pull/3039#discussion_r1185951534)
Related PR in webapp: [Link](https://github.com/Telefonica/webapp/pull/3149)


src/date-field.tsx
```diff
@@ -10,8 +10,9 @@ import {isFirefox} from './utils/platform';
 
 import type {CommonFormFieldProps} from './text-field-base';
 
-export interface DateFieldProps extends CommonFormFieldProps {
```
atabel, line 13: maybe we could make `CommonFormFieldProps` generic and have something like `CommonFormFieldProps<HTMLInputElement>`, we could even make `HTMLInputElement` the default generic value, so `CommonFormFieldProps` would be equivalent to `CommonFormFieldProps<HTMLInputElement>`, in case we need a textarea we'd use `CommonFormFieldProps<HTMLTextAreaElement>`

src/text-field.tsx
```diff
@@ -20,7 +20,7 @@ const useKeepMaxLength = (
     }, [onChange, value, maxLength, input]);
 };
 
-export interface TextFieldProps extends CommonFormFieldProps {
+export interface TextFieldProps extends CommonFormFieldProps<Element> {
```
atabel, line 23: why `<Element>` ?

atabel (approved):

PR #770, opened 2023-05-31
feat(Blocks): new Mistica Community Components

daviteixeira-tlf:


src/advanced-slots.tsx
```diff
@@ -0,0 +1,350 @@
...
+export const HighlightedValueBlock: React.FC<HighlightedValueBlockProps> = ({
+    tag,
+    value,
+    text,
+    value2,
+    text2,
+    secondaryValue,
+    title,
+    description,
+    textColor = vars.colors.textPrimary,
```
yceballost, line 210: valueColor fit better than text?

src/advanced-slots.tsx
```diff
@@ -0,0 +1,350 @@
...
+import type Image from './image';
+import type Tag from './tag';
+import type {RendersNullableElement} from './utils/renders-element';
+
+interface SlotContentProps {
+    title?: string;
+    description?: string | Array<string>;
+}
+
+const SlotContent: React.FC<SlotContentProps> = ({title, description}) => {
```
yceballost, line 22: these components can be used out of the slots. Maybe we need to rename this to block?

src/advanced-slots.tsx
yceballost, line 1: should we change the name of the file to `blocks.tsx`?

src/advanced-slots.tsx
```diff
@@ -0,0 +1,350 @@
...
+        </div>
+    );
+};
+
+interface ProgressBlockProps {
+    title?: string;
+    stackingGroup?: RendersNullableElement<typeof StackingGroup>;
+
+    progressPercent?: number;
+
```
yceballost, line 296: @wyunreal progressBar has a property to reverse animation... should we include a prop here to access to this progress prop? Would it be useful for the CMS to be able to control this for the "te quedan" / "llevas" case?

https://tinyurl.com/2hpfba3y

src/advanced-slots.tsx
```diff
@@ -0,0 +1,350 @@
...
+                            </div>
+                        </div>
+                    )}
+                </>
+            )}
+        </div>
+    );
+};
+
+interface SimpleBlockProps {
```
yceballost, line 118: buggy for long texts 😬
https://tinyurl.com/2gz238s7

yceballost (changes requested):

src/advanced-slots.tsx
yceballost: I think the component should grow without limitation by default. We can include the possibility to add truncation like in cards, but force ellipsis by default is a bad practice in terms of accessibility.

If you need this truncation, include it as a prop and add it to specs @vinrosa 

example of truncation prop
https://tinyurl.com/2gwfsruu

src/advanced-slots.tsx
daviteixeira-tlf: Done

src/advanced-slots.tsx
```diff
@@ -0,0 +1,350 @@
...
+}
+
+export const SimpleBlock: React.FC<SimpleBlockProps> = ({image, description, 'aria-label': ariaLabel}) => {
+    return (
+        <div aria-label={ariaLabel}>
+            <div
+                className={sprinkles({
+                    display: 'flex',
+                    alignItems: 'center',
+                    height: 40,
```
yceballost: I think that the component just has to hug the content. So, I think that you can remove this height

src/blocks.tsx
```diff
@@ -0,0 +1,350 @@
...
+}) => {
+    return (
+        <div
+            style={{
+                display: 'flex',
+                alignItems: 'center',
+            }}
+            aria-label={ariaLabel}
+        >
+            {(title || stackingGroup || description) && (
```
yceballost, line 70: I find the behavior of "not having the title prop, the stackingGroup is not showing" quite odd.
https://tinyurl.com/2lalc6qz

I would appreciate the flexibility to use this component with just the title + progress, or solely with the stackingGroup + progress, for example

I plan to discuss this matter with the Vivo design team to propose a change.

src/advanced-slots.tsx
daviteixeira-tlf: But in the design specifications it has a fixed size, both with or without an image

src/blocks.tsx
```diff
@@ -0,0 +1,343 @@
+import * as React from 'react';
```
atabel, line 1: please, move all these components to the community folder.
FYI: https://github.com/Telefonica/mistica-web#mistica-community and https://mistica-web.vercel.app/?path=/story/community-welcome--default

atabel, line 1: I miss stories and screenshot tests

atabel:
Please, include a link to figma specs in the PR

src/advanced-slots.tsx
atabel: we can never have fixed heights because height depends on the text content, and font size can be bigger (for accesibility reasons) if the user has change their phone settings (you can use storybook font size control to simulate this).

yceballost (changes requested):

src/community/__stories__/blocks-story.tsx
```diff
@@ -0,0 +1,256 @@
...
+                    secondaryValue={secondaryValue}
+                    text={text}
+                    aria-label="aria-label"
+                />
+            </Box>
+        </ResponsiveLayout>
+    );
+};
+
+BlockHighlightedValue.storyName = 'Highlighted Value';
```
yceballost, line 157:
```suggestion
BlockHighlightedValue.storyName = 'Highlighted Value Block';
```

yceballost (changes requested):

BrunoGuimaraes103:
@yceballost @atabel Hello, I just created the test-screenshot of the blocks

src/blocks.tsx
```diff
@@ -0,0 +1,343 @@
...
+                <Text5 color={valueColor}>{value}</Text5>
+            </div>
+        </Inline>
+    );
+};
+
+interface HighlightedValueBlockProps {
+    tag?: RendersNullableElement<typeof Tag>;
+
+    value: string | Array<string>;
```
kydorn, line 185: we shouldn't accept here an Array if we only process the first two elements

src/blocks.tsx
```diff
@@ -0,0 +1,343 @@
...
+    tag,
+    value,
+    text,
+    secondaryValue,
+    title,
+    description,
+    valueColor = vars.colors.textPrimary,
+    'aria-label': ariaLabel,
+}) => {
+    const [firtValue, secondValue] = value && Array.isArray(value) ? value : [value, ''];
```
kydorn, line 207: typo: firstValue

src/blocks.tsx
```diff
@@ -0,0 +1,343 @@
...
+                >
+                    {tag}
+                </div>
+            )}
+
+            <Stack space={2}>
+                <Inline space={8} alignItems="baseline">
+                    <Text8 color={valueColor}>{firtValue}</Text8>
+                    <Text2 regular color={vars.colors.textSecondary}>
+                        {firtText}
```
kydorn, line 226: we are allowing here a secondaryValue without a fistValue, is this correct? 

Maybe we should pass mainValue and secondaryValue as separate props, being both an object providing the value and the text

Something like:

mainHeading={{value: 'Heading', text: 'small text to the right'}}
secondaryHeading={{value: 'Heading2', text: 'small text to the right'}}

This way, we can't have a text without it's 'heading', and we are not using an array when we only have the possibility of two headings.

src/blocks.tsx
```diff
@@ -182,8 +182,15 @@ export const InformationBlock: React.FC<InformationBlockProps> = ({
...
-    value: string | Array<string>;
-    text: string | Array<string>;
+    mainHeading: {
+        value: string;
+        text: string;
+    };
+
+    secondHeading?: {
+        value: string;
+        text: string;
```
kydorn, line 192: this should be optional, right?

src/blocks.tsx
```diff
@@ -182,8 +182,15 @@ export const InformationBlock: React.FC<InformationBlockProps> = ({
 interface HighlightedValueBlockProps {
     tag?: RendersNullableElement<typeof Tag>;
 
-    value: string | Array<string>;
-    text: string | Array<string>;
+    mainHeading: {
+        value: string;
+        text: string;
```
kydorn, line 187: this should be optional, right?

src/blocks.tsx
```diff
@@ -182,8 +182,15 @@ export const InformationBlock: React.FC<InformationBlockProps> = ({
...
+    };
+
+    secondHeading?: {
+        value: string;
+        text: string;
+    };
 
     secondaryValue?: string;
```
kydorn, line 195: is this used?

src/blocks.tsx
```diff
@@ -182,8 +182,15 @@ export const InformationBlock: React.FC<InformationBlockProps> = ({
 interface HighlightedValueBlockProps {
     tag?: RendersNullableElement<typeof Tag>;
 
-    value: string | Array<string>;
-    text: string | Array<string>;
+    mainHeading: {
```
kydorn, line 185: as a suggestion, you could use string | {value: string, text: string}, so if you pass a string here, it uses it as 'value'. This will reduce verbosity in places in which you don't need a 'text'

src/community/__stories__/blocks-story.tsx
```diff
@@ -144,9 +145,8 @@ export const BlockHighlightedValue: StoryComponent<HighlightedValueBlockArgs> =
                     tag={<Tag type={tag}>{`${tag}`}</Tag>}
                     title={title}
                     description={description}
-                    value={value}
+                    mainHeading={{value, text}}
                     secondaryValue={secondaryValue}
```
kydorn, line 149: not used, right?

src/blocks.tsx
BrunoGuimaraes103: yes, the second headline is optional

src/blocks.tsx
BrunoGuimaraes103: The mainHeading, I believe it is not optional

src/blocks.tsx
BrunoGuimaraes103: yes, it is used in this part of the code:

Captura de tela 2023-06-29 092150

src/community/__stories__/blocks-story.tsx
BrunoGuimaraes103: in this case the value and text are being used to pass the object with their information in the mainHeadling

src/blocks.tsx
BrunoGuimaraes103: Thank you for the suggestion, I tried to use this way but ended up giving some errors

BrunoGuimaraes103:
@atabel Is there anything missing on our part for this PR to be approved?

src/community/__stories__/blocks-story.tsx
```diff
@@ -0,0 +1,256 @@
...
+    return (
+        <ResponsiveLayout>
+            <Box paddingY={24} dataAttributes={{testid: 'value-block'}}>
+                <ValueBlock title={title} description={description} value={value} aria-label="aria-label" />
+            </Box>
+        </ResponsiveLayout>
+    );
+};
+
+BlockValue.storyName = 'Value Block';
```
atabel, line 189:
```suggestion
BlockValue.storyName = 'ValueBlock';
```

src/community/__stories__/blocks-story.tsx
```diff
@@ -0,0 +1,256 @@
...
+                    text={text}
+                    description={description}
+                    aria-label="aria-label"
+                />
+            </Box>
+        </ResponsiveLayout>
+    );
+};
+
+BlockProgress.storyName = 'Progress Block';
```
atabel, line 247:
```suggestion
BlockProgress.storyName = 'ProgressBlock';
```

src/__screenshot_tests__/blocks-screenshot-test.tsx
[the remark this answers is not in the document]
BrunoGuimaraes103: Unfortunately, a new test folder would have to be created within community for the test-screenshot, I tried to run the test with the file in community but I got this error:
Captura de tela 2023-07-05 091136

src/community/blocks.tsx
```diff
@@ -0,0 +1,347 @@
...
+
+    progressPercent?: number;
+    reverse?: boolean;
+
+    value: string;
+    text: string;
+    description?: string;
+
+    valueColor?: string;
+    'aria-label'?: string;
```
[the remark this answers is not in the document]
BrunoGuimaraes103: I changed the value and text to be passed together in an object, in the example of figma it seems that the heading is the only one required, the example of use cases are in figma -> Slots - Props

I also changed value to accept number as well as string

src/community/blocks.tsx
```diff
@@ -0,0 +1,347 @@
...
+    return (
+        <div aria-label={ariaLabel}>
+            <div>
+                {title && (
+                    <Box paddingBottom={progressPercent ? 8 : 0}>
+                        <Inline space="between" alignItems="flex-end">
+                            <div style={{paddingRight: '32px'}}>
+                                <Text2 regular>{title}</Text2>
+                            </div>
+                            {stackingGroup && <div style={{zIndex: '0'}}>{stackingGroup}</div>}
```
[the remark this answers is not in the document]
BrunoGuimaraes103: as per the use case in Figma, yes, the Stacking Group cannot be rendered if it does not have the title

src/community/blocks.tsx
```diff
@@ -0,0 +1,347 @@
...
+        </div>
+    );
+};
+
+interface ValueBlockProps {
+    title?: string;
+    value?: string;
+    description?: string | Array<string>;
+    valueColor?: string;
+    'aria-label'?: string;
```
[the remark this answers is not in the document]
BrunoGuimaraes103: In fact it is quite strange to be like this, however in the specifications in figma -> Handoff -> Slots - Props, it is specified that they are optional for greater versatility of use

src/community/blocks.tsx
```diff
@@ -0,0 +1,347 @@
...
+    description?: string | Array<string>;
+}
+
+const BlockContent: React.FC<BlockContentProps> = ({title, description}) => {
+    const normalizedDescription =
+        description && Array.isArray(description) ? (
+            description.map((paragraph, i) => (
+                <Text2 regular color={vars.colors.textSecondary} key={i}>
+                    {paragraph}
+                </Text2>
```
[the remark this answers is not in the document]
BrunoGuimaraes103: The intended behavior is to allow passing an array of description to the ValueBlock, so that multiple descriptions can be passed, and the reason why it is set to Text2 is because the Text2 specification is regular in Figma -> handoff -> Slots - text and color

The reason it is not also placed as an as='p' is that this customization option is only for the title and pretitle

src/community/blocks.tsx
```diff
@@ -0,0 +1,347 @@
...
+        value: string;
+        text: string;
+    };
+
+    secondHeading?: {
+        value: string;
+        text: string;
+    };
+
+    secondaryValue?: string;
```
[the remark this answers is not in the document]
BrunoGuimaraes103: In this case the secondHeading is a second header that is mandatory to have both value and text.

The secondaryValue, on the other hand, would be a second value (in relation to the mainHeading value) with a different color and without the need for a text next to it.

Before, an array of heading elements was allowed, but as suggested in this PR and because it only has 2 headings, with no further possibility, in the documentation, these two mainHeading and secondaryHeading objects were created

src/community/blocks.tsx
```diff
@@ -0,0 +1,347 @@
...
+                    display: 'flex',
+                    alignItems: 'center',
+                })}
+            >
+                {image && (
+                    <div
+                        className={sprinkles({
+                            paddingRight: 16,
+                        })}
+                        style={{zIndex: '0'}}
```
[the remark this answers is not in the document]
BrunoGuimaraes103: the z Index:0 was because it was not possible to click on the images when it was the anchor, but with the current state of the Advanced-Data-Card, this is no longer necessary

src/community/blocks.tsx
```diff
@@ -0,0 +1,347 @@
...
+}) => {
+    return (
+        <div
+            style={{
+                display: 'flex',
+                alignItems: 'center',
+            }}
+            aria-label={ariaLabel}
+        >
+            {(title || stackingGroup || description) && (
```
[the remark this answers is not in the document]
BrunoGuimaraes103: These blocks are highly customizable targeting many use cases, which is why the vast majority of them have optional properties in all of their props.

src/community/blocks.tsx
```diff
@@ -0,0 +1,347 @@
...
+    );
+};
+
+interface InformationBlockProps {
+    title?: string;
+    description?: string | Array<string>;
+    value: string;
+    secondaryValue?: string;
+    valueColor?: string;
+    'aria-label'?: string;
```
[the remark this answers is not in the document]
BrunoGuimaraes103: From what I saw in figma, it looks like mainValue was also optional, just missing a use case showing the block without the value, I just fixed that

src/community/__stories__/blocks-story.tsx
```diff
@@ -44,55 +50,60 @@ export const BlockRow: StoryComponent<RowBlockArgs> = ({title, description, show
...
     );
 };
 
-BlockRow.storyName = 'Row Block';
+BlockRow.storyName = 'RowBlock';
 BlockRow.args = {
     title: 'title',
     description: 'description',
     showStackingGroup: false,
+    ariaLabel: 'aria-label',
```
atabel, line 60: I think it doesn't make sense to include a storybook control for ariaLabel, as the result of changing it isn't visual.

src/community/blocks.tsx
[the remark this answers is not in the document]
atabel: you are not responding my question.

What's the difference of sending an array of string vs just an string in `description`?
`description="some description"` vs `description={['some', 'description']}` are these equivanet? If they are, then why do we support both?

src/community/blocks.tsx
```diff
@@ -0,0 +1,293 @@
...
+                    space="between"
+                    alignItems="center"
+                >
+                    {title && (
+                        <Box paddingRight={32}>
+                            <Text2 regular>{title}</Text2>
+                        </Box>
+                    )}
+                    {stackingGroup ? (
+                        <>{stackingGroup}</>
```
atabel, line 94: you dont need the fragment
```suggestion
                        stackingGroup
```

src/community/blocks.tsx
```diff
@@ -0,0 +1,293 @@
...
+}) => {
+    return (
+        <div
+            style={{
+                display: 'flex',
+                alignItems: 'center',
+            }}
+            aria-label={ariaLabel}
+        >
+            {(title || stackingGroup || description) && (
```
atabel, line 79: I'd remove this condition

src/community/blocks.tsx
```diff
@@ -0,0 +1,347 @@
...
+                    {secondaryValue}
+                </Text2>
+                <Text5 color={valueColor}>{value}</Text5>
+            </div>
+        </Inline>
+    );
+};
+
+interface HighlightedValueBlockProps {
+    tag?: RendersNullableElement<typeof Tag>;
```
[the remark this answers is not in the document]
atabel: please, don't respond a comment with :+1: if you have not solved it

src/community/blocks.tsx
```diff
@@ -0,0 +1,293 @@
...
+    'aria-label': ariaLabel,
+}) => {
+    return (
+        <div aria-label={ariaLabel}>
+            <Stack space={8}>
+                <Inline space="between" alignItems="flex-end">
+                    <Box paddingRight={32}>
+                        <Text2 regular>{title}</Text2>
+                    </Box>
+                    {stackingGroup && <>{stackingGroup}</>}
```
atabel, line 276: just
```suggestion
                    {stackingGroup}
```

src/community/blocks.tsx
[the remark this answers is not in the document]
atabel: hmmm.. I still see this quite confusing. Can I have secondHeading and secondaryValue at the same time in the component?

src/community/blocks.tsx
```diff
@@ -0,0 +1,293 @@
...
+    stackingGroup,
+    description,
+    'aria-label': ariaLabel,
+}) => {
+    return (
+        <div
+            style={{
+                display: 'flex',
+                alignItems: 'center',
+            }}
```
atabel, line 76: remove this, it's not needed

src/community/blocks.tsx
```diff
@@ -0,0 +1,293 @@
...
+            aria-label={ariaLabel}
+        >
+            {(title || stackingGroup || description) && (
+                <Inline
+                    className={sprinkles({
+                        display: 'flex',
+                        flex: 1,
+                    })}
```
atabel, line 84: remove this, it's not needed

src/community/blocks.tsx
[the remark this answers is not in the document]
atabel: It doesn't make sense to have a `ProgressBlock` without a `ProgresBar` If figma specs say so, then figma specs must be wrong.

Why did you changed  value to accept number? if it should only accept numbers, then type it as `number`, if it should accept any string then type it as `string` but don't type it as `number | string`, please

src/community/blocks.tsx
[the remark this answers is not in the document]
BrunoGuimaraes103: Sorry, I thought you were referring to the storybook, but I've corrected that error now.

src/community/blocks.tsx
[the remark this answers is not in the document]
BrunoGuimaraes103: we did it this way to facilitate passing the prop in case the programmer wanted to pass only one element

src/community/blocks.tsx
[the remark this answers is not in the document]
BrunoGuimaraes103: yes, you can:
image

src/community/blocks.tsx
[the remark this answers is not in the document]
BrunoGuimaraes103: It happens that there are cases that do not have consumption, so there is this possibility that the progressBar does not appear.

Already for value, I will change then to accept only number

src/community/blocks.tsx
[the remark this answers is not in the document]
atabel: Please, don't do that. We shouldn't support more than one way of doing the same thing

src/community/blocks.tsx
```diff
@@ -0,0 +1,277 @@
...
+                        <Text2 regular color={vars.colors.textSecondary}>
+                            {secondHeading.text}
+                        </Text2>
+                    </Inline>
+                )}
+                {secondaryValue && <Text8 color={vars.colors.textSecondary}>{secondaryValue}</Text8>}
+            </Stack>
+            <Box paddingTop={8}>
+                <BlockContent title={title} description={description} />
+            </Box>
```
atabel, line 194: I'd render the Box conditionally. If there isn't title/description you don't need an extra padding:
```suggestion
            {title || description ? (
                <Box paddingTop={8}>
                    <BlockContent title={title} description={description} />
                </Box>
            ) : null}
```

src/community/blocks.tsx
```diff
@@ -0,0 +1,277 @@
...
+    reverse?: boolean;
+
+    heading: {
+        value: number;
+        text: string;
+    };
+
+    description?: string;
+
+    valueColor?: string;
```
atabel, line 239: shouldn't this be inside heading?
```ts
heading: {
    value: number;
    valueColor?: string;
    text: string;
}
```

src/community/blocks.tsx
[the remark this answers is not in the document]
atabel: Then, what's the difference between a ProgressBlock without progress and a HighlightedValueBlock?
```ts
<ProgressBlock heading={{value: 34, text: 'some text'}} description="some description"/>
<HighlightedValueBlock mainHeading={{value: 34, text: 'some text'}} description="some description" />
```
image

atabel (changes requested):

src/community/blocks.tsx
BrunoGuimaraes103: I don't think so, as it could have more than one description and only one header and thus the prop color would not allow the values ​​of the two headers to be of different colors

image

src/community/blocks.tsx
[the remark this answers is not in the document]
BrunoGuimaraes103: In fact, you can customize them in this way where there is no difference between them, because the blocks were thought to have great versatility due to several business rules.

The difference is that each one has its own resources that would allow a more unique use, for example in the one you placed each one would have the possibility of using its own props like the HighlightedValueBlock being able to use the tags while progressBlock a stackingGroup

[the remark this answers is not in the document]
atabel: I don't like the idea of having two different block components that can be used to render the same design

src/community/blocks.tsx
```diff
@@ -0,0 +1,277 @@
...
+    title,
+    description,
+    valueColor = vars.colors.textPrimary,
+    'aria-label': ariaLabel,
+}) => {
+    return (
+        <div aria-label={ariaLabel}>
+            {headline && <Box paddingBottom={24}>{headline}</Box>}
+
+            <Stack space={2}>
```
Winde, line 175: Is this component the one to render this?

image

For first slot:
I assume 300 Mega => would be mainHeading.value
I assume de velocidade => would be mainHeading.text

But there's no possibility of putting a text above this mainHeading for "Vivo Fibra 300 Mbps Especial Netflix_1"
As headline is a Tag.

----------------------

For second slot, similar problem:

I assume 110 canais is => mainHeading.value

But mainHeading.text is mandatory.
And there's no possibility of putting a text above this mainHeading for "Vivo Play Avançado HD"
As headline is a Tag.

**************

The other alternative is that we are supposed to use the ProgressBlock for this but without a progress bar?

Note: screenshot tests for the actual cases we want to use would be helpful in identifying which block to be used in each case.

Winde (approved):

Winde (changes requested):
Check the use case for landline and TV, there's some type misalignment

Neither Highlighted nor Progress blocks seem fit to do those cases

src/community/blocks.tsx
atabel: @BrunoGuimaraes103, this is important. Can you check the definition with the designers? The blocks should be able to cover these cases

BrunoGuimaraes103: We are checking about it

BrunoGuimaraes103: @Winde Sorry, I'm kind of confused on what you mean, could you be a little clearer?

The use cases in question can be done using ProgressBlock and ValueBlock as shown in this image:
image

Winde: It's a bit weird:
(1) to have to use a Progress block without progress bar
(2) That these two cases are actually different blocks

but fair enough :)

Winde (approved):

src/community/blocks.tsx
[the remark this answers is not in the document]
atabel: @BrunoGuimaraes103, this is still pending, please remove this description Array

src/community/blocks.tsx
atabel: yes, it's quite weird. I said the same in other comment :(

atabel (changes requested):

src/community/blocks.tsx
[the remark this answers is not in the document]
BrunoGuimaraes103: unfortunately I can't remove this Array because it's part of the demand in these figma use cases:

Captura de tela 2023-07-12 085419
Captura de tela 2023-07-12 085427

src/community/blocks.tsx
```diff
@@ -0,0 +1,267 @@
...
+
+import type StackingGroup from '../stacking-group';
+import type Image from '../image';
+import type Tag from '../tag';
+import type {RendersNullableElement} from '../utils/renders-element';
+import type {ExclusifyUnion} from '../utils/utility-types';
+
+interface BlockContentProps {
+    title?: string;
+    description?: Array<string>;
```
atabel, line 19: Please @BrunoGuimaraes103 , I've commented this multiple times, this description shouldn't be an `Array`, it should be just a `string`

atabel (changes requested):

src/community/blocks.tsx
atabel: fixed

atabel (approved):

PR #777, opened 2023-06-06
fix(TextField, Select): improve styles

yceballost:
- [x] perfect alignment in select input with text field in chrome
- [x] perfect alignment in select input with text field in safari 
- [x] Changed input password text to use Lucida Grande instead of Arial in MacOS
- [x] Changed icons in CVV tooltips to use Mística colors and use text presets
- [x] Changed icon in credit card field to use mistica icons. Removed the old one.
- [x] Changed search field icon to use mistica icons. Removed the old one.

⚠️ I changed the inherited fontSize and lineHeight in input field, now, input fields don't respond to font scale. We should improve this behavior in the future to adapt input box to the font size. I did it to make parity between select and input fields. Also because the behavior of text with scale in input fields is not working as best

alignment test → https://tinyurl.com/23u5s55u

---
Chrome
image

Safari
image


src/cvv-field.tsx
```diff
@@ -18,16 +22,21 @@ const TooltipContent = ({acceptedCards}: {acceptedCards: CardOptions}) => {
...
-                <div className={styles.tooltipContainer}>
-                    <IconCvvAmex size={48} role="img" />
-                    <p className={styles.cvvText}>{texts.formCreditCardCvvTooltipAmex}</p>
-                </div>
-            )}
+            <Box padding={8}>
+                <Stack space={8}>
+                    <Inline space={16} alignItems="center">
+                        <IconCvvVisaMc size={48} role="img" />
+                        <Text2>{texts.formCreditCardCvvTooltipVisaMc}</Text2>
```
yceballost, line 29: I can't understand why using `Text2` here the line height is not working

src/select.tsx
```diff
@@ -385,7 +386,19 @@ const Select: React.FC<SelectProps> = ({
 
                 <div
                     className={styles.selectTextVariants[disabled ? 'disabled' : 'default']}
-                    style={{top: label ? 27 : 17}}
+                    style={{
+                        top: isDesktopOrBigger
+                            ? label
+                                ? isSafari()
+                                    ? 28
+                                    : 28
```
marcoskolodny, line 394: If both are 28, why using conditional `isSafari()?`

src/select.tsx
```diff
@@ -385,7 +386,19 @@ const Select: React.FC<SelectProps> = ({
 
                 <div
                     className={styles.selectTextVariants[disabled ? 'disabled' : 'default']}
-                    style={{top: label ? 27 : 17}}
+                    style={{
+                        top: isDesktopOrBigger
+                            ? label
+                                ? 28
```
marcoskolodny, line 392: Are we sure this is necessary? I don't understand the logic behind all these numbers

marcoskolodny (approved):

src/select.tsx
marcoskolodny: Try something like this: [Playroom](https://mistica-web.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8UCWA3AOgOwAS4GcAXATwBsYBeYYHffABwgNSNQm3l0xACcYyAQzboYPADR16Ad1RQiACy48AzAAY1DAB4Sp%2BBTFQBzBUWUgArBu2689AEaCwAayO8IAV2xRz-KLfwAX0CcAD4pFAw9QlIKalo7elwmFjYOc0F7AggyDyIxEElEmTlFcwBGax1C6P1DEzNuEErNaqKk-EcXN09vc3shFwCOoggGLjV2jvsIIlGAWwmppPnBXiNUbAAVMYy8iGGVtY3sACFZhb3Rw9xgqQB6cOwke7R0J8KQRRh5mAIEADaIFOQg8AH1UAB5ADKIAAuuIQLJ5Ap-vAASoAExqOGBIA)

atabel: yes, please, can you explain all this logic? why do we need different styles for Safari?

src/text-field-base.tsx
```diff
@@ -226,10 +226,8 @@ export const TextFieldBase = React.forwardRef<any, TextFieldBaseProps>(
 
         /* Workaround to avoid huge bullets on ios devices (-apple-system font related) */
         const fontFamily =
-            rest.type === 'password' &&
-            isIos(theme.platformOverrides) &&
-            !isRunningAcceptanceTest(theme.platformOverrides)
-                ? 'arial'
+            rest.type === 'password' && !isRunningAcceptanceTest(theme.platformOverrides)
+                ? 'Lucida Grande, Arial, Arial, sans-serif'
```
atabel, line 230: Arial is repeated

src/select.tsx
yceballost: insane pixel perfect 😅
No safari logic
image

Safari logic
image

It is true that with the removal of inherit fontsize for input fields maybe it is not worth this pixel perfect. I can remove it if you consider it too much, no problem.

src/__screenshot_tests__/__image_snapshots__/double-field-screenshot-test-tsx-double-field-1-snap.png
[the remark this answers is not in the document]
yceballost: Yes, the old one was not using mistica icons. Maybe we can use this alternative, but I think is more evident with the mastercard logo detail

image

an alternative is to use the filled variant, but we do not use filled icons as a rule
image

src/select.tsx
marcoskolodny: I'm still not sure why Safari has differences. I believe the same results can be achieved with pure CSS. Maybe we could leave this logic as it is, but also create a ticket to see if we can handle it without ifs.

What do you think @atabel ?

src/__screenshot_tests__/__image_snapshots__/double-field-screenshot-test-tsx-double-field-1-snap.png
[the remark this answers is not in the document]
atabel: what about the color? the "i" (info) icon in CVV has a different color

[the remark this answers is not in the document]
yceballost: f*ck, true xD I didn't read the word "color" in your previous message. I will change it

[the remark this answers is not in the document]
yceballost: Changed.

I also replace cvv info icon to IconInformationRegular from Mística icons.

atabel (approved):

PR #780, opened 2023-06-07
feat(AdvancedDataCard): new community component

daviteixeira-tlf:
Figma: https://www.figma.com/file/35iS2dwoogulJZPgFGKmCX/Explora%C3%A7%C3%A3o-Data-Card-%2F-Slots?node-id=1-515&t=NuqYVox6UWkc3ZV7-0


atabel:
please include a link to Figma specs in the PR

src/data-card-advanced.css.ts
```diff
@@ -0,0 +1,147 @@
...
+    sprinkles({
+        paddingTop: 16,
+        paddingBottom: 24,
+    }),
+]);
+
+export const footerText = style([
+    {
+        width: 'auto',
+        height: 'auto',
```
atabel, line 131: redundant

src/data-card-advanced.css.ts
```diff
@@ -0,0 +1,147 @@
...
+
+        display: 'flex',
+        flex: '1',
+
+        lineClamp: '2',
```
atabel, line 136: why? we have Text components which already implement clampling styles/logic

src/data-card-advanced.css.ts
```diff
@@ -0,0 +1,147 @@
...
+            [mq.supportsHover]: {
+                ':hover': {
+                    background: applyAlpha(vars.rawColors.backgroundContainer, 0.9),
+                },
+            },
+        },
+    },
+]);
+
+export const TouchableZone = style([
```
atabel, line 116:
```suggestion
export const touchableZone = style([
```

src/data-card-advanced.css.ts
```diff
@@ -0,0 +1,147 @@
...
+    sprinkles({height: '100%', flex: 1}),
+    {display: 'flex', flexDirection: 'column', alignItems: 'flex-start', alignContent: 'flex-start'},
+]);
+
+export const slots = style([
+    sprinkles({
+        paddingTop: 16,
+        paddingBottom: 24,
+    }),
+]);
```
atabel, line 126: this may be just a Box?

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,463 @@
...
+    descriptionLinesMax,
+}) => {
+    const {textPresets} = useTheme();
+    const renderHeadline = () => {
+        if (!headline) {
+            return null;
+        }
+        if (typeof headline === 'string') {
+            return <Tag type="promo">{headline}</Tag>;
+        }
```
atabel, line 132: do we want to support this use case? I think you just copy&pasted this from card.tsx, but this is this way in other cards just to support a legacy use case where headline could be just a string

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,463 @@
...
+        <div>
+            <Stack space={4}>
+                {(headline || pretitle || title || subtitle) && (
+                    <header>
+                        <Stack space={4}>
+                            {renderHeadline()}
+                            <Stack space={4}>
+                                {pretitle && (
+                                    <Text2
+                                        aria-label="Texto"
```
atabel, line 145: what?

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,463 @@
...
+        });
+    }
+
+    return finalActions;
+};
+
+type CardContentProps = {
+    headline?: string | RendersNullableElement<typeof Tag>;
+    pretitle?: string;
+    pretitleAs?: string;
```
atabel, line 101: what's this used for?

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,463 @@
...
+    pretitleAs?: textAs;
+    title?: string;
+    titleLinesMax?: number;
+    titleAs?: textAs;
+    subtitle?: string;
+    subtitleLinesMax?: number;
+    description?: string;
+    descriptionLinesMax?: number;
+
+    slots?: Array<RendersNullableElement<slotsTypeof>>;
```
atabel, line 301: why is this an array?
I'd call this extra, to follow the same naming conventions we use in other components

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,463 @@
...
+    small?: boolean;
```
atabel, line 302: whats this? does it make the card smaller? otherwise is a bad prop name

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,463 @@
...
+    button?: RendersNullableElement<typeof ButtonPrimary>;
+    footerImage?: RendersNullableElement<typeof Image>;
+    footerText?: string;
+    footerTextLinesMax?: number;
+    buttonLink?: RendersNullableElement<typeof ButtonLink>;
+
+    dataAttributes?: DataAttributes;
+    actions?: Array<CardAction>;
+    'card-aria-label'?: string;
+    'touchable-aria-label'?: string;
```
atabel, line 313: why is this needed? shouldn't be the card aria label enough?

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,463 @@
...
+    );
+};
+
+type slotsTypeof = typeof StackingGroup;
+
+type textAs = 'h1' | 'h2' | 'h3' | 'h4';
+
+interface DataCardAdvancedProps {
+    cardOnPress?: () => void;
+
```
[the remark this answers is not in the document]
BrunoGuimaraes103: Organization of the code, I will remove the spaces

src/data-card-advanced.css.ts
```diff
@@ -0,0 +1,147 @@
...
+export const footerText = style([
+    {
+        width: 'auto',
+        height: 'auto',
+
+        display: 'flex',
+        flex: '1',
+
+        lineClamp: '2',
+        boxOrient: 'vertical',
```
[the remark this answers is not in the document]
BrunoGuimaraes103: In this part of the footerText css, it is made with the requirement logic of not allowing more than 2 lines and three points, so, among other development issues, it was configured that way.

src/data-card-advanced.css.ts
BrunoGuimaraes103: explanation in comment above

src/data-card-advanced.css.ts
```diff
@@ -0,0 +1,147 @@
...
+
+        alignItems: 'center',
+
+        minHeight: '40px',
+        maxHeight: '40px',
+
+        overflowY: 'hidden',
+        overflowWrap: 'break-word',
```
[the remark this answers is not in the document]
BrunoGuimaraes103: explanation in comment above

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,463 @@
...
+import type StackingGroup from './stacking-group';
+import type Image from './image';
+// import type {
+//     ValueBlock,
+//     InformationBlock,
+//     ProgressBlock,
+//     HighlightedValueBlock,
+//     RowBlock,
+//     SimpleBlock,
+// } from './advanced-slots';
```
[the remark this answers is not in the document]
BrunoGuimaraes103: No, these comments are due to be used in the future when the PR of the slots, which this component will use, go up

src/data-card-advanced.tsx
BrunoGuimaraes103: the aria label is necessary, since it was requested to already have a text in the aria label, in the accessibility part of figma this is documented

src/data-card-advanced.tsx
BrunoGuimaraes103: because unlike the normal data card, this one will not receive a children but an array of slots which is another PR for not having the possibility of adding other elements besides the slots

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,463 @@
...
+                <Boxed
+                    className={styles.boxed}
+                    dataAttributes={{'component-name': 'DataCard', ...dataAttributes}}
+                    ref={ref}
+                    width="100%"
+                    height="100%"
+                >
+                    <div className={styles.dataCard}>
+                        {cardOnPress && (
+                            <a
```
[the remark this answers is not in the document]
BrunoGuimaraes103: for accessibility reasons due to a conflict that previously occurred with Touchable

src/data-card-advanced.tsx
BrunoGuimaraes103: yes, i will remove the touchable-aria-label

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,463 @@
...
+                                tabIndex={0}
+                                aria-label={touchableAriaLabel}
+                                href="javascript:void(0)"
+                                className={styles.anchorCard}
+                                onClick={() => {
+                                    cardOnPress();
+                                }}
+                            />
```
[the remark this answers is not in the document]
BrunoGuimaraes103: could you explain this please?

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,463 @@
...
+                    >
+                        {buttonLink}
+                    </div>
+                )}
+            </div>
+        </div>
+    );
+};
+
+type slotsTypeof = typeof StackingGroup;
```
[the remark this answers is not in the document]
BrunoGuimaraes103: Due to the entry of PR slots, this type will receive other types in the future

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,463 @@
...
+            actions,
+            'touchable-aria-label': touchableAriaLabel,
+            'card-aria-label': cardAriaLabel,
+            onClose,
+        },
+        ref
+    ) => {
+        const finalActions = useTopActions(actions, onClose);
+        const hasAcations = finalActions?.length > 0;
+        const hasStackingGroup = !!stackingGroup;
```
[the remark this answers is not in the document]
BrunoGuimaraes103: this is checked if it has a stacking group, it serves to not need to use this boolean elsewhere, for the sake of reducing more unnecessary code

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,460 @@
...
+    pretitleLinesMax?: number;
+    pretitleAs?: TextAs;
+    title?: string;
+    titleLinesMax?: number;
+    titleAs?: TextAs;
+    subtitle?: string;
+    subtitleLinesMax?: number;
+    description?: string;
+    descriptionLinesMax?: number;
+    slots?: Array<RendersNullableElement<SlotsTypeof>>;
```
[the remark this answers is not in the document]
BrunoGuimaraes103: In the specifications the min-height is 216 px, so this extra size when empty is expected, as well as the paddingBottom.

The DataCard should only accept slot cards, which is another PR that hasn't been uploaded yet so they can't be React Node

Captura de tela 2023-06-19 111404

src/data-card-advanced.css.ts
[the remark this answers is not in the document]
wyunreal: https://developer.mozilla.org/en-US/docs/Web/CSS/box-orient
this feature seems to be deprecated/not recommended for prod sites

src/data-card-advanced.tsx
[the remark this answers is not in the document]
yceballost: and call it `extra` for feature parity with any Mística component

src/data-card-advanced.css.ts
```diff
@@ -0,0 +1,147 @@
...
+    {flexWrap: 'wrap'},
+]);
+
+export const boxed = style([
+    sprinkles({display: 'flex'}),
+    {
+        isolation: 'isolate', // Needed to preserve border-radius with Video component and Safari
+        selectors: {
+            '&:hover': {
+                backgroundColor: vars.colors.backgroundContainerHover,
```
yceballost, line 42: You are applying hover without onpress state, that's a bit weird
https://tinyurl.com/29g89hfx

Use hover only when the card has onPress interaction

src/community/__stories__/data-card-advanced-story.tsx
```diff
@@ -0,0 +1,120 @@
...
+    pretitle: string;
+    title: string;
+    subtitle: string;
+    description: string;
+    footerText: string;
+    showStackingGroup: boolean;
+    showFooterImage: boolean;
+    showHeadline: boolean;
+    showButtonPrimary: boolean;
+    showButtonLink: boolean;
```
yceballost, line 29: Please, follow similar solutions already implemented in other component stories
image

[Data Card Story](https://mistica-kbglt83lx-tuentisre.vercel.app/?path=/story/components-cards-data-card--default)

This way, it is easier to understand the possible actions that the card can take

src/community/__stories__/data-card-advanced-story.tsx
```diff
@@ -0,0 +1,120 @@
...
+                    footerText={footerText}
+                    onClose={() => window.alert('close!')}
+                />
+            </Box>
+        </ResponsiveLayout>
+    );
+};
+
+Default.storyName = 'Data Card Advanced';
+Default.args = {
```
yceballost, line 101: include controls for Dismiss and Top actions please

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,460 @@
...
+
+                    {hasFooterText && (
+                        <div style={{maxWidth}} className={styles.footerText}>
+                            <Text2 truncate={footerTextLinesMax} regular>
+                                {footerText}
+                            </Text2>
+                        </div>
+                    )}
+                </div>
+                {hasButtonLink && (
```
yceballost, line 265: buttonLink is not aligning when falling off line

image

yceballost (changes requested):

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,460 @@
...
+    pretitleAs?: TextAs;
+    title?: string;
+    titleLinesMax?: number;
+    titleAs?: TextAs;
+    subtitle?: string;
+    subtitleLinesMax?: number;
+    description?: string;
+    descriptionLinesMax?: number;
+    slots?: Array<RendersNullableElement<SlotsTypeof>>;
+    smallSlotSpace?: boolean;
```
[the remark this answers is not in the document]
BrunoGuimaraes103: well the prop smallSlotSpace, would serve to decrease the spacing of the bottom of the slots, it is found in figma at: Handoff -> Advanced Card Props -> bottom of the page

src/data-card-advanced.tsx
yceballost: visually misaligned, yes. The transparent background cause this problem. ButtonLink has a property called `aligned` to solve this problem. When the buttonLink falls, buttonLink has to use this prop.

https://tinyurl.com/29pldjsn

BrunoGuimaraes103: unfortunately I can't put this property directly, because the component only receives the buttonLink, what I could do is copy the behavior of this prop, but that would leave the component misaligned when hovering, unless it has its own animation to fix it back in place smoothly

BrunoGuimaraes103: Is it really necessary to correct this alignment? after I realized that the current DataCard also has this problem:

MicrosoftTeams-image (4)

BrunoGuimaraes103: Sorry for so many messages, I understood that the component will be used more in the mobile context so it won't have a hover, so your proposal to use the Aligned prop is quite valid, but since the data-card only receives the buttonLink, it will be the programmer who will have to put this prop, when passing the buttonLink.

yceballost: In this screenshot, the link is aligned
image

also buttonGroup has the logic to include this behavior
[example](https://mistica-web.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8ARAhgFzQYTQJygB0A7AAlIEtJiBeYE88pbCvMAGxlICM0wBrAOZ4IAV2JRsEdhDx1IMvAGcAdNzxoJAGQgB3AL6klFAF4w6AFgAM%2BgHwNGTAJLUAygAsIABwBKMQaLs%2BKQKsvLSsqrqmlCGAPT2ZIxIcSxsnInk%2Bg7uMGhQ7BTE5sBIACpogqQYAJ5e5oQgXiIAthCNtgASeQVFMCkVgrbZSRgUGJw0jWXjnI0OSqLcYxMNIK5LK3MgDrBKYHgUXmMQtI0oMPuHxxSn80kwAB4YGnRIAApBYDCe7LB4pASI3I3FEGAwpzoDiYACEwRDiO9Di18DUjCj2OxSKckZclHQABQASlINFspGA%2Bjs0MYAEEwCcyEooGglKRWVB2cyuZyOTyeRy0DSUnDwTjkajMqRgTx4actEV%2BG9RQiFcR%2BNjEXg8YSSWSKVTbGr%2BCK5cRjcMSAkSCAADQgDC5FqXBAAbRAAFkIAA3ChKLB4EAAXXtugoUEdSjdAGYAGxWIP6IA)

yceballost (changes requested):

atabel:
> Hello @atabel how are you? We have a problem in the pipe(audit-a11y) and we can not solve it, could you help us?

you have a message with a link to the issue: https://github.com/Telefonica/mistica-web/pull/780#issuecomment-1580960444

Please, take a look at the html report and fix the accessibility issue: https://devmisticaweb.blob.core.windows.net/mistica-web-1687291704713/community-datacardadvanced--default.html

BrunoGuimaraes103:
Well guys, sorry for all these pushes, it turns out that I'm trying to correct the audity error, but I can't do the tests locally because when I try to run the command 'yarn audit-accessibility' I end up giving this error in the image below, even when I try the command from 'yarn storybook-static' ends up not working because it seems to have a size limit, and I've also tried to run with 'yarn server-static-storybook', but it doesn't work and due to the lack of alternatives that I know of, I can only go testing by sending git pushes to this PR

yarn storybook-static:
Captura de tela 2023-06-21 094337

yarn audit-accessibility:
Captura de tela 2023-06-21 140656

src/community/__stories__/data-card-advanced-story.tsx
```diff
@@ -0,0 +1,139 @@
...
+Default.argTypes = {
+    headlineType: {
+        options: ['promo', 'active', 'inactive', 'success', 'warning', 'error'],
+        control: {type: 'select'},
+    },
+    actions: {
+        options: ['none', 'stackingGroup', 'onClose', 'stackingGroup and onClose'],
+        control: {type: 'select'},
+    },
+    bottomActions: {
```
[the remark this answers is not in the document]
yceballost: missing onpress + buttons in story

(this is a common usage in Vivo novo app, I think is good to include this possibility that differs between this card and mística core cards (that is forbiden for mistica core cards)

for example 
https://tinyurl.com/2ox7g2pw

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,468 @@
...
+
+    const flexDirection = hasAllItens ? 'column' : 'row';
+    const alignItems = hasAllItens ? 'start' : 'center';
+    const marginTop = hasAllItens ? '8px' : '16px';
+    const maxWidth = hasButtonLink && !hasAllItens ? '178px' : '';
+
+    return (
+        <div>
+            <NegativeBox>
+                <Divider />
```
yceballost, line 239: this divider is not working properly in desktop views
image


the divider has to fill the width

playroom/snippets.tsx
```diff
@@ -1858,6 +1858,21 @@ const stackingGroupSnippets = [
     },
 ];
 
+const dataCardAdvancedSnippets = [
+    {
+        name: 'Data Advanced Card',
+        code: `
+          <DataCardAdvanced smallSlotSpace={true} 
+            title="title" subtitle={"subtitle 1"}
+            subtitle="subtitle"
```
yceballost, line 1867: `subtitle` prop is duplicated

src/community/__stories__/data-card-advanced-story.tsx
yceballost, line 1: I found this weird behavior. When the image is hide, the height increasing. I don't understand why this is happening

I think it does not seem to be a major problem, I just making visible 
https://github.com/Telefonica/mistica-web/assets/6722153/dc513780-cb4b-470b-a874-f2d7abced04c

src/data-card-advanced.tsx
yceballost: Related to this, when the link is on the right, seems to be not aligned as same as specs 
image

Maybe is missing `aligned` prop in this case too

yceballost (changes requested):

src/community/__stories__/data-card-advanced-story.tsx
BrunoGuimaraes103: it's quite strange, I'm not really sure what caused it but after the modifications at least to me it seems normal

https://github.com/Telefonica/mistica-web/assets/123959276/07a914ff-f290-4718-89e5-a2bd464a3a93

BrunoGuimaraes103: I noticed that the button link was misaligned with the primary button during the test of this video that I sent (size desktop), this last update already corrected this by putting the same margin top as the primary button

src/data-card-advanced.tsx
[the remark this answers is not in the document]
yceballost: Missing rename "smallSlotSpace" to "smallExtraSpace" right?

src/community/__stories__/data-card-advanced-story.tsx
yceballost: nice 🤩 thanks!

src/community/__stories__/data-card-advanced-story.tsx
```diff
@@ -0,0 +1,137 @@
...
+import Image from '../../image';
+import StackingGroup from '../../stacking-group';
+
+import type {TagType} from '../../tag';
+
+export default {
+    title: 'Community/DataCardAdvanced',
+};
+
+type Args = {
```
yceballost, line 17: I think this is my last comment in this review, I promise. 

I think would be nice to add a boolean control to show "extra" like in data card story.

You can include inside extra a <Placeholder /> component.
https://mistica-web.vercel.app/?path=/story/components-cards-data-card--default
image

yceballost (approved):

src/community/__stories__/data-card-advanced-story.tsx
BrunoGuimaraes103: kk, understood and well Unfortunately I think this may take a while, it turns out that the data-card-advanced will only be able to receive the PR components of the slots/blocks, until it has the merge of them in the master it will not be able to make the change, but as soon as they go up I will make this change :)

yceballost: I don't understand the relationship with that PR, just include a control to show or hide "extra" prop. You can include inside `extra` something like this https://tinyurl.com/2o2ewpkl, (like the rest of story component with extra prop)

Extra prop is just a area to include any thing, you can include Blocks component inside or just any atomic composition

BrunoGuimaraes103: It's just that the data-card-advanced was thought to only allow specific slots that are the PR of the advanced-slot, because of that, however much it is accepted in the playroom, for the storybook it is a different case, when trying this way that it passed in the storybook the page hangs and gives no response:

Captura de tela 2023-06-23 170427

yceballost: oh! I didn't know that! Is there a rationale for this? I don't think we should have that limitation on the DataCardAdvanced extra, the extra should be able to support any component.

@atabel are you agree?

BrunoGuimaraes103: Well the data-card-advanced was thought to use only the slots/blocks from the PR of advacend-slots, unfortunately I don't remember exactly why it was thought to be like that, so these two PR's are a set so to speak

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,498 @@
...
+                                        <CardActionsGroup actions={finalActions} />
+                                    </div>
+                                )}
+                            </div>
+                            <div style={{marginTop: 'auto', width: '100%'}}>
+                                {extra && extra?.length ? (
+                                    <Box paddingTop={16} paddingBottom={24}>
+                                        {extra.map((ex, index) => {
+                                            return (
+                                                <div>
```
kydorn, line 463: i think you need a 'key' prop here

src/data-card-advanced.tsx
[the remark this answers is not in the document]
atabel: the link should have a content, the content of the link should be the clickable area of the card

src/data-card-advanced.tsx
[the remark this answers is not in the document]
atabel: you don't need that. Just do `{stackingGroup && <div`

src/__screenshot_tests__/__image_snapshots__/data-card-advanced-screenshot-test-tsx-data-card-advanced-8-snap.png
atabel, line 1: Is this a valid case? looks weird to me

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,498 @@
...
+            onClose,
+        },
+        ref
+    ) => {
+        const finalActions = useTopActions(actions, onClose);
+        const hasAcations = finalActions?.length > 0;
+        const hasStackingGroup = !!stackingGroup;
+
+        const hasFooter = !!button || !!footerImage || !!footerText || !!buttonLink;
+        const footerProps = {button, footerImage, footerText, footerTextLinesMax, buttonLink};
```
atabel, line 385: Why don't you just inline this when you call `<CardFooter` ? no need to create a new variable. The code is more difficult to read when the variables aren't declared near to where they are used, when I read `<CardFooter {...footerProps}/>` I need to scroll the whole screen to go to this object definition.

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,498 @@
...
+            <div
+                className={styles.actions}
+                style={{flexDirection, alignItems, marginTop: marginTopTabletOrBigger}}
+            >
+                {hasButton && (
+                    // eslint-disable-next-line jsx-a11y/no-static-element-interactions
+                    <div
+                        aria-label="button"
+                        tabIndex={-1}
+                        onClick={(event) => event.stopPropagation()}
```
atabel, line 250: I guess this is a hack to prevent that clicking on the button would trigger the card onPress isn't it?
Instead of doing this, try to avoid nesting clickable areas

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,498 @@
...
+                                                        }}
+                                                    >
+                                                        <div style={{zIndex: '0', width: '100%'}}>{ex}</div>
+                                                    </div>
+
+                                                    {index + 1 !== extra.length && (
+                                                        <div
+                                                            className={sprinkles({
+                                                                paddingY: extraSpaceSize,
+                                                            })}
```
atabel, line 477: why don't use a Box instead?

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,498 @@
...
+                        display: 'flex',
+                        flexDirection: 'row',
+                        marginTop,
+                        flexWrap: 'wrap',
+                        marginRight,
+                    }}
+                >
+                    {hasFooterImage && (
+                        <div
+                            style={{marginRight: '16px', zIndex: '0'}}
```
atabel, line 275: why zindex 0?

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,498 @@
...
+                        aria-label="link"
+                        tabIndex={-1}
+                        className={sprinkles({
+                            display: 'flex',
+                        })}
+                        onClick={(event) => event.stopPropagation()}
+                        style={{
+                            marginTop: isMobile ? marginTop : marginTopButton,
+                            position: 'relative',
+                            zIndex: '2',
```
atabel, line 302: you have too many zindex. can you explain why?

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,498 @@
...
+    pretitleLinesMax?: number;
+    pretitleAs?: TextAs;
+    title?: string;
+    titleLinesMax?: number;
+    titleAs?: TextAs;
+    subtitle?: string;
+    subtitleLinesMax?: number;
+    description?: string;
+    descriptionLinesMax?: number;
+    extra?: Array<RendersNullableElement<ExtraTypeof>>;
```
atabel, line 333: extra shouldnt be an array, just a `React.ReactNode` like we do in the rest of Mistica components. Take into account that you can pass any react construction there, even a list of elements inside a fragment if you need, but there is no point of supporting arrays here.

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,498 @@
...
+};
+
+const CardFooter: React.FC<CardFooterProps> = ({
+    button,
+    footerImage,
+    footerText,
+    footerTextLinesMax,
+    buttonLink,
+}) => {
+    const {isMobile, isTabletOrBigger} = useScreenSize();
```
atabel, line 217: try to use mediaqueries in css.ts instead of this hook. Mediaqueries are better for SSR support

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,498 @@
...
+    const marginRight = footerCondition ? '' : 'auto';
+    const alignItems = footerCondition ? 'start' : 'center';
+    const marginTop = hasAllItens ? '8px' : '16px';
+    const marginTopTabletOrBigger = isTabletOrBigger ? '8px' : '';
+    const marginTopButton = isMobile ? '16px' : '8px';
+    const maxWidth = hasButtonLink && !hasAllItens ? '178px' : '';
+
+    return (
+        <div>
+            <div style={{marginLeft: -24, marginRight: -24}}>
```
atabel, line 237: use className instead of inline styles when possible

atabel (changes requested):

src/index.tsx
```diff
@@ -10,6 +10,8 @@ export {default as Touchable} from './touchable';
 export type {TouchableElement} from './touchable';
 export {default as Spinner} from './spinner';
 export {default as FadeIn} from './fade-in';
+export {default as DataCardAdvanced} from './data-card-advanced';
```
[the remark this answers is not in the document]
BrunoGuimaraes103: would that include the test-screenshot?

src/data-card-advanced.tsx
BrunoGuimaraes103: It was resolved this way because the CMS staff was having difficulty with the original Card since they could not control what would be added to the card's extra property, with that the Advanced Data Card was developed that would have this control with the blocks, in this way it was It was discussed that an array of blocks would be allowed so that the CMS could build the component, when this component was idealized it was agreed with someone from Novum, where we don't remember names at the moment, that it would be an array of blocks to help everyone and maintain a component control.

src/data-card-advanced.tsx
```diff
@@ -0,0 +1,463 @@
...
+    iconBackground?: string;
+    iconBackgroundInverse?: string;
+};
+
+type CardActionsGroupProps = {
+    actions: Array<CardAction>;
+    isInverse?: boolean;
+};
+
+const CardActionsGroup = ({actions, isInverse}: CardActionsGroupProps): JSX.Element => {
```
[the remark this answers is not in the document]
BrunoGuimaraes103: We are re-using cardActions from card.tsx, but changes have been made to suit our needs.

src/data-card-advanced.tsx
BrunoGuimaraes103: It was requested that there be the possibility of changing the pretitle and title, to h1 to h4, while keeping the default values ​​like p to pretitle and h3 to title, that's why the purpose of these props As.

src/data-card-advanced.tsx
BrunoGuimaraes103: could you please elaborate?

src/data-card-advanced.tsx
BrunoGuimaraes103: turns out the whole card is clickable

src/data-card-advanced.tsx
BrunoGuimaraes103: The image component, inside the touchable doesn't become clickable, but with zindex: 0, it becomes clickable

src/data-card-advanced.tsx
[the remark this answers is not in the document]
BrunoGuimaraes103: The reason for the conflict is to get the buttons working inside the touchable.

in the case where you have:
// eslint-disable-next-line jsx-a11y/no-static-element-interactions

The Div wrapped around the buttons serves to prevent the card's onClick event from being triggered when the buttons event is, because of this onClick, the div did not accept focusable (such as buttons, inputs, etc.), and this generated conflict.

without this eslint-disable, the error appears in the code of:
Static HTML elements with event handlers require a role.eslintjsx-a11y/no-static-element-interactions

however, when using a role, aria-hidden, etc, it still triggered the audit-a11y accessibility error in PR

I just managed to 'resolve' this error by using this eslint-disable because there's no way to remove the div around the button, and there's no way to put the button outside of it, or by an aria-hidden, role, aria-disable, etc.

src/data-card-advanced.tsx
BrunoGuimaraes103: Before the zIndex was used to solve problems with clickable areas in the data-card, but as it is no longer a problem, I removed them except one, which is for the top actions

src/data-card-advanced.tsx
BrunoGuimaraes103: the hasFooter const is used below the cardContentStyle, what I did was move the hasFooter above it, since the footerProps is giving an error when I try to put its content directly in the CardFooter

src/data-card-advanced.tsx
BrunoGuimaraes103: The isTabletOrBigger can be changed by putting it in the css, but not the isMobile, because it is used as a validator

src/__screenshot_tests__/__image_snapshots__/data-card-advanced-screenshot-test-tsx-data-card-advanced-8-snap.png
BrunoGuimaraes103: Yes, it is a valid case that has been passed

src/data-card-advanced.tsx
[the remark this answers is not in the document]
BrunoGuimaraes103: this part no longer exists in the code, the code has been changed and no longer has this requirement

src/index.tsx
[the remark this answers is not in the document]
atabel: yes

src/data-card-advanced.tsx
atabel: I don't see any technical reason for this to be an Array. If the CMS sends an array to the frontend, the frontend can use a afragment and add there all the items in the array, I don't see the problem.

src/data-card-advanced.tsx
atabel: you can do the same with css classes and mediaqueries. Add a class when `hasAllItems` is true, then add some mediaqueries to that class.

src/data-card-advanced.tsx
atabel: That's an accesibility issue and the CI a11y checks are warning you. Whe can't have nested clickable areas. You should divide the card in 2 zones: the clickable area (onPress) and the buttons area

BrunoGuimaraes103: it turns out that the whole card is clickable, including the area around the buttons

src/data-card-advanced.tsx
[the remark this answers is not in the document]
BrunoGuimaraes103: Cannot use CardActionsGroup from card.tsx is not exported.

src/data-card-advanced.tsx
BrunoGuimaraes103: The reason why the array is needed is due to two factors, namely the specific styling of the organization of the elements which if set to just React.node will cause an error, while there is also the issue that the advanced-data-card can only accept these blocks advanced-slot PR specifics plus StackingGroup

https://github.com/Telefonica/mistica-web/pull/770

src/community/advanced-data-card.css.ts
```diff
@@ -0,0 +1,184 @@
...
+import * as mq from '../media-queries.css';
+import {vars} from '../skins/skin-contract.css';
+import {applyAlpha} from '../utils/color';
+
+export const actions = style([
+    sprinkles({
+        display: 'flex',
+        flex: 1,
+        justifyContent: 'space-between',
+        paddingX: 16,
```
yceballost, line 12: paddingX in desktop views has to be setup in 24px to avoid this misalignment

image

yceballost (changes requested):

wyunreal (approved):

src/index.tsx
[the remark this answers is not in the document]
BrunoGuimaraes103: unfortunately trying to move the test-screenshot to the community folder did not work, I think I would have to create a new folder for it or some configuration, because of that I left it like this

src/community/advanced-data-card.tsx
```diff
@@ -0,0 +1,471 @@
...
+                                        color={vars.colors.textPrimary}
+                                        truncate={pretitleLinesMax}
+                                        as={pretitleAs}
+                                        regular
+                                        hyphens="auto"
+                                    >
+                                        {pretitle}
+                                    </Text2>
+                                )}
+                                <Text
```
yceballost, line 138: please, use text4 for title as defines specs
image

yceballost (changes requested):

yceballost (approved):

src/index.tsx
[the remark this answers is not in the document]
atabel: as I told you in the Blocks PR, you need to create the folder there, yes

src/community/advanced-data-card.css.ts
```diff
@@ -0,0 +1,196 @@
...
+]);
+
+export const boxed = style([
+    sprinkles({display: 'flex'}),
+    {
+        isolation: 'isolate', // Needed to preserve border-radius with Video component and Safari
+    },
+]);
+
+export const interaction = style([
```
atabel, line 46: you don't need to use an array if you only have one style

src/community/advanced-data-card.css.ts
```diff
@@ -0,0 +1,196 @@
...
+    {
+        selectors: {
+            '&:hover': {
+                backgroundColor: vars.colors.backgroundContainerHover,
+                transition: '0.15s ease-in-out',
+            },
```
atabel, line 52: this should be inside a `@media` `[mq.supportsHover]` (see other cards)

src/community/advanced-data-card.css.ts
```diff
@@ -0,0 +1,196 @@
...
+        '@media': {
+            [mq.desktopOrBigger]: {
+                paddingTop: 32,
+                paddingBottom: 32,
+            },
+        },
+    },
+]);
+
+export const minHeight = style([{minHeight: 216}]);
```
atabel, line 78: don't use array if you only have one style
```suggestion
export const minHeight = style({minHeight: 216});
```

src/community/advanced-data-card.css.ts
```diff
@@ -0,0 +1,196 @@
...
+    sprinkles({
+        display: 'flex',
+        flex: 1,
+        flexDirection: 'column',
+        width: '100%',
+        height: '100%',
+    }),
+    {
+        minHeight,
+    },
```
atabel, line 90: this is wrong, minHeight should be a number, not the result of a `style()` call. If you want to reuse the `minHeight` style here you can do this:
```suggestion
    minHeight,
```

src/community/advanced-data-card.css.ts
```diff
@@ -0,0 +1,196 @@
...
+        marginRight: -24,
+    },
+]);
+
+export const footerText = style([
+    {
+        display: 'flex',
+        flex: '1',
+
+        alignItems: 'center',
```
atabel, line 152: use sprinkles

src/community/advanced-data-card.tsx
```diff
@@ -0,0 +1,462 @@
...
+
+            dataAttributes,
+            actions,
+            'aria-label': ariaLabel,
+            onClose,
+        },
+        ref
+    ) => {
+        const finalActions = useTopActions(actions, onClose);
+        const hasAcations = finalActions?.length > 0;
```
atabel, line 338: typo. `hasActions`

src/community/advanced-data-card.tsx
```diff
@@ -0,0 +1,462 @@
...
+    iconBackground?: string;
+    iconBackgroundInverse?: string;
+};
+
+type CardActionsGroupProps = {
+    actions: Array<CardAction>;
+    isInverse?: boolean;
+};
+
+const CardActionsGroup = ({actions, isInverse}: CardActionsGroupProps): JSX.Element => {
```
atabel, line 39: Please, remove this component and use the one in `card.tsx` instead. You can make it public in `cad.tsx` and import it here.

src/community/advanced-data-card.tsx
```diff
@@ -0,0 +1,462 @@
...
+                                                >
+                                                    <div
+                                                        className={sprinkles({
+                                                            width: '100%',
+                                                        })}
+                                                    >
+                                                        {stackingGroup}
+                                                    </div>
+                                                </div>
+                                            )}
```
atabel, line 397: are these divs really needed? isn't this enough?
```suggestion
                                            {stackingGroup}
```

src/community/advanced-data-card.tsx
```diff
@@ -0,0 +1,462 @@
...
+                                                titleLinesMax={titleLinesMax}
+                                                subtitle={subtitle}
+                                                subtitleLinesMax={subtitleLinesMax}
+                                                description={description}
+                                                descriptionLinesMax={descriptionLinesMax}
+                                            />
+                                        </Stack>
+                                    </Box>
+                                    {hasAcations && (
+                                        <div style={topActionsStylesWithIcon}>
```
atabel, line 414: use classes

src/community/advanced-data-card.tsx
```diff
@@ -0,0 +1,462 @@
...
+                                            <CardActionsGroup actions={finalActions} />
+                                        </div>
+                                    )}
+                                </div>
+                                <div style={{marginTop: 'auto', width: '100%'}}>
```
atabel, line 419: use classes

src/community/advanced-data-card.tsx
```diff
@@ -0,0 +1,462 @@
...
+    pretitleAs?: TextAs;
+    title?: string;
+    titleLinesMax?: number;
+    titleAs?: TextAs;
+    subtitle?: string;
+    subtitleLinesMax?: number;
+    description?: string;
+    descriptionLinesMax?: number;
+    extra?: Array<RendersNullableElement<ExtraTypeof>>;
+    smallSlotSpace?: boolean;
```
atabel, line 291: I'd rename this:
```suggestion
    extraDividerPadding?: 8 | 24;
```
With 24 as default

src/community/advanced-data-card.tsx
```diff
@@ -0,0 +1,462 @@
...
+                    >
+                        {buttonLink}
+                    </div>
+                )}
+            </div>
+        </>
+    );
+};
+
+type ExtraTypeof = typeof StackingGroup;
```
atabel, line 272: I'd rename to
```suggestion
type AllowedExtra = typeof StackingGroup;
```

src/community/advanced-data-card.css.ts
```diff
@@ -0,0 +1,196 @@
...
+            [mq.supportsHover]: {
+                ':hover': {
+                    background: applyAlpha(vars.rawColors.backgroundContainer, 0.9),
+                },
+            },
+        },
+    },
+]);
+
+export const adjustmentDivider = style([
```
atabel, line 140: just `divider`

src/community/advanced-data-card.css.ts
```diff
@@ -0,0 +1,196 @@
...
+    {
+        marginLeft: -24,
```
atabel, line 142: I'd extract this 24 to a variable and reuse it for the horizontal padding

src/community/advanced-data-card.tsx
```diff
@@ -0,0 +1,462 @@
...
+                        className={classNames(
+                            sprinkles({
+                                display: 'flex',
+                            }),
+                            styles.marginTopButton
+                        )}
+                        style={{
+                            marginRight: 16,
+                            position: 'relative',
+                        }}
```
atabel, line 221: use classes
why position relative?

src/community/advanced-data-card.tsx
```diff
@@ -0,0 +1,462 @@
...
+            >
+                {hasButton && (
+                    <div
+                        tabIndex={-1}
+                        className={classNames(
+                            sprinkles({
+                                display: 'flex',
+                            }),
+                            styles.marginTopButton
+                        )}
```
atabel, line 217: instead of merging `sprinkles` + `styles.marginTopButton` here, I'd move the sprinkes styles to the css.ts file and rename `marginTopButton` to just `button`

src/community/advanced-data-card.tsx
```diff
@@ -0,0 +1,462 @@
...
+                    </div>
+                )}
+                <div
+                    className={hasAllItens ? styles.marginRightAuto : ''}
+                    style={{
+                        display: 'flex',
+                        flexDirection: 'row',
+                        marginTop: hasAllItens ? 8 : 16,
+                        flexWrap: 'wrap',
+                    }}
```
atabel, line 233: use classes

atabel (changes requested):

src/community/advanced-data-card.tsx
BrunoGuimaraes103: As the position relative, didn't change the design once I removed it, it looks like it's a leftover of some old logic from this component

src/community/advanced-data-card.tsx
BrunoGuimaraes103: the reason for separating marginTop and marginTopButton is to simplify the logic of the size of the top when it has or not all the elements in the footer, but to separate as much as possible the use of sprinkles with the styles, I made the marginTopButton have the elements repeated in the marginTop and the 'display:flex' which is used wherever it is used.

image

src/community/advanced-data-card.tsx
```diff
@@ -0,0 +1,462 @@
...
+
+            <div
+                className={classNames(
+                    styles.actions,
+                    styles.actionsVariants[hasAllItens ? 'mobile' : 'default']
+                )}
+            >
+                {hasButton && (
+                    <div
+                        tabIndex={-1}
```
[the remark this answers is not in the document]
BrunoGuimaraes103: before it was necessary when there was a touchable involving this part of the footer, but since it is no longer necessary, I will remove the tabIndex

atabel (approved):

kydorn (approved):

kydorn:
Please, add stories and screenshot tests with data card advanced and all types of allowed blocks. Thanks!

BrunoGuimaraes103:
@atabel @kydorn 
There is a problem that prevents the inclusion of Blocks in the Storybook and consequently test-screenshot, in the Advanced-Data-Card.

The problem is that when trying to pass an array in the extra prop, the storybook crashes (with the only exception being an array of nulls), we find that the problem boils down to not the component itself, but the version of the storybook

This link is from a case of the same problem in question:
https://github.com/storybookjs/storybook/issues/17482

If possible, could we update the storybook version to the latest?

playroom/snippets.tsx
```diff
@@ -1862,17 +1862,173 @@ const advancedDataCardSnippets = [
...
-            button={<ButtonPrimary small onPress={() => {}}>Action</ButtonPrimary>}
-          />
-  `,
+        <AdvancedDataCard
+          headline={
+            <Tag type="promo" Icon={IconStarFilled}>
+              Promo
+            </Tag>
+          }
+          smallSlotSpace={true}
```
atabel, line 1871: this isn't a valid prop for AdvancedDataCard

playroom/snippets.tsx
```diff
@@ -1862,17 +1862,173 @@ const advancedDataCardSnippets = [
...
     },
 ];
 
+const RowBlockSnippets = [
+    {
+        name: 'Row Block',
+        code: `
+          <RowBlock title="title" description="description" />
+      `,
+        group: 'RowBlock',
```
atabel, line 1950: I'd use the same group name for all the block snippets

atabel (approved):

PR #782, opened 2023-06-07
fix(FixedFooterLayout): avoid blinking during page transitions

marcoskolodny:
Fixing these two issues, caused by the same bug:
- https://jira.tid.es/browse/WEB-801
- https://jira.tid.es/browse/WEB-1409


pladaria (approved):

marcoskolodny:
Fix working correctly in Iphone 13 Pro (iOS 16.5) and in Xiaomi 12T (Android 13)

PR #783, opened 2023-06-14
fix(FixedFooterLayout): remove next js errors when using SSR

marcoskolodny:


src/hooks.tsx
```diff
@@ -173,7 +173,9 @@ export const useScreenHeight = (): number => {
     return screenHeight;
 };
 
-export const useIsWithinIFrame = (): boolean => window?.top !== window?.self;
+export const useIsWithinIFrame = (): boolean => {
+    return typeof window !== 'undefined' && window.top !== window.self;
```
atabel, line 177: why don't use `!isServerSide()` here too?

marcoskolodny: fixed

src/utils/environment.tsx
```diff
@@ -1 +1,3 @@
 export const isServerSide = (): boolean => typeof window === 'undefined';
+
+export const isClientSide = (): boolean => typeof window !== 'undefined';
```
Marcosld, line 3: matter of taste: export const isClientSide = (): boolean => !isServerSide();

Marcosld (approved):

PR #787, opened 2023-06-15
fix(Cards): aspect ratio support for old browsers

atabel:
https://jira.tid.es/browse/WEB-1429


src/card.css.ts
```diff
@@ -341,3 +341,27 @@ export const videoAction = style([
...
 ]);
+
+const aspectRatio = createVar();
+
+export const vars = {aspectRatio};
+
+export const cardContainer = style({
+    position: 'relative',
+    aspectRatio,
+    '@supports': {
```
imbysho, line 352: Maybe I would check if the browser supports the aspect ratio using JS, like **CSS.supports()**. Because it has better coverage on Safari. "@supports" just works from Safari 14 and "CSS.supports" goes until Safari 9.

CSS.supports() (https://caniuse.com/?search=CSS.supports)
@supports (https://caniuse.com/?search=@supports)

atabel: hmmm... no, you're wrong, @supports in css is supported in safari 9 https://caniuse.com/css-featurequeries

pladaria (approved):

marcoskolodny (approved):

imbysho (approved):

PR #788, opened 2023-06-15
fix(FixedFooterLayout): fix warnings in SSR

marcoskolodny:


Marcosld (approved):

src/fixed-footer-layout.tsx
atabel, line 1: I miss a SSR test detecting this

src/fixed-footer-layout.tsx
```diff
@@ -106,17 +105,17 @@ const FixedFooterLayout: React.FC<Props> = ({
         };
     }, [hasContentEnoughVSpace, platformOverrides]);
 
-    const isContentWithScroll = isClientSide() && hasScroll(getScrollableParentElement(containerRef.current));
-
-    const isFixedFooter = hasContentEnoughVSpace || !isContentWithScroll;
+    const isFixedFooter = hasContentEnoughVSpace;
```
marcoskolodny, line 108: After debugging in different devices (mobile and web) and testing several scenarios with @brtbrt, we concluded that it's useless to check for the condition `!isContentWithScroll`. This was old logic, and it makes no sense to use it in the current code.

atabel (approved):

src/fixed-footer-layout.tsx
brtbrt: specifically, it come from the old JSS styling 👍

brtbrt (approved):

PR #795, opened 2023-06-20
feat(ThemeContextProvider): built in Link components for Next and React Router

pladaria:


examples/next/pages/index.tsx
```diff
@@ -55,20 +55,20 @@ const Home = (): JSX.Element => {
                     <Box>
                         <Title1>Using links</Title1>
                         <Stack space={16}>
-                            <Text>
```
pladaria, line 58: for some reason, using the Text component produced hydration warnings

marcoskolodny: This is something I noticed while working on fixing the fixed footer SSR errors. There is an issue already: https://jira.tid.es/browse/WEB-1430

src/theme.tsx
```diff
@@ -196,12 +196,64 @@ type LinkComponent = React.ComponentType<{
...
+const AnchorLink: LinkComponent = ({to, innerRef, ...props}) => (
     <a ref={innerRef} href={typeof to === 'string' ? to : to?.pathname} {...props}>
         {props.children}
     </a>
 );
 
+const getReactRouter5Link =
+    (ReactRouterLink: React.ComponentType<any>): LinkComponent =>
+    ({...props}) =>
+        <ReactRouterLink {...props} />;
```
atabel, line 208: isn't this equivalent?
```suggestion
    (ReactRouterLink: React.ComponentType<any>): LinkComponent => ReactRouterLink
```

src/theme.tsx
```diff
@@ -196,12 +196,64 @@ type LinkComponent = React.ComponentType<{
...
+            // nt_core
+            return getReactRouter6Link(Link.Component);
+        case 'Next12':
+            // zeus-web
+            return getNext12Link(Link.Component);
+        case 'Next13':
+            // hello-world-web, global-checkout-webview
+            return getNext13Link(Link.Component);
+        default:
+            return AnchorLink;
```
atabel, line 253: You could make this switch exhaustive
```suggestion
    if (!Link) {
        return AnchorLink;
    }
    switch (Link.type) {
        case 'ReactRouter5':
            // webapp, flow-frontend
            return getReactRouter5Link(Link.Component);
        case 'ReactRouter6':
            // nt_core
            return getReactRouter6Link(Link.Component);
        case 'Next12':
            // zeus-web
            return getNext12Link(Link.Component);
        case 'Next13':
            // hello-world-web, global-checkout-webview
            return getNext13Link(Link.Component);
        default:
            const exhaustiveCheck: never = Link.type;
            throw new Error(`Invalid Link type: ${exhaustiveCheck}`);
```

src/theme-context-provider.tsx
```diff
@@ -111,7 +111,7 @@ const ThemeContextProvider: React.FC<Props> = ({theme, children, as}) => {
                 title1: {...defaultTextPresetsConfig.title1, ...theme.skin.textPresets?.title1},
                 indicator: {...defaultTextPresetsConfig.indicator, ...theme.skin.textPresets?.indicator},
             },
-            Link: theme.Link ?? AnchorLink,
+            Link: getMisticaLinkComponent(theme.Link),
```
atabel, line 114: This could cause potential remount issues because we are creating new Link components at render time.
It won't be an issue if the `theme` object reference is stable becase we have an useMemo here, but if someone does something like `<ThemeContextProvider theme={{/*define the theme inline here*/}}>` they'll have problems.
(I'm seeing we already have a warning for this in the README)

pladaria: This is somewhat out of scope of this task but perhaps a solution to this problem would be to store the theme inside a ref and don't react to changes. If someone wants a dynamic provider, they can force remounts with a key update. WDYT?

pladaria: Commented in teams, leaving this as is for now

atabel (approved):

marcoskolodny (approved):

PR #796, opened 2023-06-21
fix(Select): Fix select options positioning inside non static parents

pladaria:
Fixes this: https://t.ly/Yb_W (open select)


src/select.tsx
```diff
@@ -106,12 +106,19 @@ const Select: React.FC<SelectProps> = ({
...
-                    height,
-                } = fieldRef.current.getBoundingClientRect();
+
+                // get a possible non-static parent, like a modal dialog
+                const nonStaticParentElement = focusableRef.current?.offsetParent;
+                const parentElementRect = nonStaticParentElement?.getBoundingClientRect() || {
+                    left: 0,
+                    top: 0,
+                };
+                const clientRect = fieldRef.current.getBoundingClientRect();
```
marcoskolodny, line 116: Calling this function may cause errors if using server side rendering, right?

pladaria: yes, but this is only executed when interacting with the element

atabel (approved):

marcoskolodny (approved):

yceballost (approved):
perfectou!

PR #798, opened 2023-06-21
fix(Switch, RadioButton, Checkbox): stop propagation on click

marcoskolodny:


src/checkbox.tsx
```diff
@@ -123,7 +123,13 @@ const Checkbox: React.FC<RenderProps | ChildrenProps> = (props) => {
             role="checkbox"
             aria-checked={value ?? checkedState}
             onKeyDown={disabled ? undefined : handleKeyDown}
-            onClick={disabled ? undefined : handleChange}
+            onClick={(e) => {
+                if (!disabled) {
+                    e.preventDefault();
```
pladaria, line 128: this preventDefault shouldn't be needed. There's no default handler in divs

src/radio-button.tsx
```diff
@@ -117,7 +117,13 @@ const RadioButton: React.FC<PropsRender | PropsChildren> = ({
             aria-checked={checked}
             aria-disabled={disabled}
             aria-labelledby={labelId}
-            onClick={disabled ? undefined : () => select(value)}
+            onClick={(e) => {
+                if (!disabled) {
+                    e.preventDefault();
```
pladaria, line 122: same here

src/checkbox.tsx
pladaria: I'm unsure whether we should use the stopPropagation method on disabled fields or not. I would stopPropagation unconditionally. What's your opinion?

marcoskolodny: It can be an option. I'm not sure what would be the best option from the user experience side. I will ask design what are their thoughts about this.

src/__tests__/checkbox-test.tsx
```diff
@@ -131,3 +131,18 @@ test("with link. Clicking the link doesn't check the checkbox", async () => {
...
+        <ThemeContextProvider theme={makeTheme()}>
+            <button onClick={onPressHandler}>
+                <Checkbox name="checkbox" />
+            </button>
+        </ThemeContextProvider>
+    );
+
+    await userEvent.click(screen.getByRole('checkbox'));
+
+    await waitFor(() => expect(onPressHandler).not.toHaveBeenCalled());
```
atabel, line 147: what's the purpose of the `waitFor` here? the expect will work on the first try, and if it fails it will fail forever. So I'd remove the `waitFor`
```suggestion
    expect(onPressHandler).not.toHaveBeenCalled();
```

src/checkbox.tsx
atabel: hmmm... why would you ask design? This is just an implementation detail, isn't it?
I wouldn't add the handler if disabled

src/checkbox.tsx
```diff
@@ -105,7 +105,7 @@ const Checkbox: React.FC<RenderProps | ChildrenProps> = (props) => {
         }
     };
 
-    const handleKeyDown = (event: React.KeyboardEvent<HTMLDivElement>) => {
+    const handleKeyDown = (event: React.KeyboardEvent) => {
```
atabel, line 108: why did you remove the generic?

atabel (commented):
Please, review the behavior in Rows with onPress + switch/checkbox/radio. I think you have broken the behavior. Also I'd add some tests for those cases

src/checkbox.tsx
marcoskolodny: Because it's related to user experience. Design agrees with Pedro about not propagating the event when it's disabled

atabel: ok for me, then

marcoskolodny:
> Please, review the behavior in Rows with onPress + switch/checkbox/radio. I think you have broken the behavior. Also I'd add some tests for those cases

fixed

atabel (approved):

pladaria (approved):
nice, thanks!

PR #805, opened 2023-06-29
chore(eslint): upgrade eslint-config and fix errors

pladaria:


marcoskolodny (approved):

packages/import-mistica-icons/index.js
```diff
@@ -1,4 +1,3 @@
-// @ts-check
```
atabel, line 1: another option could be renaming this file to `.ts` and execute it with `ts-node`

atabel (approved):

PR #812, opened 2023-07-06
fix(Row): fix missing hover in lists with href

marcoskolodny:
Issue: https://jira.tid.es/browse/WEB-747

Cause of the problem: https://stackoverflow.com/a/40075917

### DIFF
--- a/src/list.css.ts
+++ b/src/list.css.ts
@@ -25,6 +25,7 @@ export const rowContent = sprinkles({
     background: 'transparent',
     padding: 0,
     display: 'block',
+    height: '100%',
 });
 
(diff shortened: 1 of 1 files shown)


yceballost (approved):
I found a very corner case, the layer not cover the entire height of the row, this is not important in my opinion.
image

https://tinyurl.com/2d94wol3

marcoskolodny:
> I found a very corner case, the layer not cover the height of the row, this is not important in my opinion. image
> 
> https://tinyurl.com/2d94wol3

It should be fixed now. Please confirm @yceballost :)

PR #816, opened 2023-07-11
Tooltip width mobile

BrunoGuimaraes103:
Figma: https://www.figma.com/file/il3GkMh7p85wpXvtWuvLRT/%F0%9F%94%B8-Tooltips-Specs?type=design&node-id=0-1&mode=design&t=tfAilK8IQPuwIzql-0

Changes:

- Position left and right that did not exist in mobile.
- Adjustment to change the value of the tooltip in mobile mode and remove the fixed one.
- Add a property to insert a new position on the tooltip.
- Add a property to allow the div that encompasses the target to have 100% size.
- Add a property to choose the type of alignment of the text inside the tooltip. 
- Add a direction change if the tooltip size exceeds the direction initially indicated.
- Adjust so that the tooltip has a spacing of 16px from the edges in mobile mode since it no longer occupies the entire width of the screen.


src/tooltip.tsx
```diff
@@ -116,6 +117,9 @@ const Tooltip: React.FC<Props> = ({
     targetLabel,
     delay = true,
     dataAttributes,
+    textAlign = 'left',
```
AlexJCelmer, line 120: I did a quick search and from the examples I saw it might be interesting to leave the center as default value. Even though it's the default of lib today it seems odd to me that it's left-aligned. This can be discussed @FernandoRRosa

src/tooltip.tsx
```diff
@@ -135,10 +140,41 @@ const Tooltip: React.FC<Props> = ({
...
 
-    const position = getPosition(rest.position);
+    const getPosition = (position: Position | undefined) => {
+        let finalPosition = position ? position : defaultPositionDesktop;
+        if (isTabletOrSmaller) {
+            finalPosition = position ? position : defaultPositionMobile;
+        }
+
+        return finalPosition;
+    };
```
AlexJCelmer, line 152: Maybe a const? and make a short handed if for the isTabletOrSmaller

`const finalPosition = isTabletOrSmaller ? (position || defaultPositionMobile) : (position || defaultPositionDesktop);`

src/tooltip.tsx
```diff
@@ -251,9 +325,17 @@ const Tooltip: React.FC<Props> = ({
...
+        if (tooltipRef.current && isVisible) {
+            const widthAux = width ? width : 0;
+            setContainerPosition(getContainerPosition(position, widthAux));
+        }
+    }, [isVisible, getContainerPosition, position, width]);
+
     return (
         <>
             <div
+                style={{width: fullWidth ? '100%' : ''}}
```
AlexJCelmer, line 338: Why not use a class here?

AlexJCelmer (changes requested):
@daviteixeira-tlf  @BrunoGuimaraes103  I made some notes

src/tooltip.tsx
FernandoRRosa: I understand that if it is 'center' it will gain a strange effect visually, I think that aligned to the left gives the feeling of a paragraph structure.

src/tooltip.tsx
AlexJCelmer: Looking here again, we can use:

`return isTabletOrSmaller ? (position || defaultPositionMobile) : (position || defaultPositionDesktop);`

There is no need for the finalPosition const

FernandoRRosa (approved):

AlexJCelmer (approved):

src/tooltip.tsx
```diff
@@ -105,6 +103,9 @@ type Props = {
     targetLabel: string;
     delay?: boolean;
     dataAttributes?: DataAttributes;
+    fullWidth?: boolean;
```
yceballost, line 106: I'm not sure why the tooltip should resize its target 🤔

src/tooltip.tsx
```diff
@@ -105,6 +103,9 @@ type Props = {
...
+    textAlign?: 'center' | 'left' | 'right' | 'justify';
+    changedPosition?: string;
```
yceballost, line 108: Just curious, what is this prop?

src/tooltip.tsx
BrunoGuimaraes103: the resizing was necessary to align the Tooltip with the thumb of the component slider.
 Without the redirect, the tooltip would extend beyond the width of the new Slider component.

image

src/tooltip.tsx
BrunoGuimaraes103: fullWidth?: boolean; 
Necessary for aligning the Tooltip with the Thumb of the Slider component.

textAlign?: 'center' | 'left' | 'right' | 'justify';  
Removed. As requested, we leave only the options:
Left when the component is larger than minWidth
Centered when the component has the size of minWidth

changedPosition?: string;
This property makes the Tooltip change its position according to the space available for presentation without breaking

src/tooltip.tsx
yceballost: why doesn't the slider component handle this with the tooltip?

AlexJCelmer (approved):

src/tooltip.tsx
yceballost, line 1: I found a buggy interaction in tooltip, when you click on some cases the tooltip remains fixed. The tooltip is an element that should work only with hover on desktop

https://github.com/Telefonica/mistica-web/assets/6722153/d6ce4f00-6aa2-42fb-a80a-1fb0d444b484

https://tinyurl.com/2qfjjsvx

yceballost (changes requested):

src/tooltip.tsx
daviteixeira-tlf: This interaction bug already exists in the tooltip that is in the master, so we will not deal with it right now. However, for a future adjustment it is possible to carry out.

image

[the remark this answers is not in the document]
BrunoGuimaraes103: This bug is tricky, the text wraps due to normally being from left to right, where this ends up making the container grow and as soon as the collision occurs it breaks and ends up misaligning the arrow.

But because of the useEffect that gets a reference the more the mouse hovers over it, it ends up 'correcting itself' as shown in this case:

https://github.com/Telefonica/mistica-web/assets/123959276/17f59a77-bce3-4d42-bb39-43441b226a19

A solution that I thought of to this, which still has problems, is to move the container to the left, but this could misalign with the design of how it occurs on the left side and there is still a lack of logic to correct the alignment of the arrow in this case (a since it would be like 'changing' the prop position from top to left in the code itself in this case):

https://github.com/Telefonica/mistica-web/assets/123959276/0b200977-ca1e-4cc2-8484-08e829ff2778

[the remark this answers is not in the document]
BrunoGuimaraes103: @yceballost I managed to fix this possible solution for the bug, what do you think?

Captura de tela 2023-07-21 101436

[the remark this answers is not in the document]
yceballost: its fine for me

yceballost: in your example it happens because you are in mobile view. That's is correct. In mobile when you tap in the element the tooltip appear and remains static until another tap is made.

yceballost, line 1: Another bug is found in mobile views

The behavior is correct except if you tap in the same element multiple times.

Try: 
- Open [this link](https://mistica-bfc32j88v-tuentisre.vercel.app/playroom/preview/#?code=N4Igxg9gJgpiBcIA8AVCEA2AXAlgBwB0A7AAhKwEMAnAcxiwF5hUYAPLAZgD4BnCAWxjlqdLABpyACxw8SYCqQBGQhQE85AvBCIwiWJAHoUbTlwC+xMrB5gq+XNoYEQAZXq4iNHs+IGuxEDEQLEkYQQQQAFkIADcZSioQMyA) in mobile view (you can use chrome dev tools in mobile mode)
- Click on the text
- [Tooltip appears]
- Click again on the text
- [Tooltip disappears]
- Click again on the text
- [Tooltip not appears] (in this case the tooltip has to appear again)

to make the tooltip reappear you must click outside the text and then click on the text again.

yceballost: I tried the interaction on mobile, but sometimes a little glitch appear in the animation. I think it is not important but would be nice to avoid this.

[tried with this example](https://mistica-bfc32j88v-tuentisre.vercel.app/playroom/#?code=N4Igxg9gJgpiBcIA8AVCEA2AXAlgBwB0A7AAhKwEMAnAcxiwF5hUYAPLAZgD4BnCAWxjlqdLABpyACxw8SYCqQBGQhQE85AvBCIwiWJAHoUbTlwC%2BxMrB5gq%2BXNoYEQAZXq4iNHs5JaeOByInECwIQhASYgMuYhAxEMkYQR4EAG0QAFkIADcZSioQAF14gHccKCxJFPhUjgA2AAYxAHY6gA5CsyA)

https://github.com/Telefonica/mistica-web/assets/6722153/285a5346-fed0-4767-856b-bc076783c5b4

yceballost (changes requested):

src/tooltip.tsx
BrunoGuimaraes103: The bug seems to also occur on the desktop, in this case it is in the playroom of the code in production

Captura de tela 2023-07-21 144947

BrunoGuimaraes103: I tested this bug and it seems that it is also in production

https://github.com/Telefonica/mistica-web/assets/123959276/d9d9fcd4-f1a0-4e49-83e4-4cda3a39e247

yceballost: i will create ticket to fix this in the future

yceballost: i will create ticket to fix this in the future

aweell (commented):
https://tinyurl.com/28xqd92g

Current

Screenshot 2023-07-26 at 11 19 54
Screenshot 2023-07-26 at 11 19 59


This PR
Screenshot 2023-07-26 at 11 20 34
Screenshot 2023-07-26 at 11 20 40


When the item right is hovered shouldn't the tooltip to appear at the limit of the available space and the nubbin pointing to the target? Is the last screenshot expected?

BrunoGuimaraes103:
fullwidth has been re-added due to requirements for the new component Slider. Where the width 100% is necessary, which by default the component Tooltip does not have this behavior.

BrunoGuimaraes103:
@aweell 
The fullWidth Property is for specific use cases such as the Slider Component (which still does not have the PR open), where unfortunately it was the only solution found to correct a size problem where without this property the Slider bar was shortened .

However, this latest Tooltip update has a solution for another bug that was found by the last comment, where the tooltip container was being rendered off-screen, now this has been resolved.

src/tooltip.tsx
```diff
@@ -251,11 +371,20 @@ const Tooltip: React.FC<Props> = ({
...
+    }, [isVisible, getContainerPosition, position, width]);
+
+    const targetStyle = fullWidth && styles.fullWidth;
+
     return (
         <>
             <div
                 ref={targetRef}
-                className={styles.wrapper}
+                className={classnames(styles.wrapper, targetStyle)}
```
atabel, line 387: you can conditionally apply classnames like this:
```ts
className={classnames(styles.wrapper, {[styles.targetStyle]: condition})}
```

src/tooltip.tsx
```diff
@@ -135,10 +142,64 @@ const Tooltip: React.FC<Props> = ({
...
     });
 
-    const getPosition = (position: Position = defaultPositionDesktop) =>
-        isTabletOrSmaller && (position === 'left' || position === 'right') ? defaultPositionMobile : position;
+    const [containerPosition, setContainerPosition] = React.useState({});
+
+    const getPosition = (position: Position | undefined) =>
+        isTabletOrSmaller ? position || defaultPositionMobile : position || defaultPositionDesktop;
+
+    const noRightSpace = (position: Position) => {
```
atabel, line 150: why do you pass `position` to this function? it doesn't make sense, this function should just return a boolean, and the postition isn't relevant.

Also, try to avoid negations in function names. I'd call this function `hasRightSpace`

src/tooltip.tsx
```diff
@@ -135,10 +142,64 @@ const Tooltip: React.FC<Props> = ({
...
+        if (!tooltipRef.current) return position;
+        const tooltipBoundingClientRect = tooltipRef.current.getBoundingClientRect();
```
atabel, line 152: `getBoundingClientRect` is an expensive operation, you could call it just once, and with the result then calculate if there is enough right/left/top/bottom space

src/tooltip.tsx
```diff
@@ -135,10 +142,64 @@ const Tooltip: React.FC<Props> = ({
...
+
+        return targetBoundingClientRect.current.left < tooltipBoundingClientRect.width;
+    };
+
+    const validatePosition = (position: Position) => {
+        if (!tooltipRef.current) return position;
+        const tooltipBoundingClientRect = tooltipRef.current.getBoundingClientRect();
+
+        const validateLeft = (position: Position) => {
+            const noTopSpace =
```
atabel, line 169: this variable has a bad name, what does this mean? I'd call it something like `alternativePosition`

```ts
return hasLeftSpace() ? 'left' : alternativePosition;
```

src/tooltip.tsx
```diff
@@ -135,10 +142,64 @@ const Tooltip: React.FC<Props> = ({
...
+                window.innerHeight
+                    ? validateRight('top')
+                    : validateRight(position),
+        };
+
+        return positionValidated[position];
+    };
 
-    const position = getPosition(rest.position);
+    const position = validatePosition(getPosition(rest.position));
```
atabel, line 202: your are making dom calculation `getBoundingClientRect` inisde `validatePosition`, you can't do that at render time, you should do that inside a react effect

src/tooltip.tsx
BrunoGuimaraes103: the reason why it had position was that the tooltipRef ended up complaining if it could not return anything because it could be null, to work around this and not need to pass the position, I changed it to return as undefined

src/tooltip.tsx
BrunoGuimaraes103: validateLeft and validateRight, are validating if there is space on the screen to the left and right, depending on whether it has, it will return the position that was passed.

In validate Left checks if there is no space on the left and also not on top, it would end up going to bottom.

atabel: I mean the `noTopSpace` variable is a bad name

src/tooltip.tsx
```diff
@@ -178,84 +249,146 @@ const Tooltip: React.FC<Props> = ({
...
+        if (!tooltipClient) {
+            return;
+        }
+
+        const containerAlign = tooltipClient.width >= minWidth ? 'left' : 'center';
+
+        return containerAlign;
+    };
+
+    const textAlign = getContainerAlign();
```
atabel, line 360: I think you don't need this. If you need to center the content when the tooltip width is `minWidth` (40), you can do that just with css.
```ts
display: 'flex,
flexDirection: 'column',
alignItems: 'center',
```
something like this should do the trick, give it a try

src/tooltip.tsx
```diff
@@ -28,7 +27,7 @@ const noOp = () => {};
 
 type Position = 'top' | 'bottom' | 'left' | 'right';
 
-const getWidthDesktop = (customWidth?: number) => (customWidth ? customWidth : defaultWidthDesktop);
+const getWidthTooltip = (customWidth?: number) => (customWidth ? customWidth : '');
```
atabel, line 30: If we don't need a default desktop width, then this function doesn't make sense anymore, please just remove it.

src/tooltip.tsx
```diff
@@ -178,84 +249,146 @@ const Tooltip: React.FC<Props> = ({
...
+            const tooltipMarginFix =
+                tooltipClient.width > targetBoundingClientRect.current.width &&
+                targetBoundingClientRect.current.left + targetBoundingClientRect.current.width / 2 <
+                    tooltipClient.width / 2 + 16
+                    ? Math.round(tooltipClient.width / 2 - targetBoundingClientRect.current.width / 2)
+                    : 0;
+
+            const bottomAdjustment = changedPosition
+                ? changedPosition
+                : !width
```
atabel, line 283: what's your intention here? `bottomAdjustment` is meant to be a number? then why are you assigning string values here? (isnt `changedPosition` a string?) this doesn't make any sense to me

src/tooltip.tsx
```diff
@@ -135,10 +139,77 @@ const Tooltip: React.FC<Props> = ({
...
+    const hasSpace = React.useCallback(
+        (position: Position) => {
+            if (!tooltipClient) return undefined;
+
+            return position === 'left'
+                ? targetBoundingClientRect.current.left < tooltipClient.width
+                : targetBoundingClientRect.current.right + tooltipClient.width > window.innerWidth;
+        },
+        [tooltipClient]
+    );
```
atabel, line 167: what's the intention of this function? it says if there is enough space in the given position? then the logic doesn't seem to be taking into account the 'top' / 'bottom' cases. It seems to be handling only the right/left cases. If that's true, then type it as `(position: 'left' | 'right') =>`. Or even better, just create two different functions (`hasLeftSpace` and `hasRightSpace`). Also I think you aren't taking into account the space between target and tooltip in the calcs

src/tooltip.tsx
```diff
@@ -135,10 +139,77 @@ const Tooltip: React.FC<Props> = ({
...
+
+    const validatePosition = React.useCallback(
+        (position: Position) => {
+            if (!tooltipClient) return position;
+
+            const alternativePositionLeft = (position: Position) => {
+                const hasTopSpace =
+                    targetBoundingClientRect.current.top < tooltipClient.height ? 'bottom' : 'top';
```
atabel, line 175: a variable with a `hasXXX` name should be a boolean.

src/tooltip.tsx
```diff
@@ -135,10 +139,77 @@ const Tooltip: React.FC<Props> = ({
...
+                : targetBoundingClientRect.current.right + tooltipClient.width > window.innerWidth;
+        },
+        [tooltipClient]
+    );
+
+    const validatePosition = React.useCallback(
+        (position: Position) => {
+            if (!tooltipClient) return position;
+
+            const alternativePositionLeft = (position: Position) => {
```
atabel, line 173: functions should have a verb name

atabel (changes requested):

src/tooltip.tsx
```diff
@@ -116,6 +117,8 @@ const Tooltip: React.FC<Props> = ({
     targetLabel,
     delay = true,
     dataAttributes,
+    fullWidth,
+    changedPosition,
```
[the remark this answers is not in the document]
BrunoGuimaraes103: This property is a string because it can receive a dynamic position or another position that you want to set, like 20 px, where you can change the position of the Tooltip.

We created this prop for the use case like the PR of the Slider, we would need to be able to dynamically change the position of the Tooltip

src/tooltip.tsx
BrunoGuimaraes103: Ok, I separated it into two different functions, regarding the calculations, the part that takes into account the space between the target and the tooltip calcs is in another part of the code (I believe it is in containerPos), originally I had separated this verification to simplify how it is used in many different parts of the code in various calculations.

src/tooltip.tsx
pladaria, line 1: Please, review this


https://github.com/Telefonica/mistica-web/assets/579417/6c40e87c-cae3-43e6-963f-9c3f65be3508

src/tooltip.tsx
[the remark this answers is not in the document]
pladaria: I understand the motivation behind this prop but this this shouldn't be needed with a correctly implemented tooltip (although refactoring it is out of scope of this task). This prop is not the best solution but until we rewrite the component, this will help us moving forward workarounding tooltip issues.

Please, rename this prop to something like `unstable_offsetX` or `unstable_leftPosition`. The type of this prop should be `number` (not string) and that number will be added to the initial X position to be able to move it (this `unstable_` prefix is a react convention to name things that will change or be removed in the future. Add the eslint ignore comment if needed).

This component currently has many issues that should be addressed in a different task:
* bad positioning:
image
* strange interactions (works with hover/click and both implementations collide)
* there should be a mechanism to prevent showing two tooltips at the same time
* perhaps we should implement new interactions, like showing on tap and hide on tap release
* Position the tooltip absolutely to a relative target, so if the target moves (by scrolling the page or sliding it), the tooltip will automatically move. [Example here](https://codepen.io/pladaria/pen/BavBWaj)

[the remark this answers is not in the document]
yceballost: Agree with this approach to move forward with this RP

pladaria (changes requested):
please, review the no delay case and apply the proposed changes in a previous comment

BrunoGuimaraes103:
@pladaria we reviewed the case of no delay, in this case this error happened due to an infinite loop in the rendering, in the last update of the PR the error was corrected

Captura de tela 2023-08-16 091053

src/tooltip.tsx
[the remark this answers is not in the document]
pladaria: @BrunoGuimaraes103 please, read my previous comment

src/tooltip.tsx
```diff
@@ -135,10 +136,77 @@ const Tooltip: React.FC<Props> = ({
...
+
+    const getPosition = React.useCallback(
+        (position: Position | undefined) =>
+            isTabletOrSmaller ? position || defaultPositionMobile : position || defaultPositionDesktop,
+        [isTabletOrSmaller]
+    );
+
+    const [tooltipClientRect, setTooltipClientRect] = React.useState<DOMRect | undefined>(undefined);
+
+    const tooltipBoundingClientRect = React.useCallback(() => {
```
pladaria, line 149: this is a function, use a verb: `getTooltipRect` or similar

src/tooltip.tsx
```diff
@@ -135,10 +136,77 @@ const Tooltip: React.FC<Props> = ({
...
+        return targetBoundingClientRect.current.right + tooltipClientRect.width > window.innerWidth;
+    }, [tooltipClientRect]);
+
+    const hasLeftSpace = React.useCallback(() => {
+        if (!tooltipClientRect) return undefined;
+
+        return targetBoundingClientRect.current.left < tooltipClientRect.width;
+    }, [tooltipClientRect]);
+
+    const validatePosition = React.useCallback(
```
pladaria, line 167: you can remove all the `useCallbacks`. They are not needed unless you want to pass the resulting function to a child component to avoid re-renders

src/tooltip.tsx
```diff
@@ -135,10 +136,77 @@ const Tooltip: React.FC<Props> = ({
...
+        },
+        [hasLeftSpace, hasRightSpace, tooltipClientRect]
+    );
+
+    const [position, setPosition] = React.useState<Position | undefined>(rest.position);
 
-    const position = getPosition(rest.position);
+    React.useEffect(() => {
+        setPosition(validatePosition(getPosition(rest.position)));
+    }, [getPosition, rest.position, validatePosition]);
```
pladaria, line 209: this is a derived state, no need to store in a new state, you can use a regular variable:
```
const position = validatePosition(getPosition(rest.position));
```

This will simplify the code and avoid re-renders

src/tooltip.tsx
pladaria, line 1: https://github.com/Telefonica/mistica-web/assets/579417/a73da362-d15a-4bac-af7a-112247a3ac08

Please, review this strange behavior. The tooltip appears with different sizes

BrunoGuimaraes103: @pladaria We detected this problem during development, unfortunately we don't have a way to solve it, due to the setTooltipClientRect, getting the updated information from the containerPosition would go into an infinite loop, causing that error you mentioned earlier: https://github.com/Telefonica/mistica-web/pull/816#discussion_r1293578502

If you have any ideas or suggestions, we would appreciate it.

pladaria (approved):

yceballost (approved):

src/tooltip.tsx
BrunoGuimaraes103: @yceballost unfortunately the prop width only accepts numbers, but how about 300?

PR #819, opened 2023-07-14
feat(Nakedcard): new component

atabel:


src/__stories__/naked-card-story.tsx
```diff
@@ -0,0 +1,202 @@
...
+        control: {type: 'select'},
+    },
+    actions: {
+        options: ['button', 'link', 'button and link', 'on press', 'none'],
+        control: {type: 'select'},
+    },
+};
+
+type SmallArgs = {
+    media: 'image' | 'circular image' | 'video';
```
yceballost, line 147: Video in smallNaked is not allowed on the design side because this card will be used for reduced spaces and the video controls will not fit in many cases. should we remove the video arg from smallNaked story?

wdyt @aweell ?

src/image.tsx
```diff
@@ -148,7 +169,7 @@ export const ImageContent = React.forwardRef<HTMLImageElement, ImageProps>(
         const [isLoading, setIsLoading] = React.useState(true);
         const [hideLoadingFallback, setHideLoadingFallback] = React.useState(false);
 
-        const ratio = typeof aspectRatio === 'number' ? aspectRatio : RATIO[aspectRatio];
+        const ratio = props.circular ? 1 : typeof aspectRatio === 'number' ? aspectRatio : RATIO[aspectRatio];
```
yceballost, line 172: what about now with <Circle backgroundImage />? are we support both ways to use circular images in Mística?

yceballost (changes requested):

src/card.tsx
yceballost, line 1: I broke it 🤭
https://tinyurl.com/2ptxpkpe

Maybe a problem with zIndex and carousel

missing in specs the behavior of buttons in a carousel like this, can you add it to specs @aweell ?
https://mistica-web.vercel.app/?path=/story/components-cards-media-card--group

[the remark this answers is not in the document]
atabel: I think this should be fixed

src/__stories__/naked-card-story.tsx
atabel: Take into account that there isn't video controls here, as in MediaCard.
From the source side we can not ban video as media, but I can remove it from the story

src/image.tsx
atabel: The problem with `<Circle backgroundImage />` is that we won't have the skeleton (loading) and fallback (error) cases that we have in Image/Video

src/card.tsx
atabel: zindex issue fixed
should buttons align to bottom?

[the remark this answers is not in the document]
yceballost: still showing this grey layer 🤔

https://tinyurl.com/2k7t32ol

src/__stories__/naked-card-story.tsx
yceballost: I would remove it from story to avoid the usage

src/image.tsx
yceballost: mmm yep, that's my point. Should we evolve Circle to allow this behavior? I understand your approach but now we have two elements in Mística (Circle + Image Circle) with the same usage, is it a problem to have 2 elements to get the same result? When do I have to use circle with image or image circular?

src/card.tsx
yceballost: To be consistency with media cards, yes, we should align the buttons to the bottom

pladaria (approved):

src/card.tsx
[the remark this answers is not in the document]
atabel: arggg, I fixed it in NakedCard, but not in SmallNakedCard. It should be fixed now (deploying)

src/__stories__/naked-card-story.tsx
atabel: only for small? regular NakedCard does allow video?

src/image.tsx
atabel: there is another difference, [you can have a children in Circle](https://mistica-web.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8BhAlgJzAGxgAgCMBDMAawHMMIBXAOygEkBbI8mAXgB0QALAFz4AHAM7wA9GMFoww6kwB0gnhD4RhYgIwAmAAw7NundzzC0ALw7AAbABYAvgD5OtPHiQAVGAA8%2BNvExgoNDkHAAkYbGwIJDFPHxsnWhj0LFxEkAAaED4eGADhBABtEABZCAA3NGE%2BIgxMkAA1NHKIEABdLIB3NCgcgvhCgGYrHTa7IA)

But If we don't think this is a useful feature, I think we could remove/deprecate backgroundImage from Circle

src/__stories__/naked-card-story.tsx
yceballost: Yes, regular naked allow video

src/card.tsx
atabel: fixed

src/__stories__/naked-card-story.tsx
atabel: done

src/image.tsx
yceballost: that's a good difference. I didn't take it into account, we can keep supporting both cases

src/__screenshot_tests__/__image_snapshots__/naked-card-screenshot-test-tsx-naked-card-closeable-1-snap.png
yceballost, line 1: Should we include testing for NakedCard Group? like in the rest of the cards

https://mistica-mqomxxc0q-tuentisre.vercel.app/?path=/story/components-cards-display-media-card--group
https://mistica-mqomxxc0q-tuentisre.vercel.app/?path=/story/components-cards-display-data-card--group
https://mistica-mqomxxc0q-tuentisre.vercel.app/?path=/story/components-cards-data-card--group
https://mistica-mqomxxc0q-tuentisre.vercel.app/?path=/story/components-cards-media-card--group
https://mistica-mqomxxc0q-tuentisre.vercel.app/?path=/story/components-cards-poster-card--group
https://mistica-mqomxxc0q-tuentisre.vercel.app/?path=/story/components-cards-snap-card--group

src/card.tsx
```diff
@@ -1070,12 +1235,13 @@ export const PosterCard = React.forwardRef<HTMLDivElement, PosterCardProps>(
         const {textPresets} = useTheme();
 
         const isTouchable = touchableProps.href || touchableProps.to || touchableProps.onPress;
-        const overlayStyle = styles.touchableCardOverlayMedia;
 
         return (
             <CardContainer
                 width={width}
                 height={height}
+                dataAttributes={{...dataAttributes, 'component-name': 'PosterCard'}}
```
marcoskolodny, line 1243: Maybe you can use `getPrefixedDataAttributes` here

marcoskolodny (approved):

src/card.tsx
atabel: no, that's already done inside CardContainer

src/card.tsx
```diff
@@ -495,6 +510,163 @@ export const MediaCard = React.forwardRef<HTMLDivElement, MediaCardProps>(
...
+                dataAttributes={{'component-name': 'NakedCard', ...dataAttributes}}
+                aria-label={ariaLabel}
+                className={isTouchable ? styles.touchableContainer : undefined}
+            >
+                <BaseTouchable maybe {...touchableProps} className={styles.touchable} aria-label={ariaLabel}>
+                    <div className={styles.mediaCard}>
+                        <div style={{position: 'relative'}}>
+                            {isTouchable && (
+                                <div
+                                    className={classNames(styles.touchableMediaOverlay, {
```
aweell, line 552: The media of the naked differs from the media hover definition, and uses the same overlay that displayMedia and poster cards with the following tokens:

`coverBackgroundHover`, `coverBackgroundPressed`

aweell (changes requested):

src/card.tsx
atabel: fixed, looks much better now!

aweell (approved):
LGTM!

PR #822, opened 2023-07-18
feat(Header): small version, allow removing vertical padding

atabel:
Jira with specs: https://jira.tid.es/browse/WEB-1469

- Added a `small` prop in `Header` to reduce the title/description font size
- Added a `noPaddingY` prop to `HeaderLayout` to remove the vertical padding (required by Novum CMS layout, which establishes its own paddings)
- Added support for responsive space in `Stack` and `Box` components:
```ts
<Stack space={{mobile: 8, tablet: 16, desktop: 24}}>

<Box paddingY={{mobile: 16, desktop: 24}}>
```
When using the responsive notation (object), setting both `mobile` and `desktop` is mandatory. `tablet` is optional, and will use `mobile` value by default when not specified.
This will allow us to reduce the use of `useScreenSize()` hook, which is problematic with SSR.
In this PR I've applied this to header components, but we should do the same with the rest of components in Mística too. It's still pending to apply the same approach in `Inline` component


src/sprinkles.css.ts
```diff
@@ -19,6 +20,26 @@ const colorProperties = defineProperties({
 
 const sizes = [0, 2, 4, 8, 12, 16, 20, 24, 32, 40, 48, 56, 64, 72, 80] as const;
 
+const responsiveProperties = defineProperties({
```
atabel, line 23: this allows us using responsive padding sprinkles:
```ts
sprinkles({padding: {mobile: 8, desktop: 16}})
```

src/header.tsx
```diff
@@ -175,81 +187,95 @@ export const HeaderLayout: React.FC<HeaderLayoutProps> = ({
...
-                    </Box>
-                ) : (
-                    <Box paddingTop={breadcrumbs ? 16 : 48} paddingBottom={isBleedActivated ? 32 : 48}>
-                        <Stack space={isTabletOrSmaller ? 24 : 32}>
-                            <Stack space={32}>
-                                {breadcrumbs}
-                                {header}
-                            </Stack>
+                    ) : (
+                        <Stack space={header ? {mobile: 24, desktop: 32} : 0}>
```
yceballost, line 234: When Bleed and noPaddingY property are enabled, the padding between extra and description disappear, I think this is not expected

https://github.com/Telefonica/mistica-web/assets/6722153/e2400862-71f1-440e-8869-c5d9f59bbaa5

yceballost (changes requested):

pladaria (approved):

wyunreal (approved):

src/__stories__/header-story.tsx
```diff
@@ -18,111 +17,105 @@ export default {
...
+    small: false,
+    isInverse: true,
+    withExtraContent: true,
+    sideBySideExtraOnDesktop: true,
+    withBreadcrumbs: true,
+    bleed: false,
+    noPaddingY: false,
+};
+Default.argTypes = {
+    pretitle: {if: {arg: 'withHeader', value: true}},
```
marcoskolodny, line 98: No need to add `value:true` here. `{if: {arg: 'withHeader'}}` is equivalent.

src/header.tsx
```diff
@@ -175,81 +187,95 @@ export const HeaderLayout: React.FC<HeaderLayoutProps> = ({
...
     children,
+    dataAttributes,
 }) => {
-    const {isTabletOrSmaller} = useScreenSize();
-
     return (
-        <ResponsiveLayout isInverse={isInverse}>
+        <ResponsiveLayout
+            isInverse={isInverse}
+            dataAttributes={{'component-name': 'MainSectionHeaderLayout', ...dataAttributes}}
```
marcoskolodny, line 267: Maybe you can use `getPrefixedDataAttributes` here?

marcoskolodny (approved):

src/header.tsx
atabel: no, that would be wrong. That's already done internally in ResponsiveLayout

src/header.tsx
atabel: fixed, good catch!

yceballost (approved):

Issue #626, opened 2022-12-28
Major release plan 17.0.0 (Breaking changes)


atabel:
* [ ] Remove Vivo New skin (**blocked**, wait for force update in Novum)
* [ ] Remove TU skin
* [ ] Remove O2 New skin
* [ ] Dont transpile dist-es/ to es5
    * (@sergivillar) As we are using VE (that requires css variables, available since chrome 49) maybe we could compile only to es6 (chrome 51)?
* [ ] Replace `data-component-name` with `data-testid` in all mistica components
* [ ] Review and remove deprecated props in several components
* [ ] Remove `legacyDisplay` border radius token and replace to `container`
* [ ] Rename palette colors
* [ ] Rename `Extra` to `Slot`
* [ ] Remove fixed width by default in input fields

yceballost:
just a check, but provide an easy instructions for the major release in release notes

brtbrt:
@yceballost  is it anything you've just made? have you seen other repos around offering this and you liked it? paste it and tag me, so that I'll take a look

PR #824, opened 2023-07-18
feat(MediaCard): add top icon support

thiagoaraujo-tlf:
Figma: https://www.figma.com/file/Y89JWviWrWV4RFFyqwCj3b/Asset-Media-Card---Circle-text?type=design&node-id=110-14234&mode=design&t=N44KdlRs1GH019sK-0


src/card.tsx
```diff
@@ -407,6 +407,7 @@ interface MediaCardBaseProps {
     subtitle?: string;
     subtitleLinesMax?: number;
     description?: string;
+    icon?: React.ReactElement
```
AlexJCelmer, line 410: In figma it was called Asset, you may need to make this switch from icon to Asset
@FernandoRRosa

src/card.css.ts
```diff
@@ -147,6 +147,22 @@ export const mediaCardContent = style([
     },
 ]);
 
+export const mediaCardIcon = style([
```
AlexJCelmer, line 150: If icon get changed to Asset will need to change here too.

FernandoRRosa (approved):

AlexJCelmer (approved):

src/card.tsx
yceballost: yes, better call "asset" will be the new name for all props that contains assets. We want to make this breaking change in the next major and rename `icon` to `asset` to be more generic →  https://github.com/Telefonica/mistica-web/issues/626

In my opinion, we have to embrace the new nomenclature and use `asset`

yceballost (approved):
nice 🤩

src/card.tsx
atabel: hmmm... I think we should still use icon, as we do in other Card components. I agree with Yayo that we have a pending rename for this prop, but we should do that rename in all the cards at the same time. That rename will be done in a future mistica major release, not now

src/__screenshot_tests__/__image_snapshots__/media-card-screenshot-test-tsx-media-card-with-body-closeable-1-snap.png
atabel, line 1: what happened here?
Please review that extra bottom padding

src/card.tsx
```diff
@@ -516,6 +518,13 @@ export const MediaCard = React.forwardRef<HTMLDivElement, MediaCardProps>(
                                     buttonLink={buttonLink}
                                 />
                             </div>
+                            {asset ? (
+                                <Box className={styles.mediaCardAsset} paddingX={16} paddingY={16}>
+                                    {asset}
+                                </Box>
+                            ) : (
+                                <Box paddingBottom={actions?.length || onClose ? 64 : 0} />
```
atabel, line 526: This is not needed. It's needed in other cards like DisplayMediaCard because the `asset` is not absolutely positioned like here.

src/card.css.ts
```diff
@@ -177,6 +177,22 @@ export const mediaCardContent = style([
...
+    {
+        position: 'absolute',
+        zIndex: 1,
+    },
+    {
+        '@media': {
+            [mq.desktopOrBigger]: {
+                paddingLeft: 24,
+                paddingTop: 24,
+                paddingBottom: 32,
```
atabel, line 190: @aweell, it's ok to have different paddings in desktop? I don't see this in the specs.

If these are needed, we should move them to the Box element instead of creating a new media query here @thiagoaraujo-tlf:

```ts
<Box paddingLeft={{mobile: 16, desktop: 24}} paddingTop={{mobile: 16, desktop: 24}}>
```
Also, I think `paddingRight` and `paddingBottom` are not needed, just left and top

atabel (approved):

FernandoRRosa (approved):

src/card.css.ts
aweell: Yes, the asset has a padding left and top of 16 in mobile and 24 in desktop.

Mobile:
Screenshot 2023-07-26 at 17 02 43

Desktop:
Screenshot 2023-07-26 at 17 02 46

This is intended to align the asset to the content since the padding of the content increases in desktop.

The other changes, it seems they're related the content padding? They are already working in the current component.

Mobile:
Screenshot 2023-07-26 at 17 06 17

Desktop:
Screenshot 2023-07-26 at 17 05 19

gabrielmarquesini-tlf: I just added icon padding for when it's on desktop @atabel

aweell (approved):

atabel (approved):

PR #827, opened 2023-07-21
fix(Select): long value overflow

marcoskolodny:
Fix overflow and added tests to verify this scenario.


src/__screenshot_tests__/form-select-screenshot-test.tsx
```diff
@@ -35,6 +35,32 @@ test.each(devices)('Select elements on a selected state appear properly on %s',
...
+            await select.select('longValue');
+        } else {
+            const select = await screen.findByLabelText('Select a fruit (opcional)');
+            await select.click();
+            const selectOptions = await screen.findAllByRole('option', {
+                name: 'A very very long text value for this option',
+            });
+            // take the last one because the options are displayed in a portal, at the end of the body
+            await selectOptions?.at(-1)?.click();
+        }
```
atabel, line 57: at this point, I'd prefer two different tests instead of having this logic here

atabel (approved):

src/__screenshot_tests__/form-select-screenshot-test.tsx
marcoskolodny: fixed

PR #828, opened 2023-07-21
fix(community): issue with community build preventing import of community components

atabel:


package.json
```diff
@@ -12,7 +12,9 @@
     "files": [
         "dist/**",
         "dist-es/**",
-        "css/**"
+        "css/**",
+        "community.d.ts",
+        "community.js"
```
atabel, line 17: these were not being publised

scripts/compile.js
```diff
@@ -19,7 +19,7 @@ const compile = () => {
 
     // Entry point for community folder (import {Component} from '@telefonica/mistica/community')
     run(`echo "export * from './dist/community';" > community.d.ts`);
-    run(`echo "export * from './dist/community';" > community.js`);
+    run(`echo "export * from './dist-es/community';" > community.js`);
```
atabel, line 22: It's important that `import {Component} from '@telefonica/mistica/community'` use  the `dist-es` folder because webpack uses `dist-es` when doing `import {ThemeContextProvider} from '@telefonica/mistica'`. If community components used `dist` instead of `dist-es` they would be using a different context provider and we'd have runtime errors.

We have a pending todo for next mistica major to remove the commonjs build

pladaria (approved):

scripts/compile.js
pladaria: IIRC, next uses the "dist" folder. But for webapp this should work fine

atabel: hmmm... then I think we may have problems with community components in next.
The right approach would be removing the non commonjs build, but that will be a major change

marcoskolodny (approved):

PR #829, opened 2023-07-25
fix(ButtonLink, Title): update alignments

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1481)

Related PR in webapp: [Link](https://github.com/Telefonica/webapp/pull/3388)


src/button.tsx
```diff
@@ -433,7 +433,7 @@ interface ButtonLinkCommonProps {
     trackEvent?: boolean;
     /** "data-" prefix is automatically added. For example, use "testid" instead of "data-testid" */
     dataAttributes?: DataAttributes;
-    aligned?: boolean;
+    aligned?: boolean | 'left' | 'right';
```
atabel, line 436: don't you prefer having 2 props? `leftAligned` and `rightAligned` and deprecate the current `aligned`

aweell:
Should the story of title + the playroom snippets "Title1 (with link)" and "Title2 (with link)" be updated to show the buttonLink instead?

src/button.tsx
pladaria: Does it make sense to align it to both sides?

marcoskolodny: @atabel as Pedro says, it makes no sense to align on both sides. I did it this way to avoid a breaking change for the buttons that are currently using the prop with value = true (in order to have backward compatibility).

pladaria (approved):
nice!

src/button.tsx
marcoskolodny: Discussed offline. We deprecated this prop and  introduced new ones that allow adding bleed to left, right and top+bottom

atabel (approved):

aweell (approved):
❤️

PR #831, opened 2023-07-26
fix(Cards): CardAction type

atabel:


Winde (approved):

src/card.css.ts
```diff
@@ -391,21 +391,6 @@ export const cardActionMedia = style([
     },
 ]);
 
-export const videoAction = style([
```
atabel, line 394: this was not being used :/

pladaria (approved):

src/card.css.ts
marcoskolodny: This logic became obsolete when we changed the hover/active states, and the top action for the cards with videos/images became more generic. Before that, the top action for videos used to have a different style (this one), and it was used in line 306 of `src/card.tsx`.

Good catch!

marcoskolodny (approved):

PR #840, opened 2023-08-03
feat(Sheet): new component

atabel:
WEB-1492

Figma: https://www.figma.com/file/nxToAYk7FGC86TS0mY61TU/%F0%9F%94%B8-Sheet-specs?type=design&node-id=3346-16622&mode=design&t=AfsHntMnROCmNvcD-0

Webview bridge spec: https://confluence.tid.es/display/CTO/%5BAPPS%5D+Shared+Spec%3A+Bottom+Sheet+opened+from+Webview-Bridge


doc/bottomSheet.md
```diff
@@ -0,0 +1,135 @@
...
+        </BottomSheet>
+      )}
+    </>
+  );
+};
+```
+
+The sheet will close when the user does the swipe down gesture or when the background overlay is touched. The
+`onClose` callback is called when the closing animation finishes, that's the right place to unmount the sheet
+as shown in the example above.
```
atabel, line 30: note that `onClose` is called when the closing animation finishes, so the idea is not unmounting the component until the transition finishes, that's why I set `showSheet` to false in `onClose` handler instead of `ButtonPrimary` `onPress` handler

doc/bottomSheet.md
```diff
@@ -0,0 +1,135 @@
...
+```jsx
+import {BottomSheet} from 'mistica';
+
+const MyComponent = () => {
+  const [showSheet, setShowSheet] = useState(false);
+  return (
+    <>
+      <ButtonPrimary onPress={() => setShowSheet(true)}>show bottom sheet</ButtonPrimary>
+      {showSheet && (
+        <BottomSheet onClose={() => setShowSheet(false)}>
```
atabel, line 43: note that `onClose` is called when the closing animation finishes, so the idea is not unmounting the component until the transition finishes, that's why I set showSheet to false in onClose handler instead of ButtonPrimary onPress handler

doc/bottomSheet.md
```diff
@@ -0,0 +1,135 @@
...
+  );
+};
+```
+
+## Sheet with predefined content
+
+Mística predefines some common bottom sheet patterns for you to use: `RadioListBottomSheet`,
+`ActionsListBottomSheet`, `InfoBottomSheet` and `ActionsBottomSheet`. You can see examples in the storybook.
+
+## `showBottomSheet` imperative api
```
atabel, line 63: this equivalent to the `confirm()`, `alert()` methods we have for dialogs

doc/bottomSheet.md
```diff
@@ -0,0 +1,135 @@
...
+`showBottomSheet` function. For this to work, you need to render a `<BottomSheetRoot/>` somewhere in your app,
+typically where you render the mistica `<ThemeContextProvider/>`, but it could be anywhere.
+
+```jsx
+import {BottomSheetRoot} from '@telefonica/mistica';
+
+export const App = () => {
+  return (
+    <>
+      <BottomSheetRoot />
```
atabel, line 75: I've decided to not include this automatically inside `ThemeContextProvider` to allow webs using mistica to optin to use this. The advantage is that we don't bloat webs with all sheet components JS code if they are not going to use it. Also, it isn't needed to use `BottomSheetRoot` if you don't wan't to use the imperative api

doc/bottomSheet.md
```diff
@@ -0,0 +1,135 @@
...
+### native implementation
+
+If you are using mistica inside Novum app, you can configure `showBottomSheet` to use the native sheet
+implementation with the webview bridge.
+
+```jsx
+import {BottomSheetRoot} from '@telefonica/mistica';
+import * as webviewBridge from '@tef-novum/webview-bridge';
+
+const nativeImplementation = createNativeSheetImplementationFromWebviewBridge(webviewBridge);
```
atabel, line 122: with this approach, mistica doesn't need to depend on webview bridge lib, because the native implementation is injected, not imported. I think we could follow a similar approach for `alert()` `confirm()` or other components using the bridge. That would allow us to remove the `"@tef-novum/webview-bridge"` dependency from mistica `package.json`

playroom/frame-component.tsx
yceballost, line 1: missing playroom snippets

src/__stories__/bottom-sheet-story.tsx
```diff
@@ -0,0 +1,541 @@
...
+                            </Box>
+                        </ResponsiveLayout>
+                    </BottomSheet>
+                )}
+            </ResponsiveLayout>
+        </Box>
+    );
+};
+
+Default.storyName = 'BottomSheet';
```
yceballost, line 75: this type of sheet also includes `Title`, `Subtitle` and `Description` elements

src/bottom-sheet.tsx
```diff
@@ -0,0 +1,748 @@
...
+                                    onScroll={onScroll}
+                                    className={styles.children}
+                                >
+                                    {typeof children === 'function'
+                                        ? children({closeModal, modalTitleId})
+                                        : children}
+                                </section>
+                            </div>
+                            <div className={styles.modalCloseButton}>
+                                <IconButton size={32} onPress={closeModal} aria-label={texts.modalClose}>
```
yceballost, line 283: Is it possible to use the same hover effect as cards onClose icon in this element?
image

yceballost (changes requested):

doc/bottomSheet.md
```diff
@@ -0,0 +1,135 @@
...
+      <RestOfYourApp />
+    </>
+  );
+};
+```
+
+Then you can call `showBottomSheet` from anywhere:
+
+```jsx
+import {BottomSheet} from 'mistica';
```
pladaria, line 85:
```suggestion
import {showBottomSheet} from '@telefonica/mistica';
```

doc/bottomSheet.md
```diff
@@ -0,0 +1,135 @@
...
+        })
+      }
+    >
+      show bottom sheet
+    </ButtonPrimary>
+  );
+};
+```
+
+### native implementation
```
pladaria, line 113:
```suggestion
### Native implementation
```

doc/bottomSheet.md
pladaria: Ideally we should be able to inject the bridge function "as is" to the Root component.

Something like:
```js
import {openBottomSheet} from '@tef-novum/webview-bridge`;

<BottomSheetRoot nativeImplementation={openBottomSheet} />
```

and the `nativeImplementation` prop would be typed as `typeof openBottomSheet`;

This way everything is tied, types are verified at compile time and we won't need a `createNativeSheetImplementationFromWebviewBridge`.

Perhaps the bridge dependency could be listed as `optionalDependency`

src/bottom-sheet-root.tsx
```diff
@@ -0,0 +1,290 @@
...
+            text: string;
+        };
+        secondaryButton?: {
+            text: string;
+        };
+        link?: {
+            text: string;
+            withChevron?: boolean;
+        };
+    }>;
```
pladaria, line 74: all this could be imported from the bridge, which should be the source of truth

src/stack.tsx
```diff
@@ -42,21 +42,26 @@ type Props = {
...
             style={assignInlineVars(calcInlineVars(space))}
             role={role}
-            aria-labelledby={props['aria-labelledby']}
-            {...getPrefixedDataAttributes(props.dataAttributes)}
+            aria-labelledby={ariaLabelledby}
+            {...getPrefixedDataAttributes(dataAttributes)}
         >
             {React.Children.map(children, (child) => (
-                <div>{child}</div>
+                <div role={role === 'list' ? 'listitem' : undefined}>{child}</div>
```
pladaria, line 64: :+1:

src/bottom-sheet.tsx
```diff
@@ -0,0 +1,748 @@
...
+            ? {
+                  // decrease opacity when dragging down the sheet
+                  opacity: 0.25 + 1 - dragDistance / (window.innerHeight - initialYRef.current),
+                  transition: 'none',
+              }
+            : undefined,
+    };
+};
+
+const useLockBodyScroll = () => {
```
pladaria, line 134: useLockBodyScrollStyleElement ?

pladaria (approved):

src/radio-button.css.ts
```diff
@@ -45,7 +45,7 @@ export const innerCircleChecked = style([
     }),
     {
         opacity: 1,
-        transform: 'scale(1)',
+        transform: 'none',
```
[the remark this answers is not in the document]
atabel: scale(1) creates a new stacking context, that was affecting zindex inside sheet

src/__stories__/bottom-sheet-story.tsx
atabel: hmmm... no. This is the basic component, it can have anything inside it. You can combine it with SheetBody if you need title/subtitle/description

src/bottom-sheet.tsx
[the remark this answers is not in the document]
atabel: hmmm... have you tried it? only the pink area is scrollable, as the spec says.
If you mean the scroll bar, that would complicate the implementation :/

playroom/frame-component.tsx
atabel: do you prefer a snippet using `showSheet` or the component?

src/bottom-sheet.tsx
atabel: I can give it a try. But I'm afraid it's not easy to reutilize the same component, but I could copy/paste the styles

atabel: maybe we could try to componentize IconButton with hover/press circle effect

atabel: please include in the specs the size of the hover/active circle and the size of the x icon.
I have used 24px (default icon size) for the X, but in cards we have 32px for the circle and 20px for the X

src/bottom-sheet.tsx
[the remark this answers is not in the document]
yceballost: Yes, I was referring to the scrollbar. In the rest of dialogs, the scrollbar is appearing inside
image

playroom/frame-component.tsx
yceballost: Like dialogs, a button that opens a sheet

src/__stories__/bottom-sheet-story.tsx
yceballost: Great, can you give me an example in playroom? ☺️

src/bottom-sheet.tsx
aweell: Updated to Icon: 20px / Container: 32px + color tokens for different states (both under a new comment in the component specs)

src/__stories__/bottom-sheet-story.tsx
atabel: something like this:
https://mistica-gseky3w6e-tuentisre.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8AFATjAbgSxgdwBUIIAbAZwAIB6APgB0A7RpAIQFcAXDiB9bAWwCGaAJ6MKFKNjKCARiRhQAvMADmMDgGUOgjjAAUdEGQAWEPEYCUAX3EUe6GGTIr9likpoVgdiWQ3augZGpuZGADQUHGhsMJYA3HbWtgz0DBKaZngUpjAazFTsXA5oAsIiaYxqATp6hsZZVhQAZM0UhukUSJl5HL72DADCJBD%2Bru6e3v1%2BNUH1oRYgkQBmguRxiZ0SyXZpEhLdJr0s0CJR2BwKSkYEFwpGOWyyHHcw18ZPL5cwD7BkYKUAA4vHjvP4A7DA7A8B60OxIKg9DRpGyMJYgDhHfhOBAAbRALBIgjYIAAuuEQHhsFBMWQ8QBmABsAAZwgBGZkAJgALKTrEA

marcoskolodny (commented):
Should we take into account a scenario in which the content overlaps with the close button?
[Playroom example](https://mistica-gseky3w6e-tuentisre.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA6A7APAIQK4BccRQAUAnASwFsBDYgT1QAJ6pSBnSgIwBsYoBeYAOYwcAZRyUcMABRIQLABYQA7rICUAXwb0CJGCxb8pq%2BrwB89YFsYthYidNkLlsgDT0cxLDFUBuLes0UUy0RRSV6BRhhVDQAemw8HTIqWmCUVEFbcUkZOTC1egAyQvoZFEY0UKicK20UAGFOCBtDYzMLWuss%2B1ynFRA3ADNKThtfWoCtNMYKgBUYAA8cACZ6ch5SLHJTSl29-YPDo%2BOT07Pzi8ur65vKOPml5emZrTiq4TSNVAGQHHkYdYsBAAbRAGE4lCwIAAui4QEpSFA-kD4MCAMwANgADC4AIxY5YAFmh6iAA)

image

atabel:
> Should we take into account a scenario in which the content overlaps with the close button? [Playroom example](https://mistica-gseky3w6e-tuentisre.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA6A7APAIQK4BccRQAUAnASwFsBDYgT1QAJ6pSBnSgIwBsYoBeYAOYwcAZRyUcMABRIQLABYQA7rICUAXwb0CJGCxb8pq%2BrwB89YFsYthYidNkLlsgDT0cxLDFUBuLes0UUy0RRSV6BRhhVDQAemw8HTIqWmCUVEFbcUkZOTC1egAyQvoZFEY0UKicK20UAGFOCBtDYzMLWuss%2B1ynFRA3ADNKThtfWoCtNMYKgBUYAA8cACZ6ch5SLHJTSl29-YPDo%2BOT07Pzi8ur65vKOPml5emZrTiq4TSNVAGQHHkYdYsBAAbRAGE4lCwIAAui4QEpSFA-kD4MCAMwANgADC4AIxY5YAFmh6iAA)
> 
> image

I don't think so. It's already managed by `SheetBody`. And if someone needs some special sheet behavior, it's their responsibility to use proper paddings to no overlap with the close button.

playroom/frame-component.tsx
atabel: done

src/bottom-sheet.tsx
```diff
@@ -0,0 +1,748 @@
...
+            if (!isDraggingRef.current) {
+                return;
+            }
+            const dragTime = Date.now() - dragInitTimeRef.current;
+            const dragDistance = currentYRef.current - initialYRef.current;
+            const dragSpeed = dragDistance / dragTime;
+
+            isDraggingRef.current = false;
+            setDragDistance(0);
+            if (dragDistance > 50 && (currentYRef.current > window.innerHeight * 0.75 || dragSpeed > 0.5)) {
```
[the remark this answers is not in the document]
atabel: I've been playing and I think the current values feel more natural, give it a try

marcoskolodny (commented):
Using chrome iOS, it looks like when I scroll the sheet's content, it assumes I'm scrolling the sheet itself, but just for a few milliseconds. After that, it starts scrolling the content.

This happens only sometimes, seems unstable. I used [this snippet](https://mistica-8v5dcbxsb-tuentisre.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8AhArgFwxAdgBQCcBLAWwEMCBPAHRwAI7dCYBnFgXmAAoBKO9gHx1gtBgxYALCAHcAyhJgwMXEfTEMMlAA4x4daiACSAOQBiAeQMAaUerpaCELSz2q7YjEQwAbXfpAAKl6%2B1rbuLGgARp4%2BfgayUTEhIDZq7rAsYMRanrh6BgAirFlEOUS4oWl2XjAkLnQA2mHuwnREUPkgAIzWdElxRhi1dD0pbZA4rn3aA5Fo3r4YBnQAvqupLWLAbR3%2BAEy9-Z2GQyR0B2NEE1OaOp1zC0rLaysbm63tnQDMh8EDJ8Mfpdrq1brN5otnutmu5tp9-AAWX6xY6nOhI4F5UEze4Qp4gVbQqqwnadACsyN8qOGFMxk2xd38D0hBJeb02cN2BgAbJT-mjeXSbjimXilqyie8PlyQAB2PnUs7yoUM8GPcWE14wuyczoADgV-gBZwNKu2YNx6qhWuJOtJ-gAnIaDMa6E6zdNGQZmfjNeyWrr-F0AAzOwbDEO9K5Y80i71i63%2BknwgxdUZWPp-RUjdPjGOetUsv3a9SB1MXDNHI1oroVvP02NekA%2BjVskt0AC6SclYhWPAA3LYVitaAJbOYdPRJIoljgkAB6dBYJjEchUMc4FIgDAKEisBANEAobxkNAgLsgaTtHcsA9fbnBjsrIA) in Playroom.

I couldn't reproduce this issue in desktop/android devices.

[Video](https://github.com/Telefonica/mistica-web/assets/25785151/0186e01c-8da5-40c7-aec8-32ec94d69d58)

src/bottom-sheet.tsx
atabel: done

atabel:
> Using chrome iOS, it looks like when I scroll the sheet's content, it assumes I'm scrolling the sheet itself, but just for a few milliseconds. After that, it starts scrolling the content.
> 
> This happens only sometimes, seems unstable. I used [this snippet](https://mistica-8v5dcbxsb-tuentisre.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8AhArgFwxAdgBQCcBLAWwEMCBPAHRwAI7dCYBnFgXmAAoBKO9gHx1gtBgxYALCAHcAyhJgwMXEfTEMMlAA4x4daiACSAOQBiAeQMAaUerpaCELSz2q7YjEQwAbXfpAAKl6%2B1rbuLGgARp4%2BfgayUTEhIDZq7rAsYMRanrh6BgAirFlEOUS4oWl2XjAkLnQA2mHuwnREUPkgAIzWdElxRhi1dD0pbZA4rn3aA5Fo3r4YBnQAvqupLWLAbR3%2BAEy9-Z2GQyR0B2NEE1OaOp1zC0rLaysbm63tnQDMh8EDJ8Mfpdrq1brN5otnutmu5tp9-AAWX6xY6nOhI4F5UEze4Qp4gVbQqqwnadACsyN8qOGFMxk2xd38D0hBJeb02cN2BgAbJT-mjeXSbjimXilqyie8PlyQAB2PnUs7yoUM8GPcWE14wuyczoADgV-gBZwNKu2YNx6qhWuJOtJ-gAnIaDMa6E6zdNGQZmfjNeyWrr-F0AAzOwbDEO9K5Y80i71i63%2BknwgxdUZWPp-RUjdPjGOetUsv3a9SB1MXDNHI1oroVvP02NekA%2BjVskt0AC6SclYhWPAA3LYVitaAJbOYdPRJIoljgkAB6dBYJjEchUMc4FIgDAKEisBANEAobxkNAgLsgaTtHcsA9fbnBjsrIA) in Playroom.
> 
> I couldn't reproduce this issue in desktop/android devices.
> 
>  IMG_5784.MP4

hmmm... it's dificult to reproduce, maybe you tapped the grab zone with your finger?

yceballost:
> > Should we take into account a scenario in which the content overlaps with the close button? [Playroom example](https://mistica-gseky3w6e-tuentisre.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA6A7APAIQK4BccRQAUAnASwFsBDYgT1QAJ6pSBnSgIwBsYoBeYAOYwcAZRyUcMABRIQLABYQA7rICUAXwb0CJGCxb8pq%2BrwB89YFsYthYidNkLlsgDT0cxLDFUBuLes0UUy0RRSV6BRhhVDQAemw8HTIqWmCUVEFbcUkZOTC1egAyQvoZFEY0UKicK20UAGFOCBtDYzMLWuss%2B1ynFRA3ADNKThtfWoCtNMYKgBUYAA8cACZ6ch5SLHJTSl29-YPDo%2BOT07Pzi8ur65vKOPml5emZrTiq4TSNVAGQHHkYdYsBAAbRAGE4lCwIAAui4QEpSFA-kD4MCAMwANgADC4AIxY5YAFmh6iAA)
> > image
> 
> I don't think so. It's already managed by `SheetBody`. And if someone needs some special sheet behavior, it's their responsibility to use proper paddings to no overlap with the close button.

not really, If I take the Marcos's example and use sheetbody inside without title prop, the content overlaps the dismiss button
https://tinyurl.com/2c3nx866

Maybe we can solved this case using the same title padding for children when no props is added in SheetBody

src/bottom-sheet.tsx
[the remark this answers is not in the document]
atabel: done

atabel:
> > > Should we take into account a scenario in which the content overlaps with the close button? [Playroom example](https://mistica-gseky3w6e-tuentisre.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA6A7APAIQK4BccRQAUAnASwFsBDYgT1QAJ6pSBnSgIwBsYoBeYAOYwcAZRyUcMABRIQLABYQA7rICUAXwb0CJGCxb8pq%2BrwB89YFsYthYidNkLlsgDT0cxLDFUBuLes0UUy0RRSV6BRhhVDQAemw8HTIqWmCUVEFbcUkZOTC1egAyQvoZFEY0UKicK20UAGFOCBtDYzMLWuss%2B1ynFRA3ADNKThtfWoCtNMYKgBUYAA8cACZ6ch5SLHJTSl29-YPDo%2BOT07Pzi8ur65vKOPml5emZrTiq4TSNVAGQHHkYdYsBAAbRAGE4lCwIAAui4QEpSFA-kD4MCAMwANgADC4AIxY5YAFmh6iAA)
> > > image
> > 
> > 
> > I don't think so. It's already managed by `SheetBody`. And if someone needs some special sheet behavior, it's their responsibility to use proper paddings to no overlap with the close button.
> 
> not really, If I take the Marcos's example and use sheetbody inside without title prop, the content do not respect the top padding https://tinyurl.com/2c3nx866
> 
> Maybe we can solved this case using the same title padding for children when no props is added in SheetBody (?)
> 
> Component in Figma image image image

What happens with scroll in those cases?

yceballost:
> > > > Should we take into account a scenario in which the content overlaps with the close button? [Playroom example](https://mistica-gseky3w6e-tuentisre.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA6A7APAIQK4BccRQAUAnASwFsBDYgT1QAJ6pSBnSgIwBsYoBeYAOYwcAZRyUcMABRIQLABYQA7rICUAXwb0CJGCxb8pq%2BrwB89YFsYthYidNkLlsgDT0cxLDFUBuLes0UUy0RRSV6BRhhVDQAemw8HTIqWmCUVEFbcUkZOTC1egAyQvoZFEY0UKicK20UAGFOCBtDYzMLWuss%2B1ynFRA3ADNKThtfWoCtNMYKgBUYAA8cACZ6ch5SLHJTSl29-YPDo%2BOT07Pzi8ur65vKOPml5emZrTiq4TSNVAGQHHkYdYsBAAbRAGE4lCwIAAui4QEpSFA-kD4MCAMwANgADC4AIxY5YAFmh6iAA)
> > > > image
> > > 
> > > 
> > > I don't think so. It's already managed by `SheetBody`. And if someone needs some special sheet behavior, it's their responsibility to use proper paddings to no overlap with the close button.
> > 
> > 
> > not really, If I take the Marcos's example and use sheetbody inside without title prop, the content do not respect the top padding [tinyurl.com/2c3nx866](https://tinyurl.com/2c3nx866)
> > Maybe we can solved this case using the same title padding for children when no props is added in SheetBody (?)
> > Component in Figma image image image
> 
> What happens with scroll in those cases? something like this? image looks weird to me

I would solve it exactly the same way as in the dialogs
https://tinyurl.com/2betx7tw

(The Divider between title and extra is defined in specs? I can't see it in specs)

yceballost (approved):
🤩

PR #841, opened 2023-08-03
feat(PosterCard): allow backgroundColor/variant

atabel:
WEB-1498


src/__stories__/poster-card-story.tsx
```diff
@@ -189,8 +200,17 @@ Default.argTypes = {
         control: {type: 'select'},
     },
     background: {
-        options: ['image', 'video'],
+        options: ['image', 'video', 'color'],
+        control: {type: 'select'},
+    },
+    backgroundColor: {
+        control: {type: 'color'},
+        if: {arg: 'background', eq: 'color'},
```
yceballost, line 208: I like the solution that was implemented by vivo
https://mistica-a10w6pvfo-tuentisre.vercel.app/?path=/story/components-cards-poster-card--default

You can fill to this field some hex value (`#FFF`) or a constant (backgroundBrandSecondary) instead of color picker. (From the design side, when possible, always try to use constants and not hex values)

src/card.tsx
yceballost, line 1: Same comment that @aweell said in the other PR. When the card doesn't have image or video and inverse is false, the hover should be the same or similar to row list hover or the rest of white background component hovers.

https://www.figma.com/file/tKdPOfcUALzVIh5oizFbm7/%F0%9F%94%B8-Cards-Specs?type=design&node-id=2094-7031&mode=design&t=UXAJ8MkmcdI5ik8a-4

@aweell maybe it is not too clear in specs, two teams overlooked it.

yceballost (changes requested):

yceballost:
> **Screenshot tests report**
> 
> ❌ poster-card / poster-card-with-video
> 

I had the same problem yesterday with that screenshot 🤔

src/__stories__/poster-card-story.tsx
atabel: hmmm... how do you know you can use a color constant name there?, that's not intuitive at all

atabel: I've tried something different, please take a look now

src/card.tsx
atabel: I think it should be fixed now

atabel:
> > **Screenshot tests report**
> > x poster-card / poster-card-with-video
> > 
> 
> I had the same problem yesterday with that screenshot thinking

This test seems unstable

src/__stories__/poster-card-story.tsx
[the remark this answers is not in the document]
atabel: fixed

src/card.tsx
aweell: The background now work fine, but the top actions states also need to change depending if the background is media or no media variant="inverse", no media variant="default/alternative".

Screenshot 2023-08-11 at 11 19 48

src/__type_tests__/card-type-test.tsx
```diff
@@ -31,6 +31,11 @@ import Image from '../image';
     onPress={() => {}}
     trackingEvent={{name: 'do-something'}}
 />;
+<PosterCard title="title" isInverse />;
+<PosterCard title="title" variant="inverse" />;
+<PosterCard title="title" variant="alternative" />;
+<PosterCard title="title" isInverse backgroundColor="red" />;
+<PosterCard title="title" variant="inverse" backgroundColor="red" />;
```
atabel, line 38: these are valid usage examples

src/__stories__/poster-card-story.tsx
aweell: LGTM

src/card.tsx
atabel: fixed

aweell (approved):
LGTM

src/__stories__/poster-card-story.tsx
yceballost: That's perfect. To be picky maybe we could filter tokens with "text" to avoid that appearing in the list 🌝

src/__screenshot_tests__/__image_snapshots__/poster-card-screenshot-test-tsx-poster-card-with-custom-background-color-1-snap.png
yceballost, line 1: This card is impossible to have in dark mode. ThemeVariant is forcing these text to white :( I think that could be a problem. Can we force themeVariant also in dark mode?
image

src/__screenshot_tests__/__image_snapshots__/poster-card-screenshot-test-tsx-poster-card-with-custom-background-color-inverse-1-snap.png
yceballost, line 1: Add playroom snippet with a poster card with `promo` background for example

yceballost (changes requested):

src/__screenshot_tests__/__image_snapshots__/poster-card-screenshot-test-tsx-poster-card-with-custom-background-color-inverse-1-snap.png
atabel: done

src/__stories__/poster-card-story.tsx
atabel: that would be tricky, which colors should we exclude/include?

src/__screenshot_tests__/__image_snapshots__/poster-card-screenshot-test-tsx-poster-card-with-custom-background-color-1-snap.png
atabel: This is tricky because Text just uses textPrimary/textPrimaryInverse depending on the variant context, but those color constants are both white in dark mode. So the only way to fix this would be to find a way to disable dark mode in this area. We could explore a solution in a different PR to manage this case

src/__type_tests__/card-type-test.tsx
```diff
@@ -42,8 +47,14 @@ import Image from '../image';
 <PosterCard title="title" trackingEvent={{name: 'do-something'}} />;
 // @ts-expect-error backgroundImage and backgroundVideo can't be used together
 <PosterCard title="title" backgroundImage="" backgroundVideo="" />;
-// @ts-expect-error backgroundImage or backgroundVideo are mandatory
+// @ts-expect-error backgroundImage and backgroundColor can't be used together
+<PosterCard title="title" backgroundImage="" backgroundColor="" />;
+// @ts-expect-error backgroundVideo and backgroundColor can't be used together
+<PosterCard title="title" backgroundVideo="" backgroundColor="" />;
+// @ts-expect-error backtroundColor, backgroundImage or backgroundVideo are mandatory
```
marcoskolodny, line 54: nit: backtroundColor -> backgroundColor

atabel: fixed

src/__type_tests__/card-type-test.tsx
```diff
@@ -42,8 +47,14 @@ import Image from '../image';
...
 <PosterCard title="title" />;
+// @ts-expect-error if you set a custom backgroundColor, you should specify the variant
```
marcoskolodny, line 56: Why not using a default value for the variant if it's not provided?

marcoskolodny (approved):

yceballost (approved):

src/__screenshot_tests__/__image_snapshots__/poster-card-screenshot-test-tsx-poster-card-with-custom-background-color-1-snap.png
atabel: https://jira.tid.es/browse/WEB-1502

src/__type_tests__/card-type-test.tsx
atabel: I think I prefer this to be explicit. You set either `backgroundImage`, `backgroundColor` or `backgroundVideo`. `backgroundImage` can be set explicitly with a custom color (`backgroundColor="red"`) or implicitly using an specific variant (`variant="inverse"` or `isInverse`)

PR #844, opened 2023-08-07
PosterCard-background

BrunoGuimaraes103:
[https://www.figma.com/file/tKdPOfcUALzVIh5oizFbm7/🔸-Cards-Specs?type=design&node-id=1868%3A3853&mode=design&t=fpLlSjW8QU44yIDg-1](https://www.figma.com/file/tKdPOfcUALzVIh5oizFbm7/%F0%9F%94%B8-Cards-Specs?type=design&node-id=1868%3A3853&mode=design&t=fpLlSjW8QU44yIDg-1)


daviteixeira-tlf:
It would be a good idea to add a snippet that uses the backgroundColor for those who are going to use the playroom.

daviteixeira-tlf:
In the storybook, it's better to move up the backgroundColor to be below the background, so when you see the two properties
being one below the other, in addition it adds an option of none in the background, so when selecting none it takes the color of the
backgroundColor, and adds the color backgroundBrand as the default backgroundColor in the storybook.

EduardoLafuente (approved):

daviteixeira-tlf (approved):

daviteixeira-tlf:
@atabel @aweell

src/card.tsx
yceballost, line 1: this usage is wrong
https://tinyurl.com/24fsdre5


the possibility to change text color its only allowed for custom background color

yceballost (changes requested):

src/card.tsx
BrunoGuimaraes103: "the possibility to change text color its only allowed for custom background color"
The text should only change the color if it receives a background color, in the case of an image, should the color be default?

"update: In windows works (or isInverse doesn't works xD) as expected. In Mac OS chrome this is happening"
I did not understand your comment. Can you explain better?

yceballost: > The text should only change the color if it receives a background color, in the case of an image, should the color be default?

Exactly. When the card has an image, the gradient covers the need to make the text visible in most of the images. So we don't need to give the possibility to make color changes in the background image and video cases

> I did not understand your comment. Can you explain better?
Sure! I tried the link from my first message in Windows, and the isInverse prop doesn't works in the same way as in Chrome in Mac OS. Quite weird in my opinion.. i don't know if anybody can check this in other computes with both OS.

In chrome window 10, isInverse prop always paint the text in white (with true and false value) and in chrome in Mac OS, the prop change color as I show in the image 🤔

Do I make myself clear?

src/card.tsx
```diff
@@ -1275,7 +1279,10 @@ export const PosterCard = React.forwardRef<HTMLDivElement, PosterCardProps>(
                     >
                         {isTouchable && <div className={styles.touchableCardOverlayMedia} />}
```
aweell, line 1280: When the card has an image or video defined should use the `touchableCardOverlayMedia` style, but when the card has no media defined or a custom color defined, depending on if it's inverse or not should use `touchableCardOverlayInverse` / `touchableCardOverlay`

Screenshot 2023-08-10 at 09 02 18

aweell (changes requested):

src/card.tsx
BrunoGuimaraes103: I did not understand your comment. Could you explain it better?

aweell: Of course, there are three cases:

* **The card has an image or video defined:** The overlay that handle the `:hover` and `pressed` styles should use `touchableCardOverlayMedia` that makes the layer to have a black with 25% opacity when hovered and 35% when pressed, that works in most images since we don't know how dark/light they are.
* **The card has no image or video defined and is inverse={false}:** The overlay that handle the `:hover` and `pressed` styles should use `touchableCardOverlay`. Since the background will be white (the default) or a light color, the styles for the overlay are more subtle.
* **The card has no image or video defined and is inverse={true}:** The overlay that handle the `:hover` and `pressed` styles should use `touchableCardOverlayInverse`. Since the background will be brand (the default) or a dark color, the styles for the overlay should be a little bit stronger in order to be visible.

The same applies to the top actions container (image posted in the previous comment), depending of the case the container will have different default an interaction states.

* **The card has an image or video defined:** The actions container by default is `inverse `at 70%, and we change the opacity for the :hover (90%) and `pressed` (100%) states.
* **The card has no image or video defined and is inverse={false}:** The actions container by default is transparent and uses the tokens `backgroundContainerHover` & `backgroundContainerPressed` for the interaction states.
* **The card has no image or video defined and is inverse={true}:** The actions container by default is transparent and uses the tokens `backgroundContainerBrandHover` & `backgroundContainerBrandPressed` for the interaction states

atabel (commented):
Hi @BrunoGuimaraes103, we have a duplicated PR which implements this same feature in https://github.com/Telefonica/mistica-web/pull/841

Given this PR needs some fixes/changes, I think it would be more effective to merge #841 instead of dedicating effort to fix this. You can take a look to #841 if you want to see if it covers your needs.

Sorry for the confusion, we didn't know you were working in the same task :(

PR #849, opened 2023-08-10
feat(ButtonLink): scale chevron according to font size

marcoskolodny:
Font size = 10
image

Font size = 24
image

Font size = 72
image


yceballost (approved):

src/button.tsx
```diff
@@ -81,7 +81,7 @@ const renderButtonElement = ({
 
 const ButtonLinkChevron: React.FC = () => {
     return (
-        <svg width="8" height="20" viewBox="0 0 8 20">
+        <svg width="0.5em" height="1.25em" viewBox="0 0 8 20">
```
atabel, line 84: please use `pxToRem()` function

src/__screenshot_tests__/__image_snapshots__/title-screenshot-test-mobile_android-with-link-and-long-text-snap.png
atabel, line 1: hmmm... this is weird, the size shouldn't change in the default case

marcoskolodny: Actually, it's the expected behaviour: in mobile, Text2 has fontSize equal to 14px, not 16px

src/button.tsx
marcoskolodny: Discussed offline. Using `rem` is not correct. We want the chevron to adapt to the button text's font size, and this size changes between devices (14px in mobile, 16 in desktop). For this reason, it's better to leave it with `em`.

PR #855, opened 2023-08-17
fix(Community): export community components with namespace

pladaria:
This workarounds issues with next12 and commonjs modules. Components imported from from community.js are not able to access the Theme context, and community.js internally uses ES imports which causes problems in next build.

Consider this a temporary workaround that will be removed in the next mistica major.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -204,9 +204,11 @@ if (process.env.NODE_ENV !== 'production' && isClientSide()) {
     }
 }
 
-// Temporary solution to export Community components until mistica gets migrated to ESModules
-// the community.js export has issues because it exports an ES module and next12 interterprets it as a CommonJS module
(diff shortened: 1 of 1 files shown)


EduardoLafuente (approved):

src/index.tsx
```diff
@@ -204,6 +204,20 @@ if (process.env.NODE_ENV !== 'production' && isClientSide()) {
     }
 }
 
+// Temporary solution to export Community components until mistica gets migrated to ESModules
+// the community.js export has issues because it exports an ES module and next12 interterprets it as a CommonJS module
+// importing from /dist/ is not an option because those modules don't get the context from the theme provider
```
AlexJCelmer, line 209: maybe use /* */ to multi line coments

AlexJCelmer (approved):

lucas-rbarbosa (approved):

PR #856, opened 2023-08-18
feat(utilities): Remove lodash dependency and replace it with custom functions

pladaria:
This will help us to make mistica a fully ES Module library. In my tests, when changing the package type to `module`, the build was complaining about `lodash` (which is a cjs library).

I tried to replace it with `lodash-es` but for some reason that I was unable to understand, that produced problems with the SSR acceptance tests, so at the end we decided to fully remove lodash from mistica components (and we saved 2.2 kB)


package.json
```diff
@@ -121,6 +121,7 @@
         "jest-environment-puppeteer": "^6.1.1",
         "jimp": "^0.16.1",
         "lint-staged": "^12.3.7",
+        "lodash": "^4.17.21",
```
pladaria, line 124: moved to devDependencies

src/utils/lodash.tsx
pladaria, line 1: Custom implementation.

`debounce` implements the features that are required by our code. If follows the lodash API but it is incomplete (some options are not implemented).

`isEqual` is missing a lot of corner cases (NaN, typed arrays, buffers...)

.github/actions/audit-accessibility/main.js
```diff
@@ -48,6 +47,7 @@ const startStorybook = () => {
         const port = 6006;
 
         const storybookServer = http.createServer((request, response) => {
+            // @ts-expect-error - type mismatch in response
```
pladaria, line 50: false positive. I've created an issue in js-toolbox to track these problems

https://github.com/Telefonica/js-toolbox/issues/new

src/utils/lodash.tsx
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        debounceTimeoutId = setTimeout(() => {
+            func(...args);
+            if (maxWaitTimeoutId) {
+                clearTimeout(maxWaitTimeoutId);
+            }
+            debounceTimeoutId = undefined;
+            maxWaitTimeoutId = undefined;
+            // eslint-disable-next-line testing-library/await-async-utils
```
pladaria, line 53: false positive. I've created an issue in js-toolbox to track these problems

https://github.com/Telefonica/js-toolbox/issues/new

src/utils/lodash.tsx
```diff
@@ -0,0 +1,116 @@
...
+    return true;
+};
+
+export const isEqual = (a: unknown, b: unknown): boolean => {
+    if (a === b) {
+        return true;
+    }
+
+    if (isPrimitive(a) || isPrimitive(b)) {
+        return a === b;
```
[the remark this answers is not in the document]
pladaria: hmm, you're right. The previous condition is redundant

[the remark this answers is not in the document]
pladaria: Not possible. Changed to `return false`;

src/utils/__tests__/lodash-test.tsx
```diff
@@ -0,0 +1,115 @@
+import {debounce, isEqual} from '../lodash';
+
+beforeEach(() => {
+    jest.useFakeTimers();
+});
+
+describe('debounce', () => {
```
atabel, line 7: why describe?

src/utils/lodash.tsx
atabel: why `lodash.tsx`? I'd just have an utils module, or different modules in this folder if you want, but what's the point of calling it lodash?

pladaria: This module replaces lodash. If after migrating Mistica to ES Module we can use lodash-es, perhaps we should remove this file. I named it "lodash" to make it obvious that this file replaces lodash functions.

If we don't want to use lodash, I can move this to a generic "helpers" or "utils" file

src/utils/__tests__/lodash-test.tsx
pladaria: It is a way to organize the test suite. All `debounce` tests are placed inside the `debounce` describe section, same with `isEqual` or any other method.

According to jest doc:
> `describe()` creates a block that groups together several related tests

And when you run the tests, they are shown with a pretty hierarchy
image

src/utils/lodash.tsx
atabel: I think we should go without lodash, even if lodash-es works in the future

src/utils/__tests__/lodash-test.tsx
atabel: I don't like it, and we don't use it anywhere. Organice the tests by file should be more than enough. And you already have a "debounce" or "isEqual" prefix in all your test names. It adds an extra nesting and can be tricky when combined with before/after methods

atabel (approved):

PR #857, opened 2023-08-21
fix(Tabs): remove extra margin in safari

marcoskolodny:
In Safari, buttons have extra padding/margin by default ([stack overflow post](https://stackoverflow.com/questions/5170041/button-extra-padding-problem-in-safari)). We need to remove this for tabs.

Issue: [Link](https://jira.tid.es/browse/WEB-1510)

### DIFF
--- a/src/tabs.css.ts
+++ b/src/tabs.css.ts
@@ -55,6 +55,10 @@ const baseTab = style([
         background: 'transparent',
     }),
     {
+        /** Setting margin to 0, in order to avoid Safari from automatically adding extra margin to
+         * the touchable container (https://stackoverflow.com/a/71093016)
+         */
+        margin: 0,
         textAlign: 'center',
         verticalAlign: 'baseline',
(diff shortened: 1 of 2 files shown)


yceballost (approved):

src/tabs.css.ts
```diff
@@ -55,6 +55,10 @@ const baseTab = style([
         background: 'transparent',
     }),
     {
+        /** Setting margin to 0, in order to avoid Safari from automatically adding extra margin to
```
atabel, line 58: this comment is bad formatted

src/tabs.css.ts
atabel, line 1: doesn't this issue affect mistica Buttons too?

src/tabs.css.ts
marcoskolodny: fixed

atabel (approved):

PR #859, opened 2023-08-22
feat(AdvancedDataCard): support for href/to

BrunoGuimaraes103:
PR for Advanced Data Card accessibility testing


src/community/advanced-data-card.tsx
```diff
@@ -289,6 +281,7 @@ export const AdvancedDataCard = React.forwardRef<HTMLDivElement, AdvancedDataCar
                         maybe
                         className={classNames(styles.touchableContainer, {[styles.hoverEffect]: !!onPress})}
                         aria-label={ariaLabel}
+                        role="link"
```
atabel, line 284: why role link?
`Touchable` renders as html `<button>` (role button) when used with `onPress` prop and as html `<a>` (role link) when used with `href`/`to` prop . If you want a link, I think the best approach is adding `href`/`to` support to `AdvancedDataCard` as we have in other cards

EduardoLafuente: Hello Abel,

This PR is still under review of accessibility team, we upload here to generate the preview link for them.
Most grateful for your contribution at this time :)

BrunoGuimaraes103: the role='link' was to be able to adjust the accessibility semantics without changing any properties, but I implemented the href according to your suggestion

src/community/__stories__/advanced-data-card-story.tsx
yceballost, line 1: just curious, my focus of the review is not this kind of technical details but I would like to understand why the replacement of onPress by href and does not support both properties ([like the rest of the cards](https://tinyurl.com/ytevnc68))

src/community/__screenshot_tests__/__image_snapshots__/advanced-data-card-screenshot-test-tsx-advanced-data-card-without-stacking-group-with-top-actions-and-too-long-title-1-snap.png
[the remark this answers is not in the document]
BrunoGuimaraes103: was not expected, but the slightly larger size is due to the touchable being moved into the internal component that encompasses pretitle, title etc.

yceballost (approved):

src/community/advanced-data-card.css.ts
```diff
@@ -38,9 +38,18 @@ export const actions = style([
     },
 ]);
 
+export const touchableAccessibility = style({
```
atabel, line 41: please use a more semantic name

src/community/advanced-data-card.css.ts
```diff
@@ -95,6 +103,8 @@ export const cardContentStyle = style([
     },
 ]);
 
+export const zindex = style({zIndex: -20});
```
atabel, line 106: why -20?

src/community/advanced-data-card.tsx
```diff
@@ -57,46 +59,51 @@ const CardContent: React.FC<CardContentProps> = ({
...
-                </Text>
+            <Touchable
+                tabIndex={0}
+                maybe
+                className={classNames(styles.touchableAccessibility)}
+                as="a"
+                href={href}
+                aria-label={ariaLabel}
+            >
+                <div />
```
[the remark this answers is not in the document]
BrunoGuimaraes103: The empty div is because the touchable must have a child, and the reason the touchable is there is so the touchable does not interfere with the screen reader's reading order.

The touchable is as a link, because it should be read as such on the screen reader.

src/community/advanced-data-card.css.ts
BrunoGuimaraes103: The negative z-index is so that images in slots (like SimpleBlock) don't overlap the Touchable

src/community/advanced-data-card.tsx
```diff
@@ -201,7 +208,7 @@ type AllowedExtra =
 type TextAs = 'h1' | 'h2' | 'h3' | 'h4';
 
 interface AdvancedDataCardProps {
-    onPress?: () => void;
+    href?: string | undefined;
```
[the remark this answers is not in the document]
BrunoGuimaraes103: The change from onPress to href is so that on the screen reader it can be read as a link instead of a button.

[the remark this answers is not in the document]
EduardoLafuente: Hi @atabel,
Is something pending on our side?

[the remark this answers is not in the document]
yceballost: We can't make a major version by this breaking change, we need to support `onPress` + `href`, and we also can include `to` to have parity with the rest of the cards. But we need to keep onPress because in master is already existing and there are teams using this prop.

@EduardoLafuente @BrunoGuimaraes103

src/community/advanced-data-card.css.ts
atabel: this number doesn't make sense to me. I'm sure you can reorder the elements, use different stacking contexts or using a more sensible zIndex value than -20

src/community/advanced-data-card.tsx
```diff
@@ -261,7 +290,7 @@ export const AdvancedDataCard = React.forwardRef<HTMLDivElement, AdvancedDataCar
         },
         ref
     ) => {
-        const isTouchable = !!onPress;
+        const isTouchable = !!href || !!onPress;
```
atabel, line 293: what about `to` ?

atabel (changes requested):

src/community/advanced-data-card.tsx
```diff
@@ -200,8 +226,10 @@ type AllowedExtra =
 
 type TextAs = 'h1' | 'h2' | 'h3' | 'h4';
 
-interface AdvancedDataCardProps {
+type AdvancedDataCardProps = {
     onPress?: () => void;
+    href?: string | undefined;
+    to?: undefined;
```
atabel, line 232: remove to/href/onPress from common props, you are already defining them in the ExclusifyUnion

src/community/advanced-data-card.tsx
```diff
@@ -43,9 +46,13 @@ type CardContentProps = {
     subtitleLinesMax?: number;
     description?: string;
     descriptionLinesMax?: number;
+    href?: string;
+    to?: undefined;
```
atabel, line 50: to shouldn't be undefined here.
You should use ExclusifyUnion here too, like you do in `AdvancedDataCardProps`

src/community/advanced-data-card.tsx
```diff
@@ -57,46 +64,65 @@ const CardContent: React.FC<CardContentProps> = ({
...
+                    className={classNames(styles.touchableArea)}
+                    onPress={onPress}
+                    to={to}
+                    aria-label={ariaLabel}
                 >
-                    {title}
-                </Text>
+                    <div />
+                </Touchable>
+            )}
```
atabel, line 98: you can simplify this by doing something like this: https://github.com/Telefonica/mistica-web/blob/master/src/card.tsx#L511

image

atabel (approved):

PR #862, opened 2023-08-28
fix(ProgressBlock): fix undefined check when progressPercent is 0

marcoskolodny:
Playroom snippet showing the issue: [Link](https://mistica-web.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8AFAThA5mmBnXAQgDYRgDWAOgHYAENALgJb1EwC8FITLMn1dACxgBDKIyqY2wYPzo0AbsKIBXGPBqcATAAYaAcQKcANLLr0YAD3rrO5q8dkBfR7IAOGbHlwoYaMDCp6KW0XWhpYXDA0RlcmCCoOEAiomLiqPioAegA%2BahAjLiEAWzwEAG0QYmFlfJAAeU0QAF0CgHdGKHoBXHKAZgA2bSbHIA)

### DIFF
--- a/src/community/blocks.tsx
+++ b/src/community/blocks.tsx
@@ -250,7 +250,9 @@ export const ProgressBlock: React.FC<ProgressBlockProps> = ({
                     </Box>
                     {stackingGroup}
                 </Inline>
-                {progressPercent && <ProgressBar progressPercent={progressPercent} reverse={reverse} />}
+                {progressPercent !== undefined && (
+                    <ProgressBar progressPercent={progressPercent} reverse={reverse} />
+                )}
                 <Inline space={8} alignItems="baseline">
                     <Text8 color={heading.valueColor || vars.colors.textPrimary}>{heading.value}</Text8>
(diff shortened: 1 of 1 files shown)


src/community/blocks.tsx
```diff
@@ -250,7 +250,9 @@ export const ProgressBlock: React.FC<ProgressBlockProps> = ({
                     </Box>
                     {stackingGroup}
                 </Inline>
-                {progressPercent && <ProgressBar progressPercent={progressPercent} reverse={reverse} />}
+                {progressPercent !== undefined && (
```
pladaria, line 253: Consider:
`Number.isFinite(progressPercent) && (...)`

marcoskolodny: Discussed offline. We are leaving it as it is for simplicity, because TS type refinement is not working well with `Number.isFinite` and we believe it's not worth it to add extra checks only for this scenario.

pladaria (approved):

Issue #842, opened 2023-08-03
Form component - loading behavior on submit


CamilaQueiroz:
Using the **Form** component, when passing a function to the **onSubmit** prop, the form keeps loading without stopping. I don't know if it's related, but I've monitored the formStatus prop of useForm and it doesn't change from filling to sending when I submit the form. **I'm using Next.js 13**

**To Reproduce** Steps to reproduce the behavior:

1. Use Form component like the example below

```

function Login() {
  const { SignIn } = useAuth();
  const router = useRouter();

  const onSubmit = async (data: FormValues) => {
    try {
      await SignIn(data as SignInProps);
      router.push("/dashboard");

    } catch (error) {
      alert("Credenciais inválidas");
    }
  };

  return (
    <>
      <Form
        onSubmit={onSubmit}
        id="login-form"
        initialValues={{ email: "", password: "" }}
      >
        <Stack space={16}>
          <EmailField label="e-mail" name="email" id="email" />

          <PasswordField label="senha" id="password" name="password" />

          <ButtonLayout>
            <ButtonPrimary submit loadingText="Sending">
              Login
            </ButtonPrimary>
          </ButtonLayout>
        </Stack>
      </Form>
    </>
  );
}





```
**Expected behavior**  Form component change the "loading" behavior when finished the function

<img width="442" alt="Captura de Tela 2023-08-03 às 16 17 08" src="https://github.com/Telefonica/mistica-web/assets/20759520/4fed90b3-c54c-479f-bc1b-3f44ccb5aeb0">

**Desktop (please complete the following information):**

- OS: MacOS Ventura 13.2
- Browser: chrome, safari

CamilaQueiroz:
Looking for more information, I understood that this is the behavior of the Form component. I found this explanation in an example:

 - When the form is being submitted (while the onSubmit promise is unresolved or rejected), all fields are automatically disabled and a spinner appears inside the submit button.

https://github.com/Telefonica/mistica-web/blob/master/doc/forms.md#important-notes


But is there any way to handle this behavior? I tried to manipulate the props "disable" and "showSpinner" but no success.

atabel:
this is the expected behavior. If you want the form to be usable again you need to resolve the onSubmit promise.
See this example:
[...]

PR #863, opened 2023-08-29
fix(Form): set mountedRef value to true on client side first render

marcoskolodny:
Solving issue reported [here](https://github.com/Telefonica/mistica-web/issues/842).

When using React with StrictMode on, in dev environment components' effects are executed twice (https://react.dev/reference/react/StrictMode#strictmode)

After mounting the component and executing the effects once, the `isMountedRef` value is set to false, but it's never reset to true in the second time the effect is triggered. This causes the form to never update its state to "filling" after performing a form submit:

image

### DIFF
--- a/src/form.tsx
+++ b/src/form.tsx
@@ -48,6 +48,12 @@ const Form: React.FC<FormProps> = ({
     const id = useAriaId(idProp);
 
     React.useEffect(() => {
+        /**
+         * When using React with Strict Mode on, the components are rendered twice. If we don't set the ref's value to true
+         * every time the component is mounted, during the second mount this value will be equal to false, preventing handleSubmit()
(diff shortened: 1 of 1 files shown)


src/form.tsx
```diff
@@ -47,12 +47,12 @@ const Form: React.FC<FormProps> = ({
     const {texts} = useTheme();
     const id = useAriaId(idProp);
 
-    React.useEffect(
-        () => () => {
+    React.useEffect(() => {
+        isMountedRef.current = true;
```
pladaria, line 51: please add a short comment explaining why this assignment is necessary

pladaria (approved):

src/form.tsx
marcoskolodny: done

PR #872, opened 2023-09-04
fix(Inline): avoid content overflow with wrap and negative space

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-816)


yceballost (approved):

src/inline.css.ts
atabel, line 1: Any idea why this was only broken for ios theme?

atabel (approved):

PR #877, opened 2023-09-07
feat(getCssVarValue): utility function

atabel:
WEB-1518


src/utils/dom.tsx
```diff
@@ -121,3 +121,32 @@ export const listenResize = (element: Element, handler: ResizeListener): (() =>
...
+
+const parseVarCall = (cssVar: string) => {
+    const [varName, fallback] = cssVar
+        .replace(/^var\(/, '')
+        .replace(/\)$/, '')
+        .split(/\s*,\s*(.*)/);
+    return {varName, fallback};
+};
+
+const isCssVar = (cssVar: string) => cssVar.startsWith('var(') || cssVar.startsWith('--');
```
marcoskolodny, line 133: Should we consider handling cases in which this returns true, but it's actually not a valid css var? (something like `(varrrr` for example).

atabel: I think this is good enough. But I don't understand your example, `(varrrr` doesn't start with `var(`

pladaria (approved):

src/utils/dom.tsx
marcoskolodny: Sorry, I meant something line `var(((`.

atabel: I think it doesn't make sense to cover that case, this function will probably be used with mistica skinVars in most of the cases, so we won't have these weird cases

PR #883, opened 2023-09-13
feat(ErrorFeedback): new component

atabel:
[WEB-1533](https://jira.tid.es/browse/WEB-1533)

Figma: https://www.figma.com/file/LpjgnU1xpmEzQvlXKwAHmn/%F0%9F%94%B8-Feedbacks-Specs?type=design&node-id=2202-130&mode=design&t=rYvgUXRoXd5BNLwo-0


src/feedback.tsx
```diff
@@ -459,3 +448,52 @@ export const SuccessFeedback: React.FC<AssetFeedbackProps> = ({
         })
     );
 };
+
+export const ErrorFeedback: React.FC<ErrorFeedbackScreenProps> = ({
+    title,
+    description,
+    primaryButton,
```
Winde, line 455: I believe the icon would need to be parametrizable if we are to keep the current VIVO "customization":

```
platform.getAppBrand() === 'VivoBR' ? (
                    <IconWarningLight color={skinVars.colors.error} size={60} />
                ) : (
                    <IconFileErrorLight color={skinVars.colors.brand} size={60} />
                )
```

atabel: hmmm... I don't think so. ErrorFeedbackScreen doesn't allow you to configure the error icon, I don't se why ErrorFeedback should allow it

atabel: we could expose some kind of generic customizable `Feedback` as we do with `FeedbackScreen` if it's really needed

Winde: if the aim is to solve the current issue I raised, yes, we would need that `Feedback`

if you look at web/src/common/components/full-page-error in webapp, we are using FeedbackScreen instead of ErrorFeedbackScreen because of the fact that the icon needs to be configurable

atabel: I see. It makes sense, but I'm a bit worried we could be over abusing this configurability, we should try to follow the standard ErrorFeedback/ErrorFeedbackScreen design whenever possible unless there is a strong reason. That's the main point of having a Design System. @aweell, what's Design Core opinion on this?

marcoskolodny (approved):

src/feedback.tsx
aweell: We're worried too we're not following the definition correctly, if we're going to display an error the icon should be consistent. So using errorFeedback / errorFeedbackScreen should be the way to go. 

I'm digging into this need with product design. @winde if you have examples of these cases at hand and can share them, it would be very valuable.

Winde: Here's a storybook where it can be seen the icon changing per brand: https://storybook.tuenti.io/?path=/story/account-common-components-fullpageerror--default

Here's the tickets of the last update of this component in webapp:
https://jira.tid.es/browse/PRODUCTDSN-1246
https://jira.tid.es/browse/ACCOUNT-21143

I'm asking @montse for more info on why it was determined a different icon for VIVO.

pladaria (approved):

PR #888, opened 2023-09-19
chore(utilities): update debounce implementation and tests

marcoskolodny:


src/utils/helpers.tsx
```diff
@@ -1,59 +1,121 @@
...
-
+type Debounced<T> = T & {cancel: () => void; flush: () => void};
+
+/**
+ * Creates a debounced function that delays invoking func until after wait milliseconds
+ * have elapsed since the last time the debounced function was invoked. The debounced function
+ * comes with a cancel method to cancel delayed func invocations and a flush method to immediately
+ * invoke them. The func is invoked with the last arguments provided to the debounced function.
+ * Subsequent calls to the debounced function return the result of the last func invocation.
+ */
```
pladaria, line 9: please add a note commenting that these functions are copied from webapp's package

pladaria (approved):

src/utils/helpers.tsx
marcoskolodny: done

PR #891, opened 2023-09-20
fix(FixedFooterLayout): prevent content from being rendered on top of the footer

marcoskolodny:
Issue: [link](https://jira.tid.es/browse/WEB-1548)

### DIFF
--- a/src/fixed-footer-layout.css.ts
+++ b/src/fixed-footer-layout.css.ts
@@ -29,15 +29,6 @@ export const fixedFooter = style({
     },
 });
 
-export const contentWithFixedFooter = style({
-    '@media': {
-        [mq.tabletOrSmaller]: {
-            position: 'relative',
-            zIndex: 0,
-        },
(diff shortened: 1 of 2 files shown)


src/fixed-footer-layout.css.ts
```diff
@@ -29,6 +29,15 @@ export const fixedFooter = style({
     },
 });
 
+export const contentWithFixedFooter = style({
```
atabel, line 32: alternative fix: make container be a new stacking context with `isolation: isolate`, this way zindex hierarchy inside container won't affect to the fixed footer.

It could be a good idea to apply `isolation: isolate` to cards too.

marcoskolodny: Updated footer in order to use isolation instead of zIndex

aweell (approved):
LGTM!

atabel (approved):

PR #892, opened 2023-09-21
feat(Menu): fix component styles and add MenuItem/MenuSection

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1537)
Specs: [Link](https://www.figma.com/file/a1pgrI7Wg41iyXWU8ETqlU/%F0%9F%94%B8-Dropdown-Menu-Specs-(web)?type=design&node-id=36-1496&mode=design&t=0zm1weXnP7EzSuSK-0)


src/menu.css.ts
```diff
@@ -52,3 +60,29 @@ export const hideItems = style({
     opacity: 0,
     transform: 'scale(0)',
 });
+
+export const menuItem = style({
+    userSelect: 'none',
+    transition: 'background-color 0.15s ease-in-out',
```
yceballost, line 66: missing borderRadius → `borderRadii.popup`

image

src/__stories__/menu-story.tsx
yceballost, line 1: I would like the possibility to see a custom menu without menuItem

src/menu.css.ts
```diff
@@ -31,15 +37,17 @@ export const menuContainer = style([
         top,
         bottom,
         left,
+        right,
         width,
```
yceballost, line 41: Should we add a prop to control the menu height? 🤔

playroom/snippets.tsx
```diff
@@ -8,44 +8,45 @@ const menuSnippet = {
...
-          <Inline space={16}>
-            <IconKebabMenuLight />
-            <Text3 regular>{isMenuOpen ? "Close" : "Open"}</Text3>
-          </Inline>
-        </Touchable>
+        <Box padding={16}>
+          <Touchable
+            ref={ref}
+            onPress={onPress}
+            style={{ maxWidth: "fit-content" }}
```
yceballost, line 16: why this inline style? 🤔

src/menu.tsx
yceballost, line 1: `<MenuItem text={"option 1"} />` without onPress has interaction states.

I think this line should not have interactive states.

yceballost, line 1: When I use the tab key on the keyboard to move around the component, it would be nice if, after the first space to open the menu, the focused item would be the first one. Is it possible?

src/__stories__/menu-story.tsx
yceballost, line 1: The open label in story example breaks with big sizes

image

yceballost (changes requested):

src/menu.css.ts
```diff
@@ -52,3 +60,29 @@ export const hideItems = style({
     opacity: 0,
     transform: 'scale(0)',
 });
+
+export const menuItem = style({
+    userSelect: 'none',
+    transition: 'background-color 0.15s ease-in-out',
+    '@media': {
+        [mq.supportsHover]: {
```
yceballost, line 68: use the same transitions as lists for hover and active states
https://github.com/Telefonica/mistica-web/blob/fcbb4a543545c21df10ccbc9537f3b5d81279b37/src/list.css.ts#L13C13-L13C61

yceballost (changes requested):

src/menu.css.ts
marcoskolodny: fixed

src/menu.css.ts
marcoskolodny: After discussion offline, we may think about this for a second iteration

playroom/snippets.tsx
marcoskolodny: Discussed offline, it's because we don't have css classes in playroom

src/menu.tsx
marcoskolodny: OnPress should not be optional. Fixed

src/__stories__/menu-story.tsx
marcoskolodny: fixed

marcoskolodny: Discussed offline. This is already allowed by passing any random thing inside renderMenu prop, but we decided to avoid showing that in stories to avoid people from using this component in weird ways.

If we notice that people start asking about this, we can add it to stories in a new iteration

src/menu.tsx
marcoskolodny: Fixed. Imitating the behavior of Radix UI's component (https://www.radix-ui.com/primitives/docs/components/dropdown-menu#dropdown-menu)

src/menu.tsx
```diff
@@ -1,18 +1,155 @@
...
+    closeMenu: () => {},
+});
+export const useMenuContext = (): MenuContextType => React.useContext(MenuContext);
+
+interface MenuItemBaseProps {
+    label: string;
+    Icon?: React.FC<IconProps>;
+    destructive?: boolean;
+    disabled?: boolean;
+    ariaLabel?: string;
```
yceballost, line 41:
```suggestion
    'aria-label'?: string;
```
??

marcoskolodny: good catch. Fixed

yceballost (approved):
very nice currazo

src/menu.tsx
```diff
@@ -1,18 +1,155 @@
...
+            {renderContent()}
+        </div>
+    );
+};
+
+type MenuSectionProps = {
+    children?: React.ReactNode;
+};
+
+export const MenuSection: React.FC<MenuSectionProps> = ({children}) => {
```
pladaria, line 140: perhaps we could skip the render when children is empty

src/menu.tsx
pladaria, line 1: The target seems to lose the focus under some circumstances:

1. focus the target
2. press space
3. press down to move to other item
4. press space (the menu closes)
5. press space again (I would expect to be able to open the menu again but it doesn't)

src/menu.tsx
marcoskolodny: Added the null check. Anyways, I believe that scenario is a bit weird and also not a very good usage of the component.

src/menu.tsx
marcoskolodny: Good catch. It should be fixed now :)

pladaria (approved):

aweell (commented):
A subte detail: the menuItem height defined by the specs is 48px, when there's no checkbox the menuItem height is 52px, and when there's checkbox, the height is 51px.

src/menu.tsx
```diff
@@ -1,18 +1,159 @@
...
+    setFocusedValue: (value: string | null) => void;
+    closeMenu: () => void;
+};
+const MenuContext = React.createContext<MenuContextType>({
+    focusedValue: null,
+    isMenuOpen: false,
+    setFocusedValue: () => {},
+    closeMenu: () => {},
+});
+export const useMenuContext = (): MenuContextType => React.useContext(MenuContext);
```
atabel, line 36: is this used outside, needs to be exported?

src/menu.tsx
```diff
@@ -1,18 +1,159 @@
...
+}
+
+type MenuItemProps = MenuItemWithControlProps | MenuItemWithoutControlProps;
+
+export const MenuItem: React.FC<MenuItemProps> = ({
+    label,
+    Icon,
+    destructive,
+    disabled,
+    'aria-label': ariaLabel,
```
atabel, line 65: does it make sense to have an aria-label != label ? do you have any example?

src/menu.tsx
```diff
@@ -1,18 +1,159 @@
...
+            <Touchable
+                onPress={() => {
+                    if (isMenuOpen) {
+                        onPress(label);
+                        closeMenu();
+                    }
+                }}
+                disabled={disabled}
+                role="menuitem"
+                aria-label={label}
```
atabel, line 111: this is redundant

src/__stories__/menu-story.tsx
atabel, line 1: I've found a bug here, but I'm not sure if the bug is in the Menu or Modal component:
- Use the keyboard to open the menu
- select an item with the arrow keys and enter/space.
- The Modal is open: click enter again
    - Expected: the modal is closed
    - Bug: the menu is open again. It seems like the focus is inside the menu instead of being in the Modal

src/menu.tsx
marcoskolodny: This was a request from vivo. They want to be able to choose their own aria labels

marcoskolodny:
> A subte detail: the menuItem height defined by the specs is 48px, when there's no checkbox the menuItem height is 52px, and when there's checkbox, the height is 51px.
> 
> EDIT: It's only when the icon is present

@aweell it should be fixed now :)

src/overlay.tsx
```diff
@@ -55,7 +54,6 @@ const Overlay: React.FC<Props> = ({
                 // In Android with children we hide children and onPress later in onClick to ensure click event doesn't hit element below overlay.
```
atabel, line 54: does this comment still apply

aweell (commented):
Should the touchable / button allow `aria-haspopup`?

> Any button that open a menu should have a role of [button](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role) or, preferably, be a [<button>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button), and also have aria-haspopup="menu" (or true) set.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-haspopup

src/__stories__/menu-story.tsx
```diff
@@ -55,50 +72,50 @@ export const Default: StoryComponent<MenuArgs> = ({
                 >
                     <Menu
                         position={horizontalPosition}
-                        width={320}
+                        width={280}
                         renderTarget={({ref, onPress, isMenuOpen}) => (
                             <Touchable
                                 ref={ref}
```
aweell, line 78:
```suggestion
                                ref={ref}
                                aria-expanded={isMenuOpen}
```

src/overlay.tsx
marcoskolodny: Nope. Good catch

src/menu.tsx
```diff
@@ -1,18 +1,159 @@
...
+    controlType?: undefined;
+    checked?: undefined;
+}
+
+interface MenuItemWithControlProps extends MenuItemBaseProps {
+    controlType: 'checkbox';
+    checked?: boolean;
+}
+
+type MenuItemProps = MenuItemWithControlProps | MenuItemWithoutControlProps;
```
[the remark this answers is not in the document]
marcoskolodny: fixed

src/__stories__/menu-story.tsx
marcoskolodny: fixed

src/menu.css.ts
```diff
@@ -31,24 +37,64 @@ export const menuContainer = style([
...
-    opacity: 1,
-    transform: 'scale(1)',
+export const menuTransitionClasses = {
+    enter: style({
+        opacity: 0,
+        transform: 'scale(0)',
+    }),
+    enterActive: style({
+        opacity: 1,
+        transform: 'scale(1)',
```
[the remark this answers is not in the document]
marcoskolodny: No, you can't. This is actually the styles that will be applied with transition when the element is entering. If you remove this, the transition when opening the menu won't be animated.

From [React Transition Group documentation](https://reactcommunity.org/react-transition-group/css-transition):

Screenshot from 2023-10-03 15-22-36

src/menu.tsx
```diff
@@ -1,18 +1,159 @@
...
+});
+export const useMenuContext = (): MenuContextType => React.useContext(MenuContext);
+
+interface MenuItemBaseProps {
+    label: string;
+    Icon?: React.FC<IconProps>;
+    destructive?: boolean;
+    disabled?: boolean;
+    'aria-label'?: string;
+    children?: undefined;
```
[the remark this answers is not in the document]
marcoskolodny: fixed

src/menu.tsx
marcoskolodny: removed aria-label as prop

src/__stories__/menu-story.tsx
```diff
@@ -55,50 +72,50 @@ export const Default: StoryComponent<MenuArgs> = ({
...
-                                        }}
-                                        checked={valuesState.includes('closing')}
-                                    >
-                                        Click to close the menu
-                                    </Checkbox>
-                                </Box>
+                                            {...getMenuItemControlProps(
+                                                checkbox,
+                                                valuesState.includes(`Option ${optionIndex + 1}`)
+                                            )}
```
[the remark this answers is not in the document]
marcoskolodny: done

src/__stories__/menu-story.tsx
```diff
@@ -55,50 +72,50 @@ export const Default: StoryComponent<MenuArgs> = ({
...
-                                <Inline space={16}>
+                                <Inline space={16} alignItems="center">
                                     <IconKebabMenuLight />
                                     <Text3 regular>{isMenuOpen ? 'Close' : 'Open'}</Text3>
                                 </Inline>
                             </Touchable>
                         )}
-                        renderMenu={({ref, className, close}) => (
+                        renderMenu={({ref, className}) => (
                             <div ref={ref} className={className}>
```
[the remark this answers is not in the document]
marcoskolodny: This has been fixed internally in Menu component

src/__stories__/menu-story.tsx
marcoskolodny: This has been fixed internally in Menu component

marcoskolodny:
> Should the touchable / button allow `aria-haspopup`?
> 
> > Any button that open a menu should have a role of [button](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role) or, preferably, be a , and also have aria-haspopup="menu" (or true) set.
> 
> https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-haspopup

This will be handled when we review accesibility of all the MIstica's components

src/menu.tsx
marcoskolodny: removed

src/menu.tsx
```diff
@@ -61,115 +230,181 @@ const Menu: React.FC<MenuProps> = ({
...
         document.addEventListener('keydown', handleKeyDown, false);
         return () => {
             document.removeEventListener('keydown', handleKeyDown, false);
         };
     });
 
+    React.useEffect(() => {
+        if (menu) {
+            menu.setAttribute('role', 'menu');
+            menu.setAttribute('aria-expanded', String(isMenuOpen));
```
aweell, line 390: `aria-expanded` is being currently added to the menu container, but it should be in the trigger (touchable / button)

marcoskolodny: good catch. Fixed

aweell (approved):
NICEgif

atabel (approved):

src/menu.tsx
```diff
@@ -371,6 +382,11 @@ export const Menu: React.FC<MenuProps> = ({
         };
     });
 
+    React.useEffect(() => {
+        menu?.setAttribute('role', 'menu');
+        target?.setAttribute('aria-expanded', String(isMenuOpen));
```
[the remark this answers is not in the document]
marcoskolodny: Discussed offline. We removed the hack for menu ref, but given the current way to handle menu/target rendering, we can't get rid of the hack for target.

PR #896, opened 2023-09-22
feat(Sheet): multiline support for description

atabel:
jira: WEB-1559


src/sheet-root.tsx
```diff
@@ -105,7 +105,8 @@ const showRadioListNativeSheet = ({
     (nativeSheetImplementation as NativeSheetImplementation)({
         title,
         subtitle,
-        description,
+        // TODO: add multiline support to native sheet
+        description: Array.isArray(description) ? description.join('\n') : description,
```
pladaria, line 109: perhaps, you should join with two line breaks, to simulate different paragraphs

pladaria: also note that:

`[].join('\n')` returns empty string, but in this case we probably want to set an `undefined` description.

atabel: I want to do some tests with the native implementation and see what happens with `\n` and `\n\n`

atabel: Tested. It seems to work quite fine with the double `\n\n`:
image

src/__stories__/sheet-story.tsx
```diff
@@ -82,13 +82,20 @@ type SheetArgs = {
...
 type RadioListSheetArgs = SheetArgs & {
     selectedId: string;
 };
 
-export const RadioList: StoryComponent<RadioListSheetArgs> = ({title, subtitle, description, selectedId}) => {
+export const RadioList: StoryComponent<RadioListSheetArgs> = ({
+    title,
+    subtitle,
+    description,
+    multilineDescription,
```
yceballost, line 96: multiParagraph name maybe fit better? 🤔

multilineDescription could be this:
image

yceballost (approved):

Marcosld (approved):

pladaria (approved):

marcoskolodny (approved):

PR #902, opened 2023-10-02
feat(PinField): new component

atabel:
https://jira.tid.es/browse/WEB-1555

WebOTP API in Chrome Android:

iOS Safari keyboard autocomplete from SMS:


src/otp-field.tsx
```diff
@@ -0,0 +1,309 @@
...
+        if (indexToFocus !== index && indexToFocus <= length - 1) {
+            inputsList[indexToFocus]?.focus();
+        }
+    };
+
+    return (
+        <Inline space={8}>
+            {Array.from({length}).map((_, index) => (
+                <div
+                    key={index}
```
pladaria, line 141: this gets reported as an error in SonarQube :/

src/otp-field.tsx
```diff
@@ -0,0 +1,309 @@
...
+                        // @ts-expect-error: otp is not in the types yet
+                        const code = otp.code.slice(0, length);
+                        changeValue(code);
+                    }
+                })
+                .catch(() => {
+                    // ignore;
+                });
+            return () => {
+                isWaitingForSms = false;
```
pladaria, line 87: this clears the lock when the component unmounts. Perhaps it would be enough to do it on the `credentials` promise `.finally()`

src/otp-field.tsx
atabel: jajaja, I won't do that

src/otp-field.tsx
atabel: hmm... no because I'm aborting it on unmount too. But I could unlock in unmount and in .finally()

src/otp-field.tsx
```diff
@@ -0,0 +1,309 @@
...
+        },
+        [controlledValue, inputsList, isControlledByParent, onChange]
+    );
+
+    React.useEffect(() => {
+        // https://developer.mozilla.org/en-US/docs/Web/API/WebOTP_API
+        if ('OTPCredential' in window && !isWaitingForSms) {
+            isWaitingForSms = true;
+            const abortController = new AbortController();
+            navigator.credentials
```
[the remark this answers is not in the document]
atabel: ok, I see your point, but in the worst case no sms would be received and nothing will heppen, not a big deal.

I can add a prop to optout, but I think listen for sms is a good default. WDYT?

src/otp-field.tsx
```diff
@@ -0,0 +1,309 @@
...
+            newControlledValue = prevChars + toPaste;
+            indexToFocus = Math.min(index + toPaste.length, length - 1);
+        } else {
+            newControlledValue =
+                controlledValue.slice(0, index) + newInputValue + controlledValue.slice(index + 1);
+            indexToFocus = index + 1;
+        }
+
+        // need to flush sync to be commit the new values to the dom before changing the focus
+        flushSync(() => {
```
pladaria, line 129: :O

src/otp-field.tsx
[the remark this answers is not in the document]
pladaria: hmm, it the worst case, a pin input that doesn't expects to be filled by SMS could receive a request because some other service/device is trying to log in. That is a strange experience, although quite corner.

pladaria (approved):
Nice! 

As commented in Teams, I think PinField would be a better name for the component

marcoskolodny:
when the focus is in an already filled position, and you press something different from a digit (a letter for example), the result is the same as when we press tab. Is that expected? I believe it shouldn't jump to the next position in that case

yceballost (approved):

marcoskolodny:
There is a corner case in which the behaviour is a bit weird (maybe it's not even worth fixing):
- Create an OTPField with length equal to 6
- Fill each of its digits
- Change the field length to 4
- Try to delete any digit: the digits that were "removed" when changing the field's length show up again at the end of the field.

marcoskolodny:
I think there is a bug when using controlled values. Using [this snippet](https://mistica-5h8runr3e-tuentisre.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8B5ALgBwGIEsYBsoACAOwEMBbGAXgB0QJN6iyAnHMgWnzICMC6IFABUACs3wwSAczQALasAAsAXyIA3MvgCuNegEYATAGYlzAPQA%2BWiRAAaEPJhUAzggDaIALIR1OF2hsIAC6DgDuOFDybvDuxgBsAAzBKkA), try pressing a digit while the field is focused. If I do that, I see the entire code in the first position. When the focus is removed from the field, it's automatically fixed.

image

src/otp-field.tsx
marcoskolodny, line 1: when the focus is in an already filled position, and you press something different from a digit (a letter for example), the result is the same as when we press tab. Is that expected? I believe it shouldn't jump to the next position in that case

marcoskolodny, line 1: There is a corner case in which the behaviour is a bit weird (maybe it's not even worth fixing):
- Create an OTPField with length equal to 6
- Fill each of its digits
- Change the field length to 4
- Try to delete any digit: the digits that were "removed" when changing the field's length show up again at the end of the field.

marcoskolodny, line 1: I think there is a bug when using controlled values. Using [this snippet](https://mistica-5h8runr3e-tuentisre.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8B5ALgBwGIEsYBsoACAOwEMBbGAXgB0QJN6iyAnHMgWnzICMC6IFABUACs3wwSAczQALasAAsAXyIA3MvgCuNegEYATAGYlzAPQA%2BWiRAAaEPJhUAzggDaIALIR1OF2hsIAC6DgDuOFDybvDuxgBsAAzBKkA), try pressing a digit while the field is focused. If I do that, I see the entire code in the first position. When the focus is removed from the field, it's automatically fixed.

image

atabel: should be fixed now

atabel: I've fixed it. But it's a weird case to force de value and don't allow it to change but let the input be editable

src/otp-field.tsx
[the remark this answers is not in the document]
atabel: As talked in Teams, changed to:

- `<PinField />` visible code, read sms
- `<PinField hideCode />` non-visible code, don't read sms
- `<PinField hideCode readSms />` non-visible code and reads sms

src/__tests__/sheet-test.tsx
```diff
@@ -45,7 +45,7 @@ test('Sheet', async () => {
 
     await userEvent.click(closeButton);
     await waitForElementToBeRemoved(sheet);
-});
+}, 20000);
```
atabel, line 48: unrelated with this PR, but I've seen some of these tests fail with timeout, I've increased the time to mitigate the issue

src/pin-field.css.ts
```diff
@@ -0,0 +1,92 @@
...
+};
+
+export const field = sprinkles({
+    ...fieldCommonStyles,
+    background: vars.colors.backgroundContainer,
+});
+
+export const focusedField = style([
+    sprinkles({
+        ...fieldCommonStyles,
```
yceballost, line 31: I think you need to include `background: vars.colors.backgroundContainer,` in this line to avoid the background changes in focus state (only is visible the problem in dark mode)
image

yceballost (changes requested):

marcoskolodny (approved):

src/pin-field.css.ts
atabel: fixed, good catch!

yceballost (approved):

yceballost:
reminder: remove the placeholder before merging

PR #907, opened 2023-10-06
feat(InformationBlock): fix component style

BrunoGuimaraes103:
adjustment to the secondaryValue property, to include strike

Figma: https://www.figma.com/file/35iS2dwoogulJZPgFGKmCX/Advanced--Data-Card?type=design&node-id=521-104207&mode=design&t=vOgyqyOtcbGeKJwW-0


atabel (approved):

src/community/blocks.tsx
```diff
@@ -118,7 +118,7 @@ export const InformationBlock: React.FC<InformationBlockProps> = ({
         <Inline space="between" alignItems="flex-end" aria-label={ariaLabel}>
             <BlockContent title={title} description={description} />
             <div className={classNames(styles.column, styles.rightContent)}>
-                <Text2 regular color={vars.colors.textSecondary}>
+                <Text2 regular color={vars.colors.textSecondary} decoration="line-through">
```
yceballost, line 121: Can you add Figma specs in the PR Body? I think this decoration has to be configurable

vinrosa: @yceballost This field is supposed to always use the line-through decoration, it's made to show the original price of the item.

yceballost (approved):
talked with @vinrosa. If in the future we need the configuration, we will add it later.

PR #910, opened 2023-10-10
fix(Video, DisplayMediaCard, PosterCard): show fallback on empty src

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1557)

### DIFF
--- a/src/card.tsx
+++ b/src/card.tsx
@@ -186,9 +186,7 @@ const CardContainer = React.forwardRef<HTMLDivElement, CardContainerProps>(
 );
 
 const renderBackgroundImage = (src?: string) => {
-    // Adding '//:0' to force an error in case src is undefined
-    // https://stackoverflow.com/a/5775621
-    return <Image width="100%" height="100%" src={src ?? '//:0'} />;
+    return <Image width="100%" height="100%" src={src || ''} />;
 };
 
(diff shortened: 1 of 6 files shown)


src/card.tsx
```diff
@@ -993,21 +991,25 @@ const DisplayCard = React.forwardRef<HTMLDivElement, GenericDisplayCardProps>(
         },
         ref
     ) => {
+        const hasImage = backgroundImage !== undefined;
+        const hasVideo = backgroundVideo !== undefined;
         const image = renderBackgroundImage(backgroundImage);
         const {video, videoAction} = useVideoWithControls(backgroundVideo, poster, backgroundVideoRef);
 
-        if (backgroundVideo) {
+        console.log(hasImage, hasVideo);
```
Winde, line 999: probably needs to be removed?

Winde (approved):

atabel (approved):
I miss a screenshot test with this case

pladaria (approved):
you could update the test case where an invalid URL is provided to force the fallback to use an empty string (or add a new case to verify the empty string behavior)

https://github.com/Telefonica/mistica-web/blob/45d07bfdc7798e2beebef3700ea014772402e774/src/__stories__/image-story.tsx#L72

PR #912, opened 2023-10-11
feat(Mobile): add interaction states when pressing touchable elements

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1553)


yceballost:
I think we need to change some transitions to make it work on mobile. For example, buttons have a long transition for background color, thats cause the effect doesn't work properly

image
image
image
image

with these changes we standardize this type of transitions (should be a global variable to avoid definition in each file? 🤔)

marcoskolodny:
> I think we need to change some transitions to make it work on mobile. For example, buttons have a long transition for background color, thats cause the effect doesn't work properly
@yceballost fixed :)

yceballost (approved):

pladaria (approved):
change the PR title from `chore` to `fix` or `feat` (you decide).

To me `chore` tasks are for stuff like applying prettier, upgrade dev dependencies, gh workflows...

src/button.css.ts
atabel, line 1: Have you tried all these changes inside the app webviews?

We used to have a bug that caused the active state to be persistent. The scenario was something like this:
- tap a button/link with active state
- that button/link navigates to a new webview (full page load, not spa/client side routing)
- go back to first webview
  - the button keeps active when it shouldn't
  
I think your changes won't reproduce the bug again because I have the impression the bug was caused by hover states, not active. But I'd manually test it just in case

marcoskolodny: Yes, this has been tested in webapp using Android and iOS. In both cases, I interacted with different buttons, rows and cards. All of them worked correctly for me  :)

atabel (approved):

PR #914, opened 2023-10-13
feat(Icons): use icon keywords from mistica design in catalog story

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1552)


packages/import-mistica-icons/index.js
```diff
@@ -174,8 +174,12 @@ const createIconComponentSource = async (name, componentName, svgIconsInfo) => {
      * To update, execute "yarn start" inside "import-mistica-icons"
      */
 
-    import * as React from 'react';
-    ${hasVariants ? "import {useTheme} from '../../hooks';" : ''}
```
marcoskolodny, line 178: this was creating an empty line between imports if `hasVariants` was false, and ESLint was complaining about it

packages/import-mistica-icons/index.js
```diff
@@ -194,6 +198,35 @@ const createIconComponentSource = async (name, componentName, svgIconsInfo) => {
...
+/**
+ * @typedef {{[key: string]: Array<string>}} iconKeywordsInfo
+ */
+const createAllIconKeywordsSource = () => {
+    const keywordsPath = join(PATH_MISTICA_ICONS_REPO, 'icons', 'icons-keywords.json');
+    const keywordsSource = fs.readFileSync(keywordsPath, 'utf8');
+    /** @type {iconKeywordsInfo} */
+    const keywordsMap = JSON.parse(keywordsSource);
+
+    // map icon names to kebab case
```
marcoskolodny, line 210: some icon filenames in Mistica design repo are not following the expected format when using kebab case. For example: `2g-filled` instead of `2-g-filled`.

Given that in Mistica web we generate the filenames using kebab case (check the name of [this file](https://github.com/Telefonica/mistica-web/blob/master/src/generated/mistica-icons/icon-2-g-filled.tsx) for example), I mapped the names in design repo to their corresponding kebab case version when generating the keywords file.

src/icons/__stories__/mistica-icons-story.tsx
```diff
@@ -19,22 +18,39 @@ const misticaIcons = ((requireContext) => {
         component.componentName = fileNameToComponentName(id);
         return component;
     });
-})(require.context('../../generated/mistica-icons/', true, /\.tsx$/));
+})(require.context('../../generated/mistica-icons/', true, /^(?!\.\/icons\-keywords\.tsx$).+\.(?:tsx)$/));
```
marcoskolodny, line 21: Ignore the file containing the keywords (`icons-keywords.tsx`)

packages/import-mistica-icons/index.js
```diff
@@ -194,6 +198,35 @@ const createIconComponentSource = async (name, componentName, svgIconsInfo) => {
     return format(source);
 };
 
+/**
+ * @typedef {{[key: string]: Array<string>}} iconKeywordsInfo
```
pladaria, line 202: use PascalCase for type names (`IconKeywordsInfo`)

pladaria (approved):

packages/import-mistica-icons/index.js
marcoskolodny: good catch. fixed

src/icons/__stories__/mistica-icons-story.tsx
```diff
@@ -87,72 +105,55 @@ export const Catalog = (): JSX.Element => {
...
+        </ResponsiveLayout>
     );
 };
+
+Catalog.args = {
+    filter: '',
+    size: 32,
+    regular: true,
+    light: true,
+    filled: true,
```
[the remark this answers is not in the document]
marcoskolodny: what if we want to see all of them? By using select, we wouldn't be able, right?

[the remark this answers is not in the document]
yceballost: oh good point, yes, makes sense. Keep it with switches

yceballost (approved):

src/icons/__stories__/mistica-icons-story.tsx
atabel: you could have moved the file inside a subfolder to avoid this

atabel (approved):

PR #917, opened 2023-10-16
feat(Buttons, Touchable, Menu): accessibility improvements

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1561)

### DIFF
--- a/src/__stories__/sheet-story.tsx
+++ b/src/__stories__/sheet-story.tsx
@@ -42,6 +42,8 @@ export const Default: StoryComponent = () => {
             <Stack space={16}>
                 <ButtonPrimary
                     disabled={open}
+                    aria-expanded={open}
+                    aria-haspopup="dialog"
                     onPress={() => {
                         setOpen(true);
                     }}
@@ -104,6 +106,8 @@ export const RadioList: StoryComponent<RadioListSheetArgs> = ({
             <Stack space={16}>
                 <ButtonPrimary
                     disabled={open}
+                    aria-expanded={open}
+                    aria-haspopup="dialog"
                     onPress={() => {
                         setOpen(true);
                         setSelected(null);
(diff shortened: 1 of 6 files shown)


pladaria (approved):

src/button.tsx
```diff
@@ -354,8 +356,11 @@ const Button = React.forwardRef<TouchableElement, ButtonProps & {type: ButtonTyp
         style: {cursor: props.fake ? 'pointer' : undefined, ...props.style},
         trackingEvent: props.trackingEvent ?? (props.trackEvent ? createDefaultTrackingEvent() : undefined),
         dataAttributes: props.dataAttributes,
+        'aria-label': props['aria-label'],
         'aria-controls': props['aria-controls'],
         'aria-expanded': props['aria-expanded'],
+        'aria-haspopup': props['aria-haspopup'],
+        'aria-busy': showSpinner ? true : undefined,
```
yceballost, line 363: the current implementation is already working with voice-over while loading state.. so... I wonder why do we need aria-busy 🤔

yceballost (approved):

src/button.tsx
atabel: If it has no effect, I'd remove it

src/__stories__/sheet-story.tsx
```diff
@@ -42,6 +42,8 @@ export const Default: StoryComponent = () => {
             <Stack space={16}>
                 <ButtonPrimary
                     disabled={open}
+                    aria-expanded={open}
```
atabel, line 45: you may want to change playroom sheet snippets too

src/button.tsx
aweell: Agree. Removed

marcoskolodny: Good catch, I tested a button with aria-busy to see if it was working, but I didn't notice it was the same behavior as the case when it doesn't have aria-busy.

Removed this prop :)

src/__stories__/sheet-story.tsx
marcoskolodny: done :)

atabel (approved):

aweell (approved):

PR #919, opened 2023-10-17
feat(OverscrollColor, Storybook): overscroll alternative variant and storybook default platform detection

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1593)

Along with the changes in the OverscrollColor component, now storybook will allow you to choose "auto" option in the platform Add-on (i.e. detect the current platform by using the browser's user agent).

### DIFF
--- a/src/overscroll-color-context.tsx
+++ b/src/overscroll-color-context.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import {useIsInverseVariant} from './theme-variant-context';
+import {useThemeVariant} from './theme-variant-context';
 import {useTheme, useScreenSize} from './hooks';
 import createNestableContext from './nestable-context';
 import {isInsideNovumNativeApp, getPlatform} from './utils/platform';
@@ -55,8 +55,14 @@ export const OverscrollColorProvider = ({children}: ProviderProps): JSX.Element 
 };
 
 const OverscrollColorComponent = () => {
-    const isInverseVariant = useIsInverseVariant();
-    useSetValue(isInverseVariant ? vars.colors.navigationBarBackground : vars.colors.background);
+    const variant = useThemeVariant();
+    useSetValue(
+        variant === 'inverse'
+            ? vars.colors.backgroundBrand
+            : variant === 'alternative'
+            ? vars.colors.backgroundAlternative
+            : vars.colors.background
+    );
 
     return null;
 };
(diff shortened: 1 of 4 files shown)


src/overscroll-color-context.tsx
```diff
@@ -55,8 +55,14 @@ export const OverscrollColorProvider = ({children}: ProviderProps): JSX.Element
 };
 
 const OverscrollColorComponent = () => {
-    const isInverseVariant = useIsInverseVariant();
-    useSetValue(isInverseVariant ? vars.colors.navigationBarBackground : vars.colors.background);
+    const variant = useThemeVariant();
```
marcoskolodny, line 58: @Winde these are the changes related to https://jira.tid.es/browse/WEB-1593 :)

src/overscroll-color-context.tsx
```diff
@@ -55,8 +55,14 @@ export const OverscrollColorProvider = ({children}: ProviderProps): JSX.Element
...
+    useSetValue(
+        variant === 'inverse'
```
Winde, line 60: I'm not a big fan about chaining ternary operators. what about:
```
const getColorFromVariant = (themeVariant: Variant): string => {
    switch (themeVariant) {
        case 'default':
            return vars.colors.navigationBarBackground;
        case 'alternative':
            return vars.colors.backgroundAlternative;
        case 'inverse':
            return vars.colors.navigationBarBackground;
        default:
            const exhaustiveCheck: never = themeVariant;
            throw new Error(`Invalid variant: ${exhaustiveCheck}`);
    }
};

const OverscrollColorComponent = () => {
  useSetValue(getColorFromVariant(variant))

  return null;
};
```

Winde: I also think there's could be a product color gap here for the "backgroundBrandSecondary" that has been used in MOVES as a background color for some sections.

In that case the Overscroll would show up as "backgroundBrand". I don't think it will look broken in the app by any means, as it will mimic the color of the navbar instead of the body, but maybe it's a limitation of the Overscroll to be thought about, that it works with variants but not with the real colors that may acompany those variants, it assumes the color.

Winde (approved):

src/overscroll-color-context.tsx
marcoskolodny: We've considered the possibility of allowing custom colors for the overscroll, but we concluded that we won't open that door until we have a clear scenario in which the current implementation is not enough.

pladaria (approved):

PR #924, opened 2023-10-25
feat(Snackbar): support persistent snackbars and dismiss button

atabel:
Jira: https://jira.tid.es/browse/WEB-1587
Figma spec: https://www.figma.com/file/yCHLIfy4WMfRdlADwyL4kZ/%F0%9F%94%B8-Snackbar-Specs?type=design&node-id=0-1&mode=design&t=0A9H1Zwcmrc3Pika-0
Bridge spec: https://github.com/Telefonica/webview-bridge/issues/117
Bridge PR: https://github.com/Telefonica/webview-bridge/pull/118


src/__stories__/snackbar-story.tsx
```diff
@@ -12,13 +12,20 @@ type Args = {
     message: string;
```
yceballost, line 12: just to keep my TOC calm, can you change the order between buttonText and message in storybook control table?
image

Order:
- type
- message
- buttonText
- withDismiss
- duration

Extra ball, can we remove "5000" & "10000" from storybook controls? This configuration is something we should deprecate in the next major

yceballost (changes requested):

src/snackbar.tsx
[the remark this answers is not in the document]
pladaria: Not a bad idea. We could implement a useVisibleTimer hook

src/snackbar.tsx
```diff
@@ -128,9 +164,31 @@ const Snackbar: React.FC<Props> = ({
...
+
+            nativeMessage({
+                message,
+                // @ts-expect-error duration can be 'PERSISTENT' in new webview-bridge lib versions, and old apps will ignore it
+                duration: duration === Infinity ? 'PERSISTENT' : undefined,
+                buttonText,
+                type,
+                ...{withDismiss},
+            }).then((result: unknown) => {
+                // there are terser ways to do this checks, but this one satisfies TS
```
pladaria, line 177: perhaps you could type the result like:
```
result: undefined | {action?: string}
```

pladaria (approved):

src/snackbar.tsx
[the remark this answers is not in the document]
atabel: that's a tricky one, because the browsers already tend to pause setTimeouts when the tab isn't in foreground.
Also, it's quite a corner case that I think it doesn't worth the added complexity. I mean, Snackbars use to appear as a result of a user action with a feedback purpose, and given they are in the screen only 5/10s it sounds weird to me that the user would switch to a different tab in that short period of time and expect the snackbar to still be there when they come back to the tab

src/snackbar.tsx
atabel: TS don't like this because current (old) version of webview-bridge returns void as result

src/snackbar.tsx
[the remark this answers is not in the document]
atabel: fixed

src/__stories__/snackbar-story.tsx
atabel: done

yceballost (approved):

aweell (approved):

marcoskolodny (approved):

PR #925, opened 2023-10-26
fix(icons): remove svgo optimizations that made some icons break in iOS

atabel:
CMS uses mistica-web to extract the icons and then serve them as svg files. For example https://fr-cert1-br.mytelco.io/2WJMF60PcwFO8ypmZ1qsBzCV03Pn2Iv0FaQojCQXOEoNeBgbIZzuejg

Then, webapp sends a bridge message (bottom sheet) to native app and sends these cms urls to native app.
The native app downloads de svg file and renders it natively. But for some reason iOS does not render some svgs properly.

After investigation, we found that iOS could render the original mistica-icons icon but not the one exposed by mistica-web and served by CMS. We found that the changes between both versions of the icon (mistica-icons/mistica-web) was the transformations made by `@svgr/plugin-svgo` when we make the import form mistica-icons to mistica-web.

After removing that svgo plugin from the import script, the generated icon is identical to the mistica-icons one. As mistica-icons already has optimized icons (svgo is used by figma at export time), it doesn't make sense to re-optimize the icons, so I've removed the `@svgr/plugin-svgo` from the script.

Discussion: https://teams.microsoft.com/l/message/19:c5b4714b5c12473c80e672a3b8cc4228@thread.skype/1696516608001?tenantId=9744600e-3e04-492e-baa1-25ec245c6f10&groupId=bef9be07-03e4-4afb-8a79-f720a63e4c28&parentMessageId=1696516608001&teamName=Novum&channelName=Vivo%20BR&createdTime=1696516608001&allowXTenantAccess=false


atabel:
> **Size stats**
> 
> master	this branch	diff
> Total JS	9.52 MB	9.56 MB	+42.9 kB
> JS without icons	1 MB	1 MB	0 B
> Lib overhead	127 kB	127 kB	+27 B
> Lib overhead (gzip)	32.6 kB	32.6 kB	+6 B

A diff of +42.9 kB, given we have 1656 icons, means around +26 Bytes per icon, not a big deal

pladaria (approved):

yceballost (approved):

marcoskolodny (approved):

PR #926, opened 2023-10-26
chore(Storybook): New UI for Storybook

yceballost:


.storybook/manager.js
```diff
@@ -6,4 +6,7 @@ import {createStorybookTheme} from './storybook-manager-theme';
 addons.setConfig({
     panelPosition: 'right',
     theme: createStorybookTheme('Movistar'),
+    toolbar: {
+        title: {hidden: true},
```
yceballost, line 10: This hides this element
Screenshot 2023-10-26 at 18 09 52

.storybook/manager-head.html
```diff
@@ -0,0 +1,103 @@
+<style type="text/css">
+    /* styles for the storybook sidebar */
+    .search-result-item {
+        border-bottom: 1px solid rgba(0, 0, 0, 0.1) !important;
```
atabel, line 4: are all those `!important` needed? There isn't any official way of customizing these storybook styles?

I'm afraid a future update of storybook could break this

.storybook/manager-head.html
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /* icon size */
+    .sidebar-svg-icon,
+    .sidebar-item svg,
+    .search-result-item svg {
+        width: 4px !important;
+        height: 4px !important;
+        color: inherit !important;
+        opacity: 0 !important; /* hide icon to keep aligment with the rest of list elements */
```
atabel, line 36: `visibility: hidden`?

.storybook/manager-head.html
yceballost: I followed the example from NASA storybook
https://github.com/nasa-jpl/explorer-1/blob/main/storybook/manager-head.html

https://storybook.js.org/docs/react/configure/theming#css-escape-hatches

.storybook/manager-head.html
yceballost: I solved the aligment by other way. Hiding icons with display none and adding some margins-right

.storybook/manager-head.html
yceballost: I removed `!important` tag from the most of cases. The cases with `!important` needs it to force override the styles :(

atabel (approved):

PR #927, opened 2023-10-27
feat(Counter): create component

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1554)
Specs: [Link](https://www.figma.com/file/LhZ5g2KAckJLzBIA1720H0/%F0%9F%94%B8-%5BREADY%5D-Counter-Component-Specs?type=design&node-id=0-1&mode=design&t=Mv7qzEqsEGZ4xrcD-0)


src/counter.tsx
yceballost, line 1: weird right spacing in numbers
https://tinyurl.com/yp3dp6ts

yceballost (changes requested):

src/counter.tsx
marcoskolodny: should be fixed now :)

src/__screenshot_tests__/__image_snapshots__/counter-screenshot-test-tsx-counter-min-3-max-4-default-value-3-removable-true-disabled-true-1-snap.png
yceballost, line 1: too much screenshots? Just asking. I dont really know if we make screenshots of all of the brands in all of the states in the rest of the component.

yceballost (changes requested):

src/__tests__/counter-test.tsx
```diff
@@ -0,0 +1,111 @@
...
+test('Counter with default value', async () => {
+    render(
+        <ThemeContextProvider theme={makeTheme()}>
+            <Counter min={3} max={5} defaultValue={4} />
+        </ThemeContextProvider>
+    );
+
+    expect(screen.getByText('4')).toBeInTheDocument();
+
+    const increaseButton = screen.getByRole('button', {name: 'Incrementar valor'});
```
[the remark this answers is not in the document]
marcoskolodny: No. Locale for unit tests is Spanish by default, and the counter label is set in the theme's language.

src/counter.css.ts
```diff
@@ -0,0 +1,144 @@
+import {style} from '@vanilla-extract/css';
+import * as mq from './media-queries.css';
+import {vars} from './skins/skin-contract.css';
+
+const transitionTiming = '0.2s ease-in-out';
+const BUTTON_SIZE = 32;
```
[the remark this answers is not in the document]
marcoskolodny: fixed

src/__screenshot_tests__/__image_snapshots__/counter-screenshot-test-tsx-counter-default-skin-vivo-new-is-inverse-true-is-dark-mode-true-1-snap.png
pladaria, line 1: @yceballost some dark skins don't have enough contrast in the +/- buttons

yceballost: yep, we know it. In this case is not a problem because Vivo don't use dark mode by the moment. Vivo skin is leading by Vivo design team

src/counter.css.ts
```diff
@@ -0,0 +1,145 @@
...
+    transition: 'transform 0.15s ease-in-out',
+});
+
+export const defaultButtonBackground = style([
+    baseButtonBackground,
+    {
+        background: vars.colors.brandLow,
+        selectors: {
+            [`${buttonContainer}:active:not(${isButtonDisabled}) &`]: {
+                transform: 'scale(1.0625)',
```
pladaria, line 91: is this scale noticeable?

src/__screenshot_tests__/__image_snapshots__/counter-screenshot-test-tsx-counter-default-skin-vivo-new-is-inverse-true-is-dark-mode-true-1-snap.png
marcoskolodny: This will be considered by design in a separate task, because this is not the only component with this issue.

src/counter.css.ts
marcoskolodny: At least I can notice it, but maybe @yceballost can double check this

yceballost: where does this number come from? 🤔

pladaria: Ok, forget my comment. I thought this scale was a different thing

pladaria (approved):

src/counter.css.ts
yceballost: I tested 1.1 but it is too much. We can use 1.06

yceballost (approved):

src/theme.tsx
```diff
@@ -44,6 +44,12 @@ const TEXTS_ES = {
     progressBarCompletedLabel: 'completo',
     progressBarStepLabel: 'Paso 1$s de 2$s',
     pinFieldInputLabel: 'Dígito 1$s de 2$s',
+    counterRemoveLabel: 'Borrar elemento',
+    counterIncreaseLabel: 'Incrementar valor',
+    counterDecreaseLabel: 'Decrementar valor',
```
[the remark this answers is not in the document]
aweell: Utilizaría aumentar/disminuir en castellano en vez de incrementar/decrementar, suena mucho mas natural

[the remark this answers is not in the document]
marcoskolodny: updated :)

aweell (approved):

src/__screenshot_tests__/counter-screenshot-test.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+        const counter = await screen.findByTestId('counter');
+
+        const image = await counter.screenshot();
+        expect(image).toMatchImageSnapshot();
+    }
+);
+
+test.each`
+    min  | max   | defaultValue | removable | disabled
```
atabel, line 39: does it make sense to test all these combinations?

src/counter.css.ts
```diff
@@ -0,0 +1,132 @@
...
+const transitionTiming = '0.2s ease-in-out';
+const BUTTON_SIZE = 32;
+const TOUCHABLE_AREA_WIDTH = 40;
+
+export const hasTrashIcon = style({});
+export const isButtonDisabled = style({});
+
+export const counter = style({
+    padding: 3,
+    boxSizing: 'border-box',
```
atabel, line 14: probably not needed

src/counter.css.ts
```diff
@@ -0,0 +1,132 @@
...
+    borderRadius: vars.borderRadii.button,
+    background: vars.colors.backgroundContainer,
+    display: 'inline-flex',
+    cursor: 'default',
```
atabel, line 18: probably not needed

src/counter.css.ts
```diff
@@ -0,0 +1,132 @@
...
+
+export const button = style({
+    position: 'absolute',
+    top: -(TOUCHABLE_AREA_WIDTH - BUTTON_SIZE) / 2,
+    left: -(TOUCHABLE_AREA_WIDTH - BUTTON_SIZE) / 2,
+    width: TOUCHABLE_AREA_WIDTH,
+    height: TOUCHABLE_AREA_WIDTH,
+    display: 'flex',
+    justifyContent: 'center',
+    alignItems: 'center',
```
atabel, line 46: try to use sprinkles when possible

src/counter.css.ts
```diff
@@ -0,0 +1,132 @@
...
+});
+
+export const defaultButtonIcon = style([
```
atabel, line 49: array not needed

src/counter.css.ts
```diff
@@ -0,0 +1,132 @@
...
+                opacity: 1,
+            },
+        },
+    },
+]);
+
+export const decreaseButtonIcon = style([
+    defaultButtonIcon,
+    {
+        pointerEvents: 'auto',
```
atabel, line 79: why needed?

src/counter.tsx
```diff
@@ -0,0 +1,214 @@
...
+    const getValueInRange = React.useCallback(
+        (value?: number) => {
+            return value === undefined ? min : Math.max(min, Math.min(max, value));
+        },
+        [min, max]
+    );
+
+    const [currentValue, setCurrentValue] = React.useState<number>(getValueInRange(defaultValue));
+
+    React.useEffect(() => setCurrentValue(getValueInRange(defaultValue)), [defaultValue, getValueInRange]);
```
atabel, line 44: you may not need this effect, just apply the `getValueInRange` when calling `updateValue`

src/counter.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+    return [isControlledByParent ? getValueInRange(value) : currentValue, updateValue];
+};
+
+type Props = {
+    value?: number;
+    defaultValue?: number;
+    min?: number;
+    max?: number;
+    onChange?: (value: number) => void;
```
atabel, line 63: Not sure, but I think we should call this `onChangeValue` to follow the same naming conventions we use in other components (`onChange` props use to receive an event, not a value)

src/counter.tsx
```diff
@@ -0,0 +1,214 @@
...
+    onRemove?: () => void;
+    children?: void;
```
atabel, line 65: I'd say you don't need to explicitly set children to void. It used to be needed in old react TS types, newer versions of `React.FC` don't have an implicit children by default.

src/counter.tsx
```diff
@@ -0,0 +1,214 @@
...
+    disabled?: boolean;
+    dataAttributes?: DataAttributes;
+    removeLabel?: string;
+    increaseLabel?: string;
+    decreaseLabel?: string;
```
atabel, line 70: I'd prefer `removeButtonLabel`, `increaseButtonLabel`, `decreaseButtonLabel`

src/counter.tsx
```diff
@@ -0,0 +1,214 @@
...
+    disabled,
+    removeLabel,
+    increaseLabel,
+    decreaseLabel,
+    valueLabel,
+}) => {
+    const variant = useThemeVariant();
+    const {texts} = useTheme();
+
+    const minValue = min === undefined ? 0 : min;
```
atabel, line 91: isn't this equivalent to doing this on function props destructuring?
```ts
min =  0
```

src/counter.tsx
```diff
@@ -0,0 +1,214 @@
...
+        defaultValue,
+        min: minValue,
+        max: maxValue,
+        onChange,
+    });
+
+    const hasTrashIcon = !!onRemove && currentValue === minValue;
+
+    const getRemoveLabel = () => {
+        return removeLabel === undefined ? texts.counterRemoveLabel : removeLabel;
```
atabel, line 105:
```suggestion
        return removeLabel  ?? texts.counterRemoveLabel;
```
(same for other labels)

src/counter.tsx
```diff
@@ -0,0 +1,214 @@
...
+                    variant === 'inverse' ? vars.colors.backgroundContainer : vars.colors.border
+                }`,
+            }}
+        >
+            <Inline space={8} alignItems="center">
+                <div
+                    className={classNames(styles.buttonContainer, {
+                        [styles.hasTrashIcon]: hasTrashIcon,
+                        [styles.isButtonDisabled]: disabled || (!hasTrashIcon && currentValue === minValue),
+                        [styles.disabled]: !disabled && !hasTrashIcon && currentValue === minValue,
```
atabel, line 137: this logic is weird to me, are you applying a disabled class when `disabled` prop is false?

src/counter.tsx
marcoskolodny: Not really, because the value label depends on whether the min/max were actually provided by the user or not. If it was undefined, then the label won't include information about it

src/counter.tsx
marcoskolodny: yes, because there are two possible disabled states. One when the value is equal to min/max (only disable button), and another one when the entire counter is disabled (everything has less opacity, including the value and container). In the second case, if the value equals min/max, we don't want to apply the opacity filter twice to buttons

src/counter.css.ts
marcoskolodny: The icons for decrease and remove are both inside the same touchable, but they are positioned so that one of them is hidden. If we don't use the pointerEvents auto/none, hovering on the hidden one would trigger the hover effect in the touchable.

src/counter.tsx
marcoskolodny: fixed

src/counter.tsx
marcoskolodny: fixed

src/counter.tsx
marcoskolodny: fixed

src/counter.tsx
marcoskolodny: fixed

src/counter.css.ts
marcoskolodny: removed

src/counter.css.ts
marcoskolodny: removed

src/__screenshot_tests__/counter-screenshot-test.tsx
marcoskolodny: Yes, these are different scenarios for when some of the controls are disabled because of value being equal to min/max, or when the buttons are disabled but the counter itself had a disabled state.

I've removed many useless tests that were combining different skins with inverse/darkMode

src/counter.tsx
marcoskolodny: Discussed offline. We concluded that default value should be set at the beginning, but changes on it should be ignored. There are other components in Mistica that are not following this approach. We will review them in a separate task ([WEB-1620](https://jira.tid.es/browse/WEB-1620))

atabel (approved):

PR #932, opened 2023-11-03
feat(FeedbackLoader): CHECKOUT-6410 Add FeedbackLoader to Mistica Community

Marcosld:


yceballost:
This component need an evolution to be included in Mística. Vivo is evolving the specs to remove some problems with animation definition. https://jira.tid.es/browse/OBVIVO-2050

atabel:
Hi @Marcosld, Design Core (cc @yceballost) have the idea of making this component more global with the intention of adding it to mistica core (instead of community). They are working on global specs for the component. Once the specs are ready we can continue with this, and this PR could be a good starting point. Meanwhile we can keep this PR as draft.

Are you fine with this approach?

PR #936, opened 2023-11-08
feat(CardActionIconButton): element to use in card actions

atabel:
This change is motivated by this: https://jira.tid.es/browse/ACCOUNT-23470
And how CPS team is using card actions in Novum https://github.com/Telefonica/webapp/blob/master/web/src/account/components/modules/data-card-advanced/index.tsx#L90

This allows using card actions like this:
```ts
actions={[
    <CardActionIconButton onPress={doSomething} label="Do something" Icon={SomeIcon}/>
]}
```

We still support the classic way:

```ts
actions={[
    {onPress: doSomething, label: "Do something", Icon: SomeIcon}
]}
```
The advantage of being able to use a `CardActionIconButton`  `ReactElement` instead of a plain object is that someone could implement complex logic in their card actions if needed, like using hooks:

```ts
const MySpecialCardAction = () => {
    const onPress = useHandlerForSpecialAction();
    
    return <CardActionIconButton onPress={onPress} label="Do something" Icon={SomeIcon}/>
}

<DataCard actions={[<MySpecialCardAction/>]} />
```

The disadvantage of this is that we are allowing any `React.ReactElement` in the `actions` array because TS doesn't provide a way to limit the element type, so potentially someone could pass some incorrect element there. We'd need to trust the developer will use components that end rendering a `CardActionIconButton` element


penchochris (approved):

src/card.tsx
```diff
@@ -89,34 +89,52 @@ export const CardActionsGroup = ({
...
+                    position: 'absolute',
+                    right: padding,
+                    top: padding,
+                    zIndex: 3, // needed because images has zIndex 1 and touchable overlay has zIndex 2
+                }}
+            >
+                <div className={sprinkles({display: 'flex'})}>
+                    {finalActions.map((action, index) => {
+                        if ('label' in action) {
+                            // action it's a CardAction object
```
pladaria, line 129:
```suggestion
                            // action is a CardAction object
```

src/card.tsx
```diff
@@ -89,34 +89,52 @@ export const CardActionsGroup = ({
...
+                            return <CardActionIconButton key={index} {...action} />;
+                        }
+                        // action it's a React.ReactElement
```
pladaria, line 132:
```suggestion
                        // action is a React.ReactElement
```

pladaria (approved):
I'd add some stories + tests covering this new usage and the "default/inverse/media" context types

atabel:
> I'd add some stories + tests covering this new usage and the "default/inverse/media" context types

I'm not sure if a story is needed, because there isn't any visual change, just a new api to achieve the same result. About the default/inverse/media context, it's everything internal.

I agree it would be nice to have some example documenting the new api, but I'm not sure where is the best place to add those examples, I don't want to use the current card stories for this purpose, I could create a new story but where should I place it inside storybook?

marcoskolodny (approved):

PR #937, opened 2023-11-08
fix(Counter): hover in control buttons not working

marcoskolodny:


src/counter.css.ts
```diff
@@ -32,7 +32,7 @@ export const valueContainer = sprinkles({
     justifyContent: 'center',
 });
 
-export const buttonContainer = sprinkles({
+export const buttonContainer = style({
```
atabel, line 35: can you explain why this was causing the bug?

yceballost (approved):

src/counter.css.ts
atabel: talked offline

atabel (approved):

PR #941, opened 2023-11-14
feat(createNestableContext): allow custom valuesReducer

atabel:
jira: https://jira.tid.es/browse/WEB-1632

Webapp uses nestable-context for navbar state.
There is an special attribute in navbar state called `resetToDefaultState` that is used to reset the state of the navbar. 
https://github.com/Telefonica/webview-bridge/blob/master/src/utils.ts#L106-L110

When that attribute is used, the merging logic should be different. Something like this:
image

### DIFF
--- a/src/__tests__/nestable-context-test.tsx
+++ b/src/__tests__/nestable-context-test.tsx
@@ -132,3 +132,74 @@ test('works as expected with transitions', async () => {
 
     expect(await screen.findByText('The value is: b')).toBeInTheDocument();
 });
+
+test('when value is an object, nested values are merged', () => {
+    type Value = {foo?: string; bar?: string};
+
+    const {Provider, Getter, Setter} = createContext<Value>({});
+
+    const C = () => <Setter value={{bar: 'c'}} />;
+
+    const B = () => (
+        <>
+            <Setter value={{bar: 'b'}} />
+            <C />
+        </>
+    );
+
+    const A = () => (
+        <>
+            <Setter value={{foo: 'a'}} />
+            <B />
+        </>
+    );
+
+    render(
+        <Provider>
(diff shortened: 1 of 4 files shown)


src/__tests__/nestable-context-test.tsx
```diff
@@ -132,3 +132,74 @@ test('works as expected with transitions', async () => {
 
     expect(await screen.findByText('The value is: b')).toBeInTheDocument();
 });
+
+test('when value is an object, nested values are merged', () => {
```
atabel, line 136: this isn't new behavior, I've just added a test for that

src/__tests__/nestable-context-test.tsx
```diff
@@ -132,3 +132,74 @@ test('works as expected with transitions', async () => {
...
+        <Provider>
+            <Getter>{({foo, bar}) => `foo is ${foo} and bar is ${bar}`}</Getter>
+            <A />
+        </Provider>
+    );
+
+    expect(screen.getByText('foo is a and bar is c')).toBeInTheDocument();
+});
+
+test('can use a custom valuesReducer', () => {
```
atabel, line 167: this covers the new feature

src/nestable-context.tsx
pladaria, line 1: Unrelated to this task but I wouldn't export the `createNestableContext` as `default`. Better a named export.

And, in the text, don't change the function name (use `createNestableContext` instead of `createContext`)

pladaria (approved):

src/nestable-context.tsx
atabel: agree, changed

marcoskolodny (approved):

PR #943, opened 2023-11-14
fix(FixedFooterLayout): consider bottom inset area for calculations

brtbrt:


src/test-utils/index.tsx
```diff
@@ -57,6 +60,22 @@ const DEVICES: DeviceCollection = {
             isLandscape: false,
         },
     },
+    [MOBILE_DEVICE_IOS_INSET]: {
```
brtbrt, line 63: I thought I'd find it at handy having a default device to choose from to simulate the inset at handy.
This is an initial proposal and it surely depends on your point of view/what do you want to accomplish with this etc. Since I surely don't know all the scenarios we could fall into, share your best generic "inset" device case we should consider for testing purposes here, and I'll update it!

For instance: what about the upper inset? I wouldn't simulate it just because the pages within the app consider the top safe area already, in some way. But much more scenarios could need to encompass that.

src/fixed-footer-layout.tsx
```diff
@@ -54,17 +55,16 @@ const FixedFooterLayout: React.FC<Props> = ({
     const containerRef = React.useRef<HTMLDivElement>(null);
     const {isTabletOrSmaller} = useScreenSize();
     const {platformOverrides} = useTheme();
-    const {height: realFooterHeight, ref} = useElementDimensions();
+    const {height: domFooterHeight, ref} = useElementDimensions();
```
brtbrt, line 58: semantically better to my eyes. everything's "real" :D

src/test-utils/index.tsx
```diff
@@ -57,6 +60,22 @@ const DEVICES: DeviceCollection = {
...
+        platform: 'ios',
+        viewport: {
+            width: 375,
+            height: 667,
+            deviceScaleFactor: 2,
+            isMobile: true,
+            hasTouch: true,
+            isLandscape: false,
+            safeAreaInset: {
+                bottom: '34px',
```
brtbrt, line 75: this is the inset I got on an iphone 13 mini, I think on a max could be more higher?

atabel, line 75: please add a comment here explaining this is the size of iPhone bottom handle (I don't know how is this thing called)

src/__screenshot_tests__/__image_snapshots__/fixed-footer-screenshot-test-tsx-fixed-footer-with-safe-inset-at-the-bottom-2-snap.png
marcoskolodny, line 1: I don't see the shadow in here. Is that expected?

src/__screenshot_tests__/fixed-footer-screenshot-test.tsx
```diff
@@ -37,3 +37,35 @@ test('Fixed footer with static position when height is smaller than the footer i
...
+test('Fixed footer with safe inset at the bottom', async () => {
+    const page = await openStoryPage({
+        id: 'layout-fixed-footer-layout--more-complex-footer',
+        device: 'MOBILE_IOS_INSET',
+    });
+
+    const image = await page.screenshot();
+    expect(image).toMatchImageSnapshot();
+
+    await (await screen.findByRole('button', {name: 'Load more text'})).click();
```
marcoskolodny, line 50: maybe you could use `.click({clickCount: 3})` instead of repeating the same line 3 times

marcoskolodny (approved):

src/utils/css.tsx
```diff
@@ -2,3 +2,8 @@ const BASE_FONT_SIZE = 16; // browser's default font-size in pixels
 
 export const pxToRem = (px: number | string): string =>
     `${(parseFloat(px as string) / BASE_FONT_SIZE).toFixed(3)}rem`;
+
+export const safeAreaInsetLeft = `max(env(safe-area-inset-left), var(--acceptance-test-override-safe-area-inset-left, 0px))`;
+export const safeAreaInsetTop = `max(env(safe-area-inset-top), var(--acceptance-test-override-safe-area-inset--top, 0px))`;
+export const safeAreaInsetBottom = `max(env(safe-area-inset-bottom), var(--acceptance-test-override-safe-area-inset--bottom, 0px))`;
+export const safeAreaInsetRight = `max(env(safe-area-inset-right), var(--acceptance-test-override-safe-area-inset--right, 0px))`;
```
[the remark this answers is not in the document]
brtbrt: agreed. coming up!

atabel (approved):

PR #946, opened 2023-11-16
fix(FixedFooterLayout): footer alignment in large desktop viewport

atabel:
This is the diff after this fix (I took the screenshot before the fix to show the difference):
fixed-footer-screenshot-test-tsx-button-fixed-footer-layout-on-large-desktop-1-diff


src/responsive-layout.css.ts
```diff
@@ -71,9 +71,6 @@ export const responsiveLayout = style({
     paddingRight: 'env(safe-area-inset-right)',
 
     margin: `0 ${sideMargin}`,
-});
-
-export const expandedResponsiveLayout = style({
     '@media': {
         [mq.largeDesktop]: {
             margin: `0 calc((100vw - ${LARGE_DESKTOP_MAX_WIDTH}px) / 2)`,
```
atabel, line 76: the problem was this mq was only being applied to `ResponsiveLayout`, and not to `InternalResponsiveLayout` (used by `ButtonFixedFooterLayout`)

pladaria (approved):

marcoskolodny (approved):

PR #949, opened 2023-11-17
fix(useDisableBodyScroll): fix nested usage

pladaria:
The main issue was that the hook state wasn't shared between different hook instances. Extracting the state to module variables fixed the issue.

The issue was reproduced on creating multiple Overlays with scroll disabled: one for the menu, another for the modal.

Added some unit tests to cover those scenarios

### DIFF
--- a/src/hooks.tsx
+++ b/src/hooks.tsx
@@ -18,12 +18,13 @@ export const useTheme = (): Theme => {
     return theme;
 };
 
+let bodyStyles = '';
+let bodyScrollTop = 0;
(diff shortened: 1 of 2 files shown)


src/hooks.tsx
pladaria, line 1: Instead of this, I considered to store the hook state as data-attributes in the body, but discarded that option because it would add additional complexity with little benefit

Winde (approved):

atabel (approved):

marcoskolodny (approved):

PR #951, opened 2023-11-21
feat(Slider): rewrite component from scratch

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1583)
Specs: [Link](https://www.figma.com/file/UZHm7V3gO7nVnbf1b0Vj7b/%F0%9F%94%B8-Slider-Component-Specs?type=design&node-id=0-1&mode=design&t=hHBtlrMxYTuZQGqQ-0)

**Remarks about firefox**
- There is a known firefox bug caused by using setPointerCapture(). If you press the slider, drag the pointer on top of a button and then release it, the button will be clicked. The issue doesn't happen in Chrome or Safari, and it can be reproduced by using basic HTML (https://codepen.io/Marcos-Kolodny/pen/oNmdMxM). This was reported to firefox a long time ago and many users mention different scenarios where it happens, but it seems they are not working on it (https://bugzilla.mozilla.org/show_bug.cgi?id=1648893).
- A similar issue happens when you drag the pointer outside the element and then release it: the hover state of the element stays activated. This can also be observed in the above codepen example.


src/slider.tsx
```diff
@@ -1,164 +1,428 @@
...
-            steps.sort((a, b) => {
-                return a - b;
-            });
         }
-    }, [steps]);
 
-    const handleSlider = (value: number) => {
-        setValueRanger(value);
-    };
+        step = step <= 0 ? 1 : step;
```
pladaria, line 124: consider this:

```js
step = step | 0; // to avoid float or non numeric steps
step = step <= 0 ? 1 : step; // your current check
step = step > max ? max : step; // also avoid too large steps, which would make the final value unreachable
```

src/slider.tsx
```diff
@@ -1,164 +1,428 @@
...
+            document.addEventListener('keydown', handleKeyDown, false);
+            document.addEventListener('keyup', handleKeyUp, false);
+            return () => {
+                document.removeEventListener('keydown', handleKeyDown, false);
+                document.removeEventListener('keyup', handleKeyUp, false);
+            };
+        }, [isFocused]);
+
+        const handleKeyDown = (e: React.KeyboardEvent<HTMLDivElement>) => {
+            switch (e.keyCode) {
```
pladaria, line 208: `keyCode` is deprecated, consider using `key`

pladaria: we should remove the key-code constants file (in a different task)

src/slider.tsx
```diff
@@ -1,164 +1,428 @@
...
+        const sliderRef = React.useRef<HTMLDivElement>(null);
+
+        const [isPointerDown, setIsPointerDown] = React.useState(false);
+        const [isThumbHovered, setIsThumbHovered] = React.useState(false);
+        const [isFocused, setIsFocused] = React.useState(false);
+        const {isIos} = useTheme();
+
+        const isPointerOverElement = (element: HTMLElement | null, x: number, y: number) => {
+            const box = element?.getBoundingClientRect();
+            return box !== undefined && box.left <= x && x <= box.right && box.top <= y && y <= box.bottom;
```
pladaria, line 175:
```suggestion
            return !!box && box.left <= x && x <= box.right && box.top <= y && y <= box.bottom;
```

src/slider.tsx
marcoskolodny: we are using keyCode in several components. I believe this is something that should be updated in all the components at once in a separate PR in order to maintain a clean code.

I've created an issue for this: https://jira.tid.es/browse/WEB-1653

src/slider.tsx
```diff
@@ -1,164 +1,428 @@
...
+                default:
+                // do nothing
             }
+        };
 
-            setValueRanger(ranger);
-            setStep(steps);
-        }
-    }, [steps, max, min, getApproximation, value, step, getOrder]);
+        const isTouchableDevice = isClientSide() ? window.matchMedia('(pointer: coarse)').matches : false;
```
pladaria, line 233: perhaps we could move this `isTouchableDevice` to `utils/environment`. The same is implemented `Tooltip`

src/slider.tsx
yceballost, line 1: Interaction with voice over is not working properly

With voiceover:
1. You can choose the entire slider, that cause too much elements to interacts. I think you should only be able to select the knob 🤔
2. When you focus knob, voiceover say "group" and not the value

https://github.com/Telefonica/mistica-web/assets/6722153/f80f6960-dc22-437e-8560-e679181c02b4

yceballost (changes requested):

src/slider.tsx
```diff
@@ -1,164 +1,428 @@
...
-            <div className={styles.targetContainer}>
+        const onPointerMove = (e: PointerEvent) => {
+            cancelEvent(e);
+            updateCurrentValue(e.clientX);
+        };
+
+        const capturePointerMove = (e: React.PointerEvent<HTMLDivElement>) => {
+            const thumb = thumbRef.current;
+            if (thumb) {
+                thumb.onpointermove = onPointerMove;
```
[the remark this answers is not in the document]
marcoskolodny: This was intended, because if I write onPointerMove as a stable function, it will require handleChange() to be stable too, and we can't do that because handleChange depends on the function exposed by the form context (and that function is not memorized). If I do that, there will be an infinite loop of re-renders

src/slider.tsx
marcoskolodny: Discussed offline.

I will add the second condition, but the third one is unnecessary. As of now, it behaves exactly like the native html input, and I don't think we should change that. Also, even if we used it, there are still cases when you can't reach the maximum value (min=0, max=10, step=3).

src/slider.tsx
marcoskolodny: updated

src/slider.tsx
marcoskolodny: done

src/slider.tsx
marcoskolodny: should be fixed now

src/slider.tsx
```diff
@@ -1,164 +1,398 @@
...
+        };
+
+        const isTabKeyDownRef = React.useRef(false);
+
+        React.useEffect(() => {
+            const handleKeyDown = (e: KeyboardEvent) => {
+                if (isFocused) {
+                    switch (e.keyCode) {
+                        case ESC:
+                            inputRef.current?.blur();
```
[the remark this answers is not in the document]
marcoskolodny: I don't know why I thought that pressing ESC removed the focus of input fields. Removed this logic.

src/tooltip.tsx
```diff
@@ -544,7 +558,10 @@ const Tooltip: React.FC<Props> = ({
                                                 {extra || children}
                                             </ThemeVariant>
                                         </div>
-                                        <div className={styles.arrowContainer} style={arrowComputedStyles}>
+                                        <div
+                                            className={styles.arrowContainer}
+                                            style={{...arrowComputedStyles}}
```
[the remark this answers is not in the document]
marcoskolodny: fixed

yceballost (approved):

src/__stories__/slider-story.tsx
```diff
@@ -1,33 +1,72 @@
...
 
-export const Default: StoryComponent<Args> = ({disabled, steps, max, min, tooltip}) => {
+export const Controlled: StoryComponent<Args> = ({disabled, step, max, min, tooltip}) => {
+    const [value, setValue] = React.useState(min);
     return (
-        <Box paddingY={48} dataAttributes={{testid: 'slider'}}>
-            <Stack space={24}>
-                <Slider disabled={disabled} steps={steps} max={max} min={min} tooltip={tooltip} />
-            </Stack>
+        <Box paddingY={48} paddingX={20}>
```
yceballost, line 19: why not responsiveLayout?

pladaria (approved):
Excellent!

src/__tests__/slider-test.tsx
```diff
@@ -129,3 +129,88 @@ test('slider is accessible', () => {
...
+    expect(handleSubmitSpy).toHaveBeenCalledWith(7);
+
+    await clearAndType(maxField, '6');
+    expect(slider).toHaveValue('5');
+    expect(handleSubmitSpy).toHaveBeenCalledWith(5);
+
+    await clearAndType(stepField, '1');
+    expect(slider).toHaveValue('5');
+    // onChangeValue should not have been called here
+    expect(handleSubmitSpy).toHaveBeenCalledTimes(4);
```
atabel, line 215: I miss a test case:
- slider with min=0, max=10, step=1 and value in 3
- step changes to 2
- value should round and `handleSubmitSpy` should be called

pladaria (approved):

src/__tests__/slider-test.tsx
atabel: nevermind, already tested in line 204

PR #954, opened 2023-11-22
feat(Next.js): improved RSC support

atabel:
Main changes:
- Added "use client" directive to all mistica modules using client features (mainly hooks). This helps anyone using mistica in Next to just import the mistica client components from their app RSC without the need of an intermediary module with "use client" as boundary.
  -  Added a custom eslint rule to force adding "use client" directive when needed.
- Fixed an issue with `@vanilla-extract/dynamic` function `assignInlineVars` which returns an object with functions (it has a `.toString()` method`) instead of a plain object. This makes react complain when trying to pass these objects from RSC to client components as props. Props passing from server to client must be serializable.
  - created a custom `applyCssVars` function as a replacement.
  
Following work once this is merged:

- Migrate the repo example to latest next version and use RSC
- There are components that have the "use client" directive in this PR but could be migrated to be RSC with some refactoring.
- We could consider migrating the size stats script to use Next instead of webpack


packages/eslint-plugin-mistica-local-rules/use-client.js
atabel, line 1: custom eslint rule to automatically add "use client" directive when needed

src/utils/css.tsx
```diff
@@ -5,6 +5,24 @@ const BASE_FONT_SIZE = 16; // browser's default font-size in pixels
...
+    const matches = variable.match(/^var\((.*)\)$/);
+
+    if (matches) {
+        return matches[1];
+    }
+
+    return variable;
+};
+
+export const applyCssVars = (vars: Record<string, string>): Record<string, string> => {
```
atabel, line 18: This function replaces `assignInlineVars`

vite.config.js
```diff
@@ -12,6 +13,11 @@ export default defineConfig({
             // Change .css.js files to something else so that they don't get re-processed by consumer apps using vanilla extract too
             fileNames: ({name}) => `${name.replace(/\.css$/, '.css-mistica')}.js`,
         }),
+        {
+            ...preserveDirectivesPlugin(),
```
atabel, line 17: this plugin is needed to prevent vite removing the `"use client"` directives

README.md
```diff
@@ -87,6 +87,19 @@ that case we recommend to extract it to an external `const` variable outside of
 object reference will be the same in every re-render. If for some reason the `theme` must be dynamic in your
 app, consider to memoize it (for example, with `React.useMemo` hook).
 
+### Mistica in Next.js app router
```
atabel, line 90: this is important for tree shaking to work in next

README.md
```diff
@@ -87,6 +87,19 @@ that case we recommend to extract it to an external `const` variable outside of
...
+
+If you are using Next.js app router, we highly recommend enabling the follwing experimental config:
```
marcoskolodny, line 92:
```suggestion
If you are using Next.js app router, we highly recommend enabling the following experimental config:
```

packages/eslint-plugin-mistica-local-rules/use-client.js
```diff
@@ -0,0 +1,113 @@
...
+const HOOK_REGEX = /^use[A-Z]/;
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+    defaultOptions: [],
+    meta: {
+        type: 'suggestion',
+        messages: {
+            addUseClientHooks:
+                '{{hook}} only works in Client Components. Add the "use client" directive at the top of the file to use it.',
```
marcoskolodny, line 18:
```suggestion
                '{{hook}} only works in client components. Add the "use client" directive at the top of the file to use it.',
```

marcoskolodny (approved):

pladaria (approved):

PR #956, opened 2023-11-28
fix(Blocks): Set array props as read only

pladaria:
The motivation of this is to be able to pass API received values (typed as read only) to the component without a cast, spread or any other conversion to mutable array.

[WEB-1654](https://jira.tid.es/browse/WEB-1654)

### DIFF
--- a/src/community/blocks.tsx
+++ b/src/community/blocks.tsx
@@ -16,7 +16,7 @@ import type {ExclusifyUnion} from '../utils/utility-types';
 
 interface BlockContentProps {
     title?: string;
-    description?: Array<string> | string;
+    description?: ReadonlyArray<string> | string;
 }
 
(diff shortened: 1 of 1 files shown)


src/community/blocks.tsx
pladaria, line 1: perhaps we should do this with any prop that accepts an array. 

* ReadonlyArray props accept mutable and readonly arrays.
* Mutable Array props only accept mutable arrays.

So the most compatible API is the one that accepts ReadOnly Arrays

SusanaGonzalezPereira11P (approved):
Thanks!

src/community/blocks.tsx
marcoskolodny: I support this idea. Maybe we should create an issue for this

marcoskolodny (approved):

atabel (approved):

PR #961, opened 2023-12-01
feat(SSR): avoid CLS when hydrating some SSRed components

atabel:
jira: WEB-1657

Some mistica components render a different markup in mobile/desktop. When the component is server-side-rendered, it's rendered using the desktop markup, but when hydrated in front-end in a mobile viewport, the markup changes. This doesn't break React hydrate because the `useScreenSize` hook does a first render as desktop, but a second render happens immediately causing the [layout shift](https://web.dev/articles/cls?hl=es-419) 

There are some mitigations to avoid this issue:

- use css media queries instead of rendering different markup in mobile/desktop when possible. Avoid `useScreenSize` hook when possible.
- Automate testing this with SSR tests. I've found this solution
  - SSR
  - Open the page and take a screenshot of the SSR result before hydrate
  - Call hydrate from puppeteer
  - Take a second screenshot after hydrate
  - Compare both screenshots, should be the same:  0 [CLS](https://web.dev/articles/cls?hl=es-419)

After enabling this automate testing solution, I've identified issues in some components:

### `Image` (fixed in this PR)
The skeleton is not removed until client code is downloaded/parsed/executed and hydration is completed. This made SSRed Images loading suboptimal. With the changes in this PR, we can remove the skeleton when the image finishes loading without waiting for hydration.
image-ssr-acceptance-test-tsx-ssr-image-diff

### `MainNavigationBar` (fixed in this PR)
There is a big CLS issue in mobile, because the markup uses de desktop layout. After hydration, in mobile, a whole different markup is used (video attached to jira ticket)
navigation-bar-ssr-acceptance-test-tsx-ssr-main-navigation-bar-diff

### `FeedbackScreen` (fix pending)
Here we have the same problem described in `MainNavigationBar`
feedback-screen-ssr-acceptance-test-tsx-ssr-info-feedback-screen-diff

### `Popover` / `LoadingBar`
These are components that are only rendered in client inside a Portal, so there isn't a possible fix, and the rendering difference is expected:
loading-bar-ssr-acceptance-test-tsx-ssr-loading-bar-diff
popover-ssr-acceptance-test-tsx-ssr-popover-diff


src/image.tsx
atabel, line 1: better review with "Hide whitespace" option

src/image.tsx
```diff
@@ -178,55 +178,72 @@ export const ImageContent = React.forwardRef<HTMLImageElement, ImageProps>(
...
+                />
+                {/* When using SSR, we render a small script that makes the img visible as soon as it finishes loading, without waiting for React client hydrate */}
+                {withLoadingFallback && (
+                    <script
+                        // eslint-disable-next-line react/no-danger
+                        dangerouslySetInnerHTML={{
+                            __html: `document.getElementById("${imageId}").addEventListener('load', (e) => e.target.style.opacity = "1")`,
+                        }}
+                    />
+                )}
```
atabel, line 243: this is the trick to be able to remove the skeleton before hydration

src/inline.tsx
```diff
@@ -5,30 +5,37 @@ import {getPrefixedDataAttributes} from './utils/dom';
...
-        | 24
-        | 32
-        | 40
-        | 48
-        | 56
-        | 64
-        | 'between'
-        | 'around'
-        | 'evenly';
+    space: FlexSpace | ByBreakpoint<NumericSpace>;
```
atabel, line 38: support setting `space` by breakpoint as we already support in `Stack` and `Box` components

src/test-utils/index.tsx
```diff
@@ -179,24 +183,64 @@ export const openStoryPage = ({
...
+                '__diff_output__'
+            ),
+        });
+    } catch (error: any) {
+        Error.captureStackTrace(error, checkHydrationMissmatch);
+        throw error;
+    } finally {
+        fs.unlinkSync(baselineImagePath);
+    }
+};
```
atabel, line 222: this is how we check there isn't any visual changes between SSRed content and hydrated ui

src/test-utils/index.tsx
```diff
@@ -179,24 +183,64 @@ export const openStoryPage = ({
...
 /**
  * Renders a page with a React component in the server and opens it in the browser, where it's hydrated client side.
  * `name` is the name (without extension) of a file in the __ssr_pages__ folder. This file exports the component to be rendered.
  */
-export const openSSRPage = ({
+export const openSSRPage = async ({
     name,
     device = TABLET_DEVICE,
     skin = MOVISTAR_SKIN,
+    checkHidrationVisualMissmatch = true,
```
atabel, line 232: I've enabled this new check in all the SSR tests by default. But I have disabled it in some of them

src/test-utils/ssr.tsx
```diff
@@ -31,17 +31,19 @@ const createWebpackEntries = (): {[entryName: string]: string} => {
...
-                <ThemeContextProvider
-                    theme={{
-                        skin: getSkinByName(skin || 'Movistar'),
-                        i18n: {locale: 'es-ES', phoneNumberFormattingRegionCode: 'ES'}
-                    }}
-                >
-                    <Component />
-                </ThemeContextProvider>
-            );`
+            window.hydrate = () => {
```
atabel, line 34: this function is called by puppeteer to have full control

src/__acceptance_tests__/carousel-ssr-acceptance-test.tsx
atabel, line 1: I'll create a ticket to fix carousel

src/__acceptance_tests__/feedback-screen-ssr-acceptance-test.tsx
atabel, line 1: the same for feedback screens

atabel (commented):
bett

src/image.tsx
```diff
@@ -178,55 +178,74 @@ export const ImageContent = React.forwardRef<HTMLImageElement, ImageProps>(
...
         React.useEffect(() => {
             // Needed because there is some race condition with SSR and onLoad events
+            // load event could be fired before the component is hydrated and mounted client side
             // https://github.com/facebook/react/issues/15446
             if (imageRef.current?.complete) {
                 onLoadHandler();
             }
         }, [onLoadHandler]);
 
+        const isLoading =
```
marcoskolodny, line 200: Why isLoading should always be true in server side? If I understood correctly, that case is handled by the trick of listening to the load event in line 239.

marcoskolodny (approved):

src/image.tsx
pladaria: clever

src/test-utils/index.tsx
```diff
@@ -179,24 +183,64 @@ export const openStoryPage = ({
...
     });
 
+const hydrateSSRPage = async (page: PageApi): Promise<void> => {
+    await page.evaluate(() => {
+        // @ts-expect-error this executes in the browser, see ssr.tsx
+        window.hydrate();
+    });
+};
+
+const checkHydrationMissmatch = async (page: PageApi): Promise<void> => {
```
pladaria, line 193: mismatch

src/image.tsx
atabel: the initial state of the image is with the skeleton (loading), and the server should render that.
The purpose of the `<script>` trick is to transition from the "loading" (skeleton) state, to the "loaded" (image shown) state removing the opacity 0. But we can't do that until the image has finished loading.

pladaria (approved):

PR #963, opened 2023-12-01
fix(Tooltip): make component reactive to changes in tooltip content

marcoskolodny:
Some of the tooltip screenshot tests were failing randomly. The cause was the tooltip's width changing for some reason that we couldn't explain. Given that the component was not reactive to these changes, the tooltip ended up misaligned after the width changed.

In this PR, we force the tooltip to recompute its position when it's width/height changes, and therefore fixing the issue.

### DIFF
--- a/src/tooltip.tsx
+++ b/src/tooltip.tsx
@@ -193,6 +193,7 @@ const Tooltip: React.FC<Props> = ({
     const isInverse = useIsInverseVariant();
 
     const targetRect = useBoundingRect(targetRef, isTooltipOpen);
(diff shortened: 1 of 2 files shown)


pladaria (approved):

src/tooltip.tsx
```diff
@@ -349,6 +350,7 @@ const Tooltip: React.FC<Props> = ({
     }, [
         tooltip,
         targetRect,
+        tooltipRect,
```
atabel, line 353: not a big fan of having dependencies in useEffect that are not used inside it

atabel (approved):

PR #971, opened 2023-12-05
fix(Checkbox): wrong clickable area

atabel:
https://jira.tid.es/browse/WEB-1669

In global ecommerce, we have a list of checkboxes. Sometimes, when the user clicks on one, the below one is activated instead. This is caused by the clickable area of the checkbox:

image

The issue is caused by Inline margins:
image

The simpler fix I could think of is removing pointer-events from `Inline` parent `div`. I can't think of a case where this could be problematic. After this fix, the clickable area is the one of the `Checkbox` parent `div`, not the `Inline` children:

image

A simpler/elegant fix would be stop using margins and use flex gap in `Inline`, but we can't use it yet:
image

grid gap has better support, but not enough:
image

### DIFF
--- a/src/inline.css.ts
+++ b/src/inline.css.ts
@@ -5,7 +5,7 @@ const space = createVar();
 export const vars = {space};
 
 export const inline = style({
-    pointerEvents: 'none', // to avoid the negative margins to affect clickable areas
+    pointerEvents: 'none', // to prevent negative margins from affecting clickable areas
     flexDirection: 'row',
     gridAutoFlow: 'column',
     marginTop: `calc(${space} * -1)`,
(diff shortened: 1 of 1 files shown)


src/inline.css.ts
```diff
@@ -5,6 +5,7 @@ const space = createVar();
 export const vars = {space};
 
 export const inline = style({
+    pointerEvents: 'none', // to avoid the negative margins to affect clickable areas
```
marcoskolodny, line 8:
```suggestion
    pointerEvents: 'none', // to prevent Inline negative margins from affecting clickable areas
```

marcoskolodny (approved):

ivmos (approved):

pladaria (approved):
Perhaps we should review the tappeable area for touch devices. It is too small now

PR #972, opened 2023-12-05
feat(skin): update design tokens

tuentisre:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action


src/skins/vivo-new.tsx
```diff
@@ -24,6 +24,7 @@ export const palette = {
     orangeLight40: '#FFB84C',
     pink: '#EB3D7D',
     pepper: '#CC1F59',
+    pepper10: '#FFEBED',
```
marcoskolodny, line 27: @yceballost had to add this for vivo and vivo-new. Is this correct?

src/skins/o2-classic.tsx
```diff
@@ -114,6 +114,9 @@ export const getO2ClassicSkin: GetKnownSkin = () => {
             buttonDangerBackgroundHover: palette.pepper55,
             buttonLinkBackgroundSelected: palette.o2SkyBlueLight10,
             buttonLinkBackgroundInverseSelected: applyAlpha(palette.white, 0.2),
+            buttonLinkDangerBackgroundSelected: palette.pepper10,
```
marcoskolodny, line 117: @yceballost in o2-classic the new tokens didn't exist. I copied the ones from o2. Verify if this makes sense

src/button.css.ts
```diff
@@ -421,23 +421,23 @@ const dangerLinkInverse: ComplexStyleRule = [
     link,
     sprinkles({
         color: vars.colors.textLinkDanger,
```
marcoskolodny, line 423: @yceballost this should be textLinkDangerInverse according to specs, but that token doesn't exist :(

src/skins/o2-classic.tsx
yceballost: we haven't support o2-classic in mistica-design for a long time ago 😬

src/button.css.ts
yceballost: I think textLinkDangerInverse is not needed 🤔

PR #978, opened 2023-12-13
fix(Image,Video): remove `fallbackVar` from components code

pladaria:
For some reason, importing `fallbackVar` in component file includes `chalk` and all their dependencies in frontend bundles. 

I noticed that because webapp build failed because `chalk` uses modern JS, so the `es-check` step failed.

I've implemented an alternative function `fallbackStyles` that accepts vars and regular styles. Check the test cases.

WEB-1678


src/image.tsx
```diff
@@ -151,7 +151,8 @@ export const ImageContent = React.forwardRef<HTMLImageElement, ImageProps>(
             ? '50%'
             : noBorderRadius
             ? '0px'
-            : fallbackVar(styles.vars.mediaBorderRadius, vars.borderRadii.container);
+            : fallbackStyles(styles.vars.mediaBorderRadius, vars.borderRadii.container);
```
atabel, line 154: an alternative fix would be to move these fallbackVar calls to the image.css.ts and video.css.ts files instead

atabel (approved):

src/image.tsx
pladaria: I tried but this is not so simple. Some components expect the borderRadius to be passed via props and using a className would require a bigger refactor

atabel: Ok, we can merge with this approach, I've created a ticket to review this:
https://jira.tid.es/browse/WEB-1679

PR #983, opened 2023-12-19
feat(Storybook): upgrade to version 7.6.6

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-811)


.github/actions/audit-accessibility/main.js
```diff
@@ -30,9 +30,6 @@ process.on('unhandledRejection', (error) => {
 
 const getStories = () => {
     console.log('Extracting stories information');
-    // creates "stories.json" file with stories information
```
marcoskolodny, line 33: This is not neccesary in Storybook 7, because `stories.json` is created automatically when building storybook.

.storybook/main.js
```diff
@@ -67,20 +67,8 @@ if (shouldIncludePrivateStories) {
...
 
 module.exports = {
-    core: {
-        builder: 'webpack5',
-    },
-
     stories,
     addons: [
-        {
-            name: '@storybook/addon-postcss',
```
marcoskolodny, line 77: This plugin is not compatible with storybook 7

src/__screenshot_tests__/inline-screenshot-test.tsx
```diff
@@ -8,7 +8,7 @@ test('Inline ', async () => {
 });
 
 test('Inline wrap', async () => {
-    await openStoryPage({id: 'layout-inline--wrap'});
+    const page = await openStoryPage({id: 'layout-inline--wrap'});
```
marcoskolodny, line 11: Using Puppeteer's page instead of our wrapper was causing screenshots to be taken too soon (before the story finishes loading)

src/__screenshot_tests__/navigation-bar-screenshot-test.tsx
```diff
@@ -16,7 +16,7 @@ test.each`
 
     const image = await page.screenshot();
 
-    expect(image).toMatchImageSnapshot();
+    expect(image).toMatchImageSnapshot({failureThreshold: 0.00001});
```
marcoskolodny, line 19: This test seems unstable because of the small picture:

unstable

src/__screenshot_tests__/tabs-screenshot-test.tsx
```diff
@@ -55,23 +55,25 @@ test('Tabs with long text and icon', async () => {
 });
 
 test('Tabs selected line appears properly', async () => {
-    await openStoryPage({
+    const page = await openStoryPage({
         id: 'components-tabs--default',
         device: 'MOBILE_IOS',
     });
 
+    const tabs = await screen.findByRole('tablist');
```
marcoskolodny, line 63: Looks like storybook 7 is a bit slower than storybook 6 when loading stories. If we don't wait until the tabs are added to the DOM, `page.evaluate()` doesn't find the tab elements

src/__screenshot_tests__/form-select-screenshot-test.tsx
```diff
@@ -20,7 +20,7 @@ test('Select - appears properly with selected option on DESKTOP', async () => {
         device: 'DESKTOP',
     });
 
-    const field = await screen.findByLabelText('Select a fruit');
+    const field = await screen.findByRole('button');
```
marcoskolodny, line 23: I don't know why, but if we try to do `await field.click()` without this change, the test fails because of the element being detached from the DOM:

image

.babelrc.json
atabel, line 1: why do you need a babelrc file? we don´t use babel in mistica repo, we use swc instead

.storybook/main.js
```diff
@@ -36,7 +36,7 @@ const addVanillaExtractSupport = (config) => {
         }
     });
 
-    config.plugins.push(new VanillaExtractPlugin(), new MiniCssExtractPlugin());
+    config.plugins.push(new VanillaExtractPlugin(), new MiniCssExtractPlugin({ignoreOrder: true}));
```
atabel, line 39: what does this change do?

.storybook/main.js
atabel: hmmm, but we need it, there isn't any alternative?

package.json
```diff
@@ -112,6 +114,7 @@
         "es6-promise-pool": "^2.5.0",
         "eslint": "^8.11.0",
         "eslint-plugin-mistica-local-rules": "workspace:*",
+        "eslint-plugin-storybook": "^0.6.15",
```
atabel, line 117: what's this?

src/__screenshot_tests__/form-select-screenshot-test.tsx
atabel, line 23: better:
```suggestion
    const field = await screen.findByRole('button', {name: 'Select a fruit'});
```

package.json
marcoskolodny: It was installed automatically when upgrading storybook version. I don't think we need it, so maybe I can remove it

.storybook/main.js
marcoskolodny: Basically, ignore warnings about order of imports in bundles. Without it, there are a lot of warnings when building the storybook:

image

.babelrc.json
marcoskolodny: It was automatically added when upgrading storybook's version. I've removed it

.storybook/main.js
marcoskolodny: Discussed offline. It looks like we don't need the plugin anymore. Storybook is adding the postcss config by default

src/__screenshot_tests__/form-select-screenshot-test.tsx
marcoskolodny: updated

.storybook/main.js
atabel: hmmm... and why do we have those warnings?

marcoskolodny: Discussed offline. We believe it's safe to ignore these warnings for the storybook. I've added a comment in the code providing a link where there is an explanation of the reason this happens ([Link](https://stackoverflow.com/questions/51971857/mini-css-extract-plugin-warning-in-chunk-chunkname-mini-css-extract-plugin-con/67579319#67579319))

atabel (approved):

package.json
yceballost, line 1: upgrade playroom? 🥺

yceballost (approved):

pladaria (approved):

pladaria:
nice!

PR #986, opened 2023-12-20
fix(Inline): Workaround for iOS to avoid cutting elements like Chips inside Inline containers

pladaria:
|before|after|
|-|-|
|image|image|


src/chip.tsx
```diff
@@ -144,7 +146,7 @@ const Chip: React.FC<ChipProps> = (props: ChipProps) => {
             )}
             paddingLeft={paddingLeft}
             paddingRight={paddingRight}
-            {...getPrefixedDataAttributes(dataAttributes, 'Chip')}
+            dataAttributes={dataAttributes}
```
pladaria, line 149: This wasn't working. Spreading data attributes over the Box resulted in ignored props

src/inline.tsx
```diff
@@ -63,9 +67,24 @@ const Inline: React.FC<Props> = ({
...
-            {React.Children.map(children, (child) => (!!child || child === 0 ? <div>{child}</div> : null))}
+            {React.Children.map(children, (child) =>
+                !!child || child === 0 ? (
+                    <div
+                        style={{
+                            // Hack to fix https://jira.tid.es/browse/WEB-1683
+                            // In iOS the inline component sometimes cuts the last line of the content
+                            paddingBottom:
+                                isIos(platformOverrides) && !isRunningAcceptanceTest(platformOverrides)
+                                    ? 0.5
```
pladaria, line 80: on my tests, 0.5 seems to be enough to fix the issue

src/inline.tsx
```diff
@@ -1,9 +1,12 @@
+'use client';
```
atabel, line 1: :(
Ideally, `Inline` shouldn't need `"use client"`

src/chip.tsx
atabel: hmmm... I think this is still wrong, because Box is applying `getPrefixedDataAttributes` again, so you'll end up with `data-data-xxx` attributes

pladaria: oh, right!

atabel (approved):

src/inline.tsx
pladaria: we can remove the "use client" in the next iteration and put the 1px padding unconditionally when flex-gap is not supported

marcoskolodny (approved):

PR #990, opened 2023-12-28
fix(Carousel): avoid vertical scroll caused by bullets

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1676)


src/carousel.css.ts
```diff
@@ -28,11 +31,13 @@ const bulletBase = style([
 const bulletActiveBase = style([
     bulletBase,
     {
-        transform: 'scale(1.25)', // 10px
+        width: BULLET_DESKTOP_SIZE * 1.25,
+        height: BULLET_DESKTOP_SIZE * 1.25,
```
atabel, line 35: with this change you are breaking the `transition: transform`, and animating width/height is usually slower than `transform: scale`. A posible alternative solution would be adding a small bottom padding to bullets container.

marcoskolodny: Oops, good catch. I didn't notice the issue with the animation. Updated, thanks!

atabel (approved):

PR #994, opened 2024-01-09
fix(BrandLoadingScreen): issue with lottie animation when CSP is enabled

atabel:


pladaria (approved):
perhaps we should ask design to export lotties without embedded code (not sure whether the app has that export option)

yceballost:
> perhaps we should ask design to export lotties without embedded code (not sure whether the app has that export option)

This is the config in Lottie plugin in After Effects
image

PR #997, opened 2024-01-10
feat(Dialogs): Improve implementation. Create `useDialog` hook. Deprecate `alert`, `confirm` and `dialog` functions

pladaria:
WEB-1646

As you can see, almost no changes in style or layout, just logic.

## Notable changes
* Removed the old `DialogRoot` component and refactored its logic inside the `ModalDialog` component.
* The logic has been greatly simplified, the previous implementation was quite messy.
* New `DialogRoot` component inside `dialog-context.tsx` file. This component lazy-imports the dialog code when needed. This component is now a small context that contains the current dialog props and very little logic.
* New `useDialog` hook provides the `alert`, `confirm` and `dialog` functions. For backwards compatibility, those functions are directly exported when the `ThemeContextProvider` is mounted so they can be called without using the hook
* While the component is transitioning/animating it has the `aria-hidden` attribute set to true

## To do
* [x] Verify native implementation
    * [x] iOS
    * [x] Android


src/dialog-context.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * @deprecated Use useDialog to get this function
+ */
+export const confirm = (params: ConfirmProps): void => currentDialogFunctions.confirm(params);
+
+/**
+ * @deprecated Use useDialog to get this function
+ */
+export const dialog = (params: ExtendedDialogProps): void => currentDialogFunctions.dialog(params);
```
pladaria, line 82: I've deprecated this because you should get those function from the `useDialog` hook, but I find this very convenient so I'm not sure whether to keep or remove the deprecation :/

src/__tests__/dialog-test.tsx
```diff
@@ -40,6 +45,18 @@ test('renders alert dialog correctly when alert function called', async () => {
...
 
+test('works with nested theme context providers', async () => {
+    render(
+        <ThemeContextProvider theme={makeTheme()}>
+            <ThemeContextProvider theme={makeTheme()} />
+        </ThemeContextProvider>
+    );
+    act(() => {
+        alert(alertProps);
+    });
```
atabel, line 56: I think the tests should use the hook instead of the deprecated function

atabel (approved):

src/__tests__/dialog-test.tsx
atabel: the same for stories

src/dialog.tsx
```diff
@@ -20,45 +20,39 @@ import * as styles from './dialog.css';
...
     cancelText?: string;
     onCancel?: () => void;
 }
-interface ExtendedDialogProps extends BaseDialogProps {
+
+export interface ExtendedDialogProps extends BaseDialogProps {
+    icon?: React.ReactElement;
     subtitle?: string;
     extra?: React.ReactNode;
-    forceWeb?: boolean;
```
marcoskolodny, line 53: Isn't this a breaking change? Should we deprecate them instead?

pladaria: strictly speaking, yes, but there are 0 usages of this in webapp

PR #999, opened 2024-01-11
fix(Carousel): remove bottom padding when withBullets=false

atabel:


src/__screenshot_tests__/carousel-screenshot-test.tsx
```diff
@@ -18,6 +18,20 @@ test('Carousel mobile', async () => {
     expect(await page.screenshot()).toMatchImageSnapshot();
 });
 
+test('Carousel mobile without bullets', async () => {
```
atabel, line 21: this case was not covered

pladaria (approved):

PR #1001, opened 2024-01-12
feat(Text): prioritize wrapping before breaking if word fits in container

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1696)

### DIFF
--- a/src/text.css.ts
+++ b/src/text.css.ts
@@ -1,6 +1,8 @@
 import {style, createVar} from '@vanilla-extract/css';
 import * as mq from './media-queries.css';
 
+import type {ComplexStyleRule} from '@vanilla-extract/css';
+
 const mobileSize = createVar();
 const desktopSize = createVar();
 const mobileLineHeight = createVar();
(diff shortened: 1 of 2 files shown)


src/text.css.ts
```diff
@@ -17,15 +19,8 @@ export const vars = {
...
-        'not (overflow-wrap: anywhere)': {
-            // "overflow-wrap: anywhere" is not supported in Safari
-            // "word-break: break-word" has the same effect as "word-break: normal" and "overflow-wrap: anywhere",
-            // regardless of the actual value of the overflow-wrap property.
-            wordBreak: 'break-word',
-        },
-    },
-});
+    wordBreak: ['keep-all', 'auto-phrase'],
+} as ComplexStyleRule);
```
pladaria, line 23: just curious, why is this type cast needed? And why wasn't needed before?

marcoskolodny: Because we are now using auto-phrase, an experimental property, and TS complains without the type cast.

More context here: https://github.com/vanilla-extract-css/vanilla-extract/discussions/740#discussioncomment-3199094

yceballost (approved):
gg!

pladaria (approved):

pladaria (approved):
Please add a private story with an example.

atabel (approved):

PR #1002, opened 2024-01-16
feat(Input fields): multiple bugs fixed and improved accesibility

marcoskolodny:
Bugs:

- [x] Field values' alignment/line height are wrong in different browsers.
- [x] Field padding/positioning is not considering border for calculations.
- [x] Prefix is not correctly aligned to value in iOS.
- [x] Calendar icon is not centered in date/date time/month fields.
- [x] Icons are not centered in credit card number field.
- [x] Native select options are not using font size.
- [x] Input fields screenshots in desktop are not including the entire field.
- [x] Pin field wrapper's width is not correct.
- [x] Long label is overlapping with the end icon.
- [x] Password field's placeholder has incorrect font family.
- [x] Credit card number field's icon is not showing properly in some devices.
- [x] Clicking on prefix in text field is not focusing the input.
- [x] Label has wrong right padding (12px vs 16px).
- [x] Prefix has wrong right padding (16px vs 12px). 
- [x] Prefix has wrong color when using inverse theme.
- [x] Date/datetime/month fields with readOnly/disabled are not showing value. 
- [x] Date fields have extra margin in Chrome (Android)

Tasks:

- [x] Make helperText describe input field by using `aria-describedby`.
- [x] Make all fields adapt to font size and remove harcoded values everywhere
  - Try to declare some variables in a single place so that it's easier to make updates in the future 
- [x] Add DataAttributes to every field and use it in stories and screenshot tests.
- [x] Update icon sizes for all fields.
- [x] Add screenshot tests to test big font in different fields.


src/__screenshot_tests__/__image_snapshots__/form-select-screenshot-test-tsx-select-display-all-options-with-native-select-1-snap.png
marcoskolodny, line 1: Native select was not considering the font size at all

src/__screenshot_tests__/__image_snapshots__/input-fields-screenshot-test-tsx-credit-card-number-field-1-snap.png
marcoskolodny, line 1: Credit card field icon was not vertically centered

src/__screenshot_tests__/__image_snapshots__/input-fields-screenshot-test-tsx-date-field-1-snap.png
marcoskolodny, line 1: date/datetime/month fields' icon was not centered

src/__screenshot_tests__/__image_snapshots__/input-fields-screenshot-test-tsx-phone-number-field-2145678901-in-vivo-skin-1-snap.png
marcoskolodny, line 1: The screenshots that use a desktop device were not including the entire field. This happened because the input element doesn't contain the border, and it was fixed by taking the screenshot on the element that contains the data-testid attribute

src/__screenshot_tests__/__image_snapshots__/input-fields-screenshot-test-tsx-pin-field-1-snap.png
marcoskolodny, line 1: Pin field had a div as wrapper that didn't have `width: 'fit-content'`. This caused the extra space on the right

src/credit-card-number-field.css.ts
```diff
@@ -5,12 +5,31 @@ export const flip = style({
...
-export const flipFront = sprinkles({position: 'absolute'});
+const baseIconContainer = style([
+    sprinkles({
+        display: 'flex',
+        alignItems: 'center',
+        justifyContent: 'center',
+        height: '100%',
+    }),
+    {
+        backfaceVisibility: 'hidden',
```
marcoskolodny, line 16: Not adding `backfaceVisibility: 'hidden'` to flipFront was causing weird behaviour in some devices

src/icons/icon-amex.tsx
```diff
@@ -2,7 +2,7 @@ import * as React from 'react';
 
 type Props = {
     role?: string;
-    size?: number;
+    size?: number | string;
```
marcoskolodny, line 5: some icons were only allowing size as `number`, while others allow `number | string`

src/text-field-base.tsx
```diff
@@ -208,25 +199,23 @@ export const TextFieldBase = React.forwardRef<any, TextFieldBaseProps>(
...
 
-        const commonStyles = sprinkles({
-            paddingRight: endIcon ? 0 : 16,
-            paddingLeft: prefix ? 0 : startIcon ? 48 : 12,
-        });
-
         /* Workaround to avoid huge bullets on ios devices (-apple-system font related) */
-        const fontFamily = rest.type === 'password' ? 'Lucida Grande, Arial, sans-serif' : 'inherit';
+        const fontFamily =
+            rest.type === 'password' && characterCount > 0 ? 'Lucida Grande, Arial, sans-serif' : 'inherit';
```
marcoskolodny, line 218: password field placeholder was using wrong font family

src/text-field-components.tsx
```diff
@@ -129,17 +135,14 @@ export const FieldContainer: React.FC<FieldContainerProps> = ({
                 [styles.disabled]: disabled,
             })}
             onClick={(e) => {
-                // workaround for the multiline case, where the textarea has a margin. We want to focus
-                // the input when the user clicks anywhere in the container (like in the label)
-                if (multiline) {
-                    e.currentTarget.querySelector('textarea')?.focus();
-                }
+                // We want to focus the input when the user clicks anywhere in the container (like in the label or the prefix)
+                e.currentTarget.querySelector(multiline ? 'textarea' : 'input')?.focus();
```
marcoskolodny, line 139: If TextField had a prefix and you clicked on it, it was not focusing the input element

src/cvv-field.tsx
```diff
@@ -95,6 +98,8 @@ const CvvField: React.FC<CvvFieldProps> = ({
         onChangeValue,
     });
 
+    const iconSize = pxToRem(16);
```
marcoskolodny, line 101: Icon size was smaller than in other input fields. Double check if this makes sense @yceballost

src/select.tsx
```diff
@@ -294,7 +295,8 @@ const Select: React.FC<SelectProps> = ({
             }
         },
     };
-    const {isDesktopOrBigger} = useScreenSize();
+
+    const iconSize = pxToRem(20);
```
marcoskolodny, line 299: Icon size was smaller than in other input fields. Double check if this makes sense @yceballost

src/text-field-base.css.ts
```diff
@@ -62,33 +110,17 @@ globalStyle(`${fullWidth} > div`, {
...
-});
-
-export const textAreaWithoutLabel = style({
-    paddingTop: 16,
-});
-
 export const input = style([
     sprinkles({
         position: 'relative',
-        height: '100%',
```
marcoskolodny, line 91: Setting the height of an HTML `<input/>` element is a bad idea. It can mess up the line height and the text alignment (and it behaves in different ways depending on the browser).

More context here: [Link](https://stackoverflow.com/questions/42682530/html-input-field-height-different-in-different-browsers/43219553#43219553)

src/__screenshot_tests__/__image_snapshots__/form-select-screenshot-test-tsx-select-appears-properly-with-long-selected-option-on-mobile-ios-1-snap.png
marcoskolodny, line 1: This change in many screenshots is caused by removing the `height` prop from the `<input/>` elements. Before this, in browsers like chrome, the line height was being set to 22 pixels, not 24 (even though we were using `lineHeight: 24px`). In [this comment](https://github.com/Telefonica/mistica-web/pull/1002/files#r1463336174) there is an explanation about why this happens.

Screenshot of the input container before this PR, where you can see that the height was 22px:
image

With the changes in this PR, now we can see that the line height is correct:
image

src/__screenshot_tests__/__image_snapshots__/input-fields-screenshot-test-tsx-text-field-long-label-1-snap.png
marcoskolodny, line 1: Notice the change in the label. According to the specs, the left padding in the input fields should be equal to 12 pixels, while the right one should be 16 pixels.

This was correct for the field's value, but the label had only 12 pixels for the right padding

src/__screenshot_tests__/__image_snapshots__/input-fields-screenshot-test-tsx-date-field-increased-font-size-1-snap.png
yceballost, line 1: Use the same hover effect to calendar icon because it is a interactive element
image

also for dateTime

src/__screenshot_tests__/__image_snapshots__/input-fields-screenshot-test-tsx-search-field-2-snap.png
yceballost, line 1: Use the same hover effect to "X" icon because it is a interactive element
image

src/text-field-base.css.ts
```diff
@@ -196,53 +283,37 @@ export const endIcon = style([
...
+        // passthrough click events to the input
+        pointerEvents: 'none',
     },
 ]);
 
 export const prefix = style([
     sprinkles({
-        paddingLeft: 12,
+        paddingLeft: fieldLeftPadding,
         paddingRight: 16,
```
yceballost, line 302:
```suggestion
        paddingRight: 12,
```

space between prefix and value is 12

I think this value may cause problems with font scaling. Use rem instead?

Just curiosity, why do you use padding instead of flexbox stuff? gap between prefix and value

src/text-field-base.tsx
```diff
@@ -235,49 +224,61 @@ export const TextFieldBase = React.forwardRef<any, TextFieldBaseProps>(
...
                             styles.prefix,
-                            hasLabel ? styles.prefixWithLabel : styles.prefixWithoutLabel
+                            hasLabel ? styles.inputWithLabel : styles.inputWithoutLabel
                         )}
                         style={{
                             opacity: inputState === 'default' ? 0 : 1,
-                            alignSelf: prefixAlignSelf,
                         }}
                     >
                         <Text3 color={vars.colors.textSecondary} regular wordBreak={false}>
```
yceballost, line 249: Prefix color should not change in inverse
image

src/__stories__/text-field-story.tsx
```diff
@@ -64,22 +64,20 @@ export const Controlled: StoryComponent<TextFieldControlledArgs> = ({
...
+                    <TextField
+                        value={rawValue}
+                        onChangeValue={(value, rawValue) => {
+                            setValue(value);
+                            setRawValue(rawValue);
+                        }}
+                        name="text"
+                        autoComplete="off"
+                        dataAttributes={{testid: 'text-field'}}
+                        maxLength={maxLength ? 200 : undefined}
```
yceballost, line 76: should we enable maxLenght helper text for text fields without multiline? 

what do you think @aweell 

image

src/text-field-base.css.ts
yceballost, line 1: Why this animation in label? 😬

https://github.com/Telefonica/mistica-web/assets/6722153/d7fe5bff-545a-46fb-99a2-49f95aab776c

yceballost (changes requested):

src/phone-number-field.tsx
yceballost, line 1: When prefix is enabled, the value positionY is different (2px)
image
image

yceballost (changes requested):

src/phone-number-field.tsx
yceballost: my bad, this test are in master TT

src/cvv-field.tsx
yceballost: it is expected

src/select.tsx
yceballost: it is expected

src/date-field.tsx
yceballost, line 1: Date picker value is not appearing when you use readonly or disable props
https://mistica-p1bsy5q85-tuentisre.vercel.app/?path=/story/components-input-fields-datefield--controlled&args=disabled:!true;readOnly:!true

(same for DateTimeField)

yceballost (changes requested):

src/__screenshot_tests__/__image_snapshots__/input-fields-screenshot-test-tsx-date-field-increased-font-size-1-snap.png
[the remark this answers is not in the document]
marcoskolodny: This is not possible. In some browsers, like chrome, the value is actually not plain text, but 3 independent focusable regions. This allows you to navigate between day/month/year with the tab key for example.

About hiding the icon, that's very tricky, because we always render a native date picker if the browser supports it. Each browser has different styles, and we don't choose that.

marcoskolodny: Actually, in date fields the calendar is not the only interactive region. The entire container opens the date picker, so it doesn't make sense to add the hover effect to the icon (similar to the arrow in select field)

src/text-field-base.css.ts
marcoskolodny: We are not using rem values for paddings that separate elements anywhere in Mística.

We could do it, but it will leave almost no space for the field's value when you increase the font size.

We can think about this a bit more in a second iteration and maybe fix this in all the components, since the problem is not happening only in input fields.

src/__stories__/text-field-story.tsx
marcoskolodny: I remember we had this conversation a few months ago and decided not to do that.

I believe that is a separate task and we should evaluate the side effects carefully before adding it.

https://teams.microsoft.com/l/message/19:380a30ed56a946c8bad31847048d0c7e@thread.tacv2/1699889469262?groupId=e265fe99-929f-45d1-8154-699649674a40&parentMessageId=1699889469262&tenantId=9744600e-3e04-492e-baa1-25ec245c6f10

src/text-field-base.css.ts
marcoskolodny: updated padding according to specs (12 instead of 16)

src/__screenshot_tests__/__image_snapshots__/input-fields-screenshot-test-tsx-search-field-2-snap.png
marcoskolodny: fixed

src/text-field-base.css.ts
marcoskolodny: That happens because Storybook addon for changing the font initially sets the fontsize to its default value (16px), and after first render it's changed to the one you chose in there.

Given that fontSize changes, the translate position for the label will also change (to center the label in the new field's container with bigger height), and translate property is animated for the label.

Long story short, this only happens in storybook because of font size addon. If you test this in iphone for example (where the font size is taken from the device instead of the addon), you won't see the animation when you load the story.

src/text-field-base.tsx
marcoskolodny: fixed

src/date-field.tsx
marcoskolodny: It should be fixed now

src/__screenshot_tests__/__image_snapshots__/dialog-screenshot-test-tsx-dialog-device-desktop-action-dialog-1-snap.png
marcoskolodny, line 1: The alignment of all fields was wrong, because it was applying padding and positioning the label WITHOUT taking into account the extra pixel added by the border (because we have `box-sizing: border-box`)

src/__stories__/date-field-story.tsx
yceballost, line 1: I think we can remove placeholder from story controls

src/__stories__/date-time-field-story.tsx
yceballost, line 1: I think we can remove placeholder from story controls

yceballost (approved):
nice improvements ✨

src/__stories__/date-field-story.tsx
marcoskolodny: removed

src/__stories__/date-time-field-story.tsx
marcoskolodny: removed

src/date-time-picker.tsx
```diff
@@ -109,34 +109,34 @@ const DateTimePicker: React.FC<DateTimePickerProps> = ({withTime, mode, isValidD
...
-                <IconButton
-                    aria-label={texts.clearButton}
-                    size={32}
-                    onPress={() => {
-                        setValue('');
-                    }}
-                >
-                    <IconCloseRegular />
-                </IconButton>
+                <>
```
pladaria, line 112: unneeded fragment?

src/text-field-base.css.ts
```diff
@@ -103,56 +165,64 @@ export const input = style([
...
                 // The background can not be overriden, but we can delay the background color transition to avoid the change
                 transitionProperty: 'background-color',
                 transitionDelay: '99999s',
             },
+
             // iOS date/time fields are centered by default, but we want them left aligned
             '&::-webkit-date-and-time-value': {
                 textAlign: 'left',
+
+                // In Android, Chrome adds margin to the date's value, causing the input to be taller than expected
```
pladaria, line 214: :O

src/text-field-base.tsx
```diff
@@ -1,19 +1,67 @@
...
+const isValidInputValue = (value?: string, inputType?: React.HTMLInputTypeAttribute) => {
+    if (!inputType) {
+        return true;
+    }
+
+    const input = document.createElement('input');
+    input.type = inputType;
+    input.value = value || '';
+    return input.value !== '';
+};
```
pladaria, line 28: Should empty string be considered as valid?

```suggestion
const isValidInputValue = (value: string = '', inputType?: React.HTMLInputTypeAttribute) => {
    if (!inputType) {
        return true;
    }

    const input = document.createElement('input');
    input.type = inputType;
    input.value = value;
    return input.value === value;
};
```

pladaria (approved):
nice!

src/text-field-base.tsx
marcoskolodny: No. This is for date inputs, and empty strings are not valid there.

aweell:
The icon size of the textField is not growing based on font size

Screenshot 2024-01-30 at 16 18 48

aweell (approved):
Great work!

src/pin-field.tsx
```diff
@@ -187,14 +189,20 @@ const PinInput = ({
...
-                                [styles.passwordInput]: hideCode,
-                            }
-                        )}
+                        className={classNames(textFieldStyles.input, styles.input, {
+                            [styles.passwordInput]: hideCode,
+                        })}
+                        /**
+                         * We need to override IntegerField styles because PinField has a different
+                         * style than other inputs (less margin, fontSize and height).
+                         */
```
atabel, line 198: then, maybe we shouldn't be using `textFieldStyles.input` here?

atabel (approved):
Nice PR, lots of bugfixes!

PR #1004, opened 2024-01-23
fix(Feedback): CLS after hydrate SSRed feedbacks

atabel:
WEB-1673


yceballost (changes requested):

src/feedback.css.ts
```diff
@@ -33,41 +37,65 @@ export const innerContainer = style({
...
-const feedbackDataAppearActive = style({
-    transitionProperty: 'opacity, transform',
-    transitionDuration: '1s',
-    transitionTimingFunction: 'cubic-bezier(0.215, 0.61, 0.355, 1)',
-    opacity: 1,
-    transform: 'none',
+const textAppearAnimationDesktop = keyframes({
+    from: {
+        opacity: 0,
+        transform: 'translateY(40px)',
```
marcoskolodny, line 68: Maybe we could use variables for these hardcoded numbers. If in the future we need to change this, we would need to review each line of this file carefully otherwise.

src/responsive-layout.tsx
```diff
@@ -41,7 +41,12 @@ export const InternalResponsiveLayout: React.FC<Props & {shouldExpandWhenNested?
                     internalVariant &&
                         internalVariant !== 'default' &&
                         styles.backgroundVariant[internalVariant],
-                    shouldExpandWhenNested && !fullWidth && styles.expandedResponsiveLayoutContainer
+                    {
+                        [styles.expandedResponsiveLayoutContainerMobile]:
+                            shouldExpandWhenNested === true && !fullWidth,
```
marcoskolodny, line 46:
```suggestion
                            shouldExpandWhenNested && !fullWidth,
```

marcoskolodny (approved):

src/responsive-layout.tsx
pladaria: `shouldExpandWhenNested` could be `"desktop"`, which is truthy. 

Perhaps a comment to communicate the intention would be useful

pladaria (approved):

src/responsive-layout.tsx
atabel: remember what I told you the other day @marcoskolodny? nobody writes `=== false` or `=== true` without a reason :D

marcoskolodny: Oooh, I missed the fact that it can be a string. My bad!

src/feedback.css.ts
atabel: this 40 isn't related with any other number/style in the file

src/feedback.tsx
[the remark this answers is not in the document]
atabel: should be fixed

yceballost (approved):

PR #1007, opened 2024-01-24
fix(Inline): improve fullWidth case

pladaria:
This branch against webapp:
https://github.com/Telefonica/webapp/actions/runs/7667071028

A single test failed, which is expected:
header-cards-single-screenshot-test-tsx-header-cards-single-two-buttons-brand-vivo-br-in-mobile-ios-1-diff


src/dialog.tsx
```diff
@@ -34,6 +34,8 @@ interface BaseDialogProps {
     destructive?: boolean;
     /** @deprecated this does nothing */
     forceWeb?: boolean;
+    /** @deprecated this does nothing */
+    showCancel?: boolean;
```
pladaria, line 38: unrelated, to keep dialog types compatible

src/__tests__/dialog-test.tsx
```diff
@@ -72,15 +72,17 @@ test('throws when we try to stack dialogs', async () => {
...
 
 test("throws when we don't instantiate theme", async () => {
     render(<TestComponent />);
 
-    expect(() => savedAlert(alertProps)).toThrow(
-        'Tried to show a dialog but the DialogRoot component was not mounted'
-    );
+    expect(() => {
+        act(() => savedAlert(alertProps));
+    }).toThrow('Tried to show a dialog but the DialogRoot component was not mounted');
```
pladaria, line 85: this fixes the unstable test

src/inline.css.ts
```diff
@@ -49,44 +45,34 @@ export const inline = style({
...
-    },
-]);
+export const fullWidth = style({
+    /**
+     * @deprecated - this should be "flex"
+     * Once changed, the usages that depend on this behavior should be migrated to use <Grid>
+     * See "inline-cases" private story
+     */
+    display: 'grid',
+});
```
atabel, line 61: I don't understand this change. With the previous code we were using flex in browsers that don't support grid

src/inline.css.ts
```diff
@@ -49,44 +45,34 @@ export const inline = style({
...
-        display: ['inline-flex', 'inline-grid'],
-        '@supports': {
-            [supportsFlexGap]: {
-                display: 'inline-flex',
-            },
-        },
-    },
-]);
+export const noFullWidth = style({
+    display: 'inline-grid',
```
marcoskolodny, line 69: Same thing that [Abel commented](https://github.com/Telefonica/mistica-web/pull/1007/files#r1470741091). Why removing the fallback to inline-flex if grid is not supported?

As far as I know, inline-grid is not supported in all the browsers that webapp supports (https://caniuse.com/?search=inline-grid)

src/inline.css.ts
pladaria: Added a fallback to flex

src/inline.css.ts
pladaria: added fallback

atabel (approved):

PR #1009, opened 2024-01-30
feat(Slideshow,Carousel): allow controlling the carousel from external component

atabel:
WEB-1715


src/__stories__/slideshow-story.tsx
```diff
@@ -45,3 +60,48 @@ Default.args = {
...
+type WithCarouselContextArgs = {numItems: number};
+
+export const WithCarouselContext: StoryComponent<WithCarouselContextArgs> = ({numItems}) => {
+    return (
+        <CarouselContextProvider>
+            <FixedFooterLayout
+                footer={
+                    <ResponsiveLayout>
+                        <Box paddingY={16}>
+                            <CarouselContextConsummer>
```
atabel, line 73: This is an example of use.
Alternatively, you could create a component and use `useCarouselContext()`

yceballost: Only in slideshow story? why not in carousel if its supported? Could be inside utils too

src/__stories__/slideshow-story.tsx
```diff
@@ -45,3 +60,48 @@ Default.args = {
...
+                                {({goNext, goPrev, bulletsProps}) => (
+                                    <Inline space="between" alignItems="center">
+                                        <ButtonLink onPress={goPrev}>Prev</ButtonLink>
```
yceballost, line 76:
```suggestion
                                        <ButtonLink onPress={goPrev} bleedLeft>Prev</ButtonLink>
```

src/__stories__/slideshow-story.tsx
```diff
@@ -45,3 +60,48 @@ Default.args = {
...
+                                        <PageBullets {...bulletsProps} />
+                                        <ButtonLink onPress={goNext}>Next</ButtonLink>
```
yceballost, line 78:
```suggestion
                                        <ButtonLink onPress={goNext} bleedRight>Next</ButtonLink>
```

src/__stories__/slideshow-story.tsx
yceballost: 'Prev' I know why you did 🌝

yceballost (approved):

src/__stories__/slideshow-story.tsx
yceballost: should the arrows be hidden in desktop using the context?

kydorn: why? afaik, the arrows are a "replacement" of the swipe mechanism, and that still applies for non-touch devices.
To hide the arrows maybe we should use a separate prop hideArrowsInDesktop?

src/carousel.tsx
```diff
@@ -73,6 +73,70 @@ export const PageBullets: React.FC<PageBulletsProps> = ({currentIndex, numPages,
...
+const defaultGoNext = defaultGoPrev;
+const defaultBulletProps = {currentIndex: 0, numPages: 0};
+
+type CarouselControls = {
+    goPrev: () => void;
+    goNext: () => void;
+    bulletsProps: PageBulletsProps;
+};
+
+const CarouselContext = React.createContext<CarouselControls>({
```
kydorn, line 88: it would be amazing if we could goToPage(X). 

In the CMS we could use that to adjust the preview to the page the operator is currently editing.

src/__stories__/slideshow-story.tsx
atabel: I can create an example story using the carousel, yes

src/__stories__/slideshow-story.tsx
yceballost: yes, could be a prop to hide this control for non touch devices

src/carousel.tsx
atabel: done

src/__stories__/slideshow-story.tsx
atabel: done

kydorn (approved):

src/carousel.tsx
```diff
@@ -73,6 +73,82 @@ export const PageBullets: React.FC<PageBulletsProps> = ({currentIndex, numPages,
     );
 };
 
+const defaultGoPrev = () => {
```
marcoskolodny, line 76: NIT: I think the name of this function should be related to the error itself (maybe something like `onMissingContextProvider`).

Then you could do:
```
const defaultGoPrev = onMissingContextProvider;
const defaultGoNext = onMissingContextProvider;
const defaultGoToPage = onMissingContextProvider;
```

src/__stories__/slideshow-story.tsx
```diff
@@ -45,3 +60,52 @@ Default.args = {
     loop: false,
     withBullets: true,
 };
+
+type WithCarouselContextArgs = {numItems: number};
+
+export const WithCarouselContext: StoryComponent<WithCarouselContextArgs> = ({numItems}) => {
+    return (
+        <CarouselContextProvider>
```
marcoskolodny, line 68: Maybe we should add some screenshot/acceptance tests to verify that the new functionality works?

atabel: I tried to cover it with a unit test, but jsdom doesn't support scroll, so it's imposible to cover with unit.
I could write an acceptance test

atabel: done

src/carousel.tsx
atabel: done (something similar)

marcoskolodny (approved):

PR #1011, opened 2024-01-31
fix(Button, ButtonLink): make icons grow with fontsize and align Link chevron to baseline

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1710).

There are some issues with alignment of the text in some brands, but that is caused by the height of letters and baseline in each font we use. I don't think there is much we can do about it. For example, we can see that aligning to center in some fonts is not causing the text to be aligned with the icon in a button:

Screenshot from 2024-01-31 14-19-40

Anyways, this PR fixes two things related to the aligment of content in buttons:
- When using ButtonLink, the chevron had some empty space on top of the svg to force the alignment in a particular font. I've fixed this by removing it and aligning the chevron to the baseline of the button's text.
- The button start/end icons' size were not being updated when changing the font size.


atabel:
>The button start/end icons' size were not being updated when changing the font size.

Do we really want this behavior?
Users increase font size to be able to read text, if we scale icons too, it's more like a zoom instead of a font scale, and it could end up with less space for the text, which will result in the oposite of what the user wanted (be able to read the text)

yceballost:
> > The button start/end icons' size were not being updated when changing the font size.
> 
> Do we really want this behavior? Users increase font size to be able to read text, if we scale icons too, it's more like a zoom instead of a font scale, and it could end up with less space for the text, which will result in the oposite of what the user wanted (be able to read the text)

I think the icon is not a decorative element in this case, the function of this icon is to discern between navigations and actions, so I think its ok to keep the scale

PR #1012, opened 2024-01-31
chore(Skins): add private story with all the components to test different skins

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1714)


src/__stories__/navigation-breadcrumbs-story.tsx
```diff
@@ -11,8 +12,8 @@ type Args = {
 
 export const Default: StoryComponent<Args> = ({inverse}) => {
     return (
-        <ThemeVariant isInverse={inverse}>
-            <div style={{background: inverse ? skinVars.colors.backgroundBrand : skinVars.colors.background}}>
+        <ResponsiveLayout isInverse={inverse} fullWidth>
```
marcoskolodny, line 15: unrelated change. This story should use ResponsiveLayout instead of setting the background by hand (like all the other stories).

src/__screenshot_tests__/__image_snapshots__/private-skin-components-screenshot-test-tsx-components-in-telefonica-inverse-1-snap.png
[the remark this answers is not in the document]
yceballost: yep, these components are not ready for inverse (we already reported to work on it in our side https://github.com/Telefonica/mistica-design/issues/736)

src/__private_stories__/skin-components-story.tsx
yceballost, line 1: missing:
- row
- boxedRow (with image to test small radius in images)
- snap cards
- carousel (arrow control maybe)
- snackbar (think this is important because critical variant could be not accessible in some cases 😬)
- skeletons (check contrast with some backgrounds inverse & alternative)
- navigation bars (for logo positioning)
- divider (use different tokens than borders is good to compare differences)
- titles (vivo has different sizes for example)
- texts
- all tags types (this is the most problematic component with contrast accessibility)
- tabs
- spinner (why not)

I prefer to remove buttons (and tags) from cards and include all buttons in separated line

src/__screenshot_tests__/__image_snapshots__/private-skin-components-screenshot-test-tsx-components-in-blau-inverse-1-snap.png
yceballost, line 1: can you also include "alternative" background in the tests?

yceballost (changes requested):

src/__screenshot_tests__/__image_snapshots__/private-skin-components-screenshot-test-tsx-components-in-blau-inverse-1-snap.png
atabel: we already have this story (with screenshots) to test different components over different background variants:
https://mistica-web.vercel.app/?path=/story/utilities-themevariant--components-over-different-theme-variants

src/__screenshot_tests__/private-skin-components-screenshot-test.tsx
```diff
@@ -0,0 +1,43 @@
+import {openStoryPage} from '../test-utils';
+
+const SKINS = ['Movistar', 'O2', 'Vivo', 'Vivo-new', 'Telefonica', 'Blau'] as const;
```
atabel, line 3: I'd remove Vivo from the list, we are not going to iterate that skin and will eventually be deprecated/replaced by vivo-new

src/__screenshot_tests__/private-skin-components-screenshot-test.tsx
```diff
@@ -0,0 +1,43 @@
...
+test.each(SKINS)('Components in %s', async (skin) => {
+    const page = await openStoryPage({
+        id: 'private-components-in-different-skins--default',
+        skin,
+    });
+
+    expect(await page.screenshot({fullPage: true})).toMatchImageSnapshot();
+});
+
+test.each(SKINS)('Components in %s (inverse)', async (skin) => {
```
atabel, line 14: we already have a story to test components over inverse/alternative backgrounds:
https://mistica-web.vercel.app/?path=/story/utilities-themevariant--components-over-different-theme-variants

I think we should use that story to test components that change over different background variants

src/__private_stories__/skin-components-story.tsx
marcoskolodny: Discussed offline.

**Added:**
- Row: it makes sense to test href chevron and divider
- Image inside accordion: to test if the border radius token is correct for media elements
- Skeleton: it makes sense to check contrast over inverse/alternative in different skins
- Titles: the font size/weight changes depending on the skin 
- Tags: the cards used only tested two types. It makes sense to put all the types in a separate example.

**Not added:**
- NavigationBar: this story is about skins, the positioning of the logo doesn't change depending on it.
- Logo: we already have a story that tests how the brand logo looks like on different skins.
- SnapCard: we included a DataCard, I don't see anything new that a snap card would add to the story.
- Carousel: nothing changes between skins (except of the bullets, which were already added to the story).
- Texts: all the components in the story use different Texts, there is no need to add a separate example. 
- Buttons: cards already have all the variants, not necessary. 
- Tabs: the only thing that changes is the control token, and that is tested by other components already.
- Spinner: it's the same in every brand, there is no point in adding it to this story.

src/__screenshot_tests__/private-skin-components-screenshot-test.tsx
marcoskolodny: removed

src/__screenshot_tests__/__image_snapshots__/private-skin-components-screenshot-test-tsx-components-in-blau-inverse-1-snap.png
marcoskolodny: discussed offline. We decided that it's better to add the theme variants in this new story too, because there are some components (like switch) that behave differently depending on the combination of the brand and the theme variant.

src/__screenshot_tests__/private-skin-components-screenshot-test.tsx
marcoskolodny: discussed offline. We decided that it's better to add the theme variants in this new story too, because there are some components (like switch) that behave differently depending on the combination of the brand and the theme variant.

atabel (approved):

pladaria (approved):
Thanks!

PR #1017, opened 2024-02-07
fix(Text): word wrapping is not working when truncate's value is greater than one

marcoskolodny:


src/text.css.ts
```diff
@@ -59,24 +59,7 @@ export const truncateToOneLine = style([
...
-        wordBreak: 'break-word',
-        '@supports': {
-            '(overflow-wrap: anywhere)': {
-                overflowWrap: 'anywhere',
-                wordBreak: 'normal',
-            },
-        },
-    },
-]);
+export const truncateToMoreThanOneLine = truncate;
```
pladaria, line 65: perhaps, instead of this, you could use the truncate style in the component

atabel (approved):

pladaria (approved):
lgtm

PR #1018, opened 2024-02-07
fix(Dialogs): Execute back navigation before accept/cancel callbacks

pladaria:


src/inline.css.ts
```diff
@@ -57,7 +57,7 @@ export const fullWidth = style({
      * Once changed, the usages that depend on this behavior should be migrated to use <Grid>
```
pladaria, line 57: Unrelated: somehow I got this commit wrong in the previous PR

brtbrt:
does this change affect the deprecated `confirm`? In that case, O2DE should change the behavior of some of its own components and hooks. Let me know so that we can sync when performing the bump on webapp 😉

PR #1020, opened 2024-02-08
fix(FeedbackScreen): inverse background color and overscroll color

atabel:


src/feedback.css.ts
```diff
@@ -46,6 +46,14 @@ export const innerContainer = style({
     },
 });
 
+export const feedbackScreenInverse = style({
```
atabel, line 49: this fixes this webapp case, where the parent component was setting a white background:
global-bundle-purchase-screenshot-test-tsx-test-bundle-purchase-3-diff

src/feedback.tsx
```diff
@@ -242,8 +243,12 @@ export const FeedbackScreen: React.FC<FeedbackScreenProps> = ({
     }
 
     return (
-        <div style={{position: 'relative'}}>
-            {isInverse && <OverscrollColor />}
+        <div style={{position: 'relative'}} className={isInverse ? styles.feedbackScreenInverse : undefined}>
+            {isInverse && (
+                <ThemeVariant isInverse>
```
atabel, line 248: without this, the overscoll color was white

pladaria (approved):

marcoskolodny (approved):

PR #1026, opened 2024-02-14
fix(Carousel): Add workaround for zero width webviews. Remount when webview size changes from 0 to any other value

pladaria:


marcoskolodny (approved):

src/index.tsx
```diff
@@ -114,7 +114,9 @@ export {
     PageBullets,
     CarouselContextProvider,
     useCarouselContext,
-    CarouselContextConsummer,
+    /** @deprecated Contains a typo. Please use CarouselContextConsumer */
```
atabel, line 117: :O

atabel (commented):
I'd try to push for a native fix. If Android devs can't find a solution, we may need to release this workaround

pladaria:
> I'd try to push for a native fix. If Android devs can't find a solution, we may need to release this workaround

Android ticket: https://jira.tid.es/browse/ANDROID-14348

As commented in teams, merging this while they investigate

PR #1031, opened 2024-02-22
fix: Proposal for fixes on snackbar and sheet so testing library does not warn

Marcosld:


src/sheet.tsx
```diff
@@ -165,7 +165,10 @@ type SheetProps = {
 
 const Sheet = React.forwardRef<HTMLDivElement, SheetProps>(({onClose, children, dataAttributes}, ref) => {
     const {texts} = useTheme();
-    const [modalState, dispatch] = React.useReducer(modalReducer, 'closed');
+    const [modalState, dispatch] = React.useReducer(
+        modalReducer,
+        process.env.NODE_ENV === 'test' ? 'open' : 'closed'
```
Marcosld, line 170: This should fix: https://github.com/Telefonica/global-checkout-webview/blob/a5dd8471303747fb9acae3a018fb5804c45507ac/web/utils/test-utils/index.tsx#L189

src/snackbar.tsx
```diff
@@ -42,7 +42,7 @@ export type ImperativeHandle = {
 const SnackbarComponent = React.forwardRef<ImperativeHandle, Props>(
     ({message, buttonText, duration, onClose, type, withDismiss = false, dataAttributes}, ref) => {
         const {texts} = useTheme();
-        const [isOpen, setIsOpen] = React.useState(false);
+        const [isOpen, setIsOpen] = React.useState(process.env.NODE_ENV === 'test');
```
Marcosld, line 45: This should fix https://github.com/Telefonica/global-checkout-webview/blob/371d2958beb0c61b69c804b441038ace7061bb2a/web/app/invoice-payment/components/__tests__/invoice-selection.test.tsx#L218

src/__tests__/snackbar-test.tsx
```diff
@@ -235,6 +235,33 @@ test('useSnackbar: happy case', async () => {
...
+        <ThemeContextProvider theme={makeTheme()}>
+            <ComponentWithSnackbar />
+        </ThemeContextProvider>
+    );
+
+    await userEvent.click(screen.getByRole('button', {name: 'Open Snackbar'}));
+
+    expect(await screen.findByText('any-message')).toBeInTheDocument();
+
+    await new Promise((resolve) => setTimeout(resolve, 1000));
```
pladaria, line 262: why? after a second the snackbar is still open. Perhaps you should manually close it

src/snackbar.tsx
pladaria: are you sure this is needed? we don't have issues in our snackbar related unit tests

src/__tests__/snackbar-test.tsx
Marcosld: This waiting time let's the 50ms `setTimeout` in the snackbar trigger. If you remove https://github.com/Telefonica/mistica-web/pull/1031#discussion_r1498888232 you will see the test fails.

Marcosld: This is the setTimeout I'm talking abount:
```
const openTimeout = setTimeout(() => {
                setIsOpen(true);
            }, 50);
```

pladaria: The timeout time when NODE_ENV==='test' is 0, and when the component is transitioning, it is set as aria-hidden, so, when you find an element inside the snackbar, it is because the snackbar has finished its transition and it is ready to be used.

See the other tests in this suite:
`await userEvent.click(await screen.findByRole('button', {name: 'action'}));`

Marcosld: Yeah, there you are querying for a button, but my test case does not have buttons, it's just a snackbar with some text :'(

Marcosld: I mean: From what I know `aria-hidden` attribute only hides elements from `byRole` queries @pladaria

Marcosld:
Closing because no activity

PR #1030, opened 2024-02-21
feat(Rows): add icon button as control

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1722)

PR #1034, opened 2024-02-23
feat(ToggleIconButton): create component

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1745)

**Note:** some changes in this PR (forcing aria-label and minor code style changes in iconButton) are also applied in [this PR](https://github.com/Telefonica/mistica-web/pull/1030). I needed these changes in both tasks, and the other pull request is still under revision. Once we merge one of them, the other one may have some conflicts


src/icon-button.tsx
```diff
@@ -356,4 +369,52 @@ export const BaseIconButton = (props: IconButtonProps): JSX.Element => {
...
+
+    const handleChange = () => {
+        if (checked === undefined) {
+            // if onChange is asynchronous, wait until it finishes and change the state if there was no error
+            const result = onChange?.(!checkedState);
+            if (result) {
+                return result.then(() => setCheckedState(!checkedState));
+            } else {
+                setCheckedState(!checkedState);
+            }
```
atabel, line 402: I think this should be equivalent
```suggestion
            Promise.resolve(result).then(() => setCheckedState(!checkedState));
```

atabel (approved):

src/icon-button.tsx
marcoskolodny: Discussed offline. This is not equivalent. In my code, the else statement returns void. In yours, it returns a promise.

This affects the RawIconButton component, because it checks if the result of calling onPress is a promise, and performs a setState if that is the case. We don't want that.

src/icon-button.tsx
```diff
@@ -156,25 +156,35 @@ export const RawDeprecatedIconButton = React.forwardRef<TouchableElement, Deprec
...
+type AriaProps = ExclusifyUnion<{'aria-label': string} | {'aria-labelledby': string}>;
+
 interface BaseProps {
     children?: undefined;
-    Icon: React.FC<IconProps>;
     trackingEvent?: TrackingEvent | ReadonlyArray<TrackingEvent>;
     dataAttributes?: DataAttributes;
     disabled?: boolean;
     showSpinner?: boolean;
-    'aria-label'?: string;
```
pladaria, line 166: this change is non backwards compatible

src/icon-button.tsx
```diff
@@ -356,4 +369,52 @@ export const BaseIconButton = (props: IconButtonProps): JSX.Element => {
...
+    TouchableElement,
+    ToggleIconButtonProps & {isOverMedia?: boolean}
+>(({checked, defaultChecked, checkedProps, uncheckedProps, onChange, ...props}, ref) => {
+    const [checkedState, setCheckedState] = React.useState(!!defaultChecked);
+
+    const handleChange = () => {
+        if (checked === undefined) {
+            // if onChange is asynchronous, wait until it finishes and change the state if there was no error
+            const result = onChange?.(!checkedState);
+            if (result) {
```
pladaria, line 398: if would be safer to check if result is thenable:
`if (result?.then) {}`

src/icon-button.tsx
```diff
@@ -356,4 +369,52 @@ export const BaseIconButton = (props: IconButtonProps): JSX.Element => {
...
+                return result.then(() => setCheckedState(!checkedState));
```
pladaria, line 399: I'd handle the rejection and do nothing (to me rejection would mean that the action was cancelled)

src/icon-button.tsx
```diff
@@ -356,4 +369,52 @@ export const BaseIconButton = (props: IconButtonProps): JSX.Element => {
...
+            } else {
+                setCheckedState(!checkedState);
```
pladaria, line 401: When toggling, I prefer to use the value from the setState callback `setCheckedState(state => !state)`

src/icon-button.tsx
```diff
@@ -356,4 +369,52 @@ export const BaseIconButton = (props: IconButtonProps): JSX.Element => {
...
+            ref={ref}
+            {...props}
+            {...(checked ?? checkedState ? checkedProps : uncheckedProps)}
+            onPress={handleChange}
+        />
+    );
+});
+
+export const ToggleIconButton = React.forwardRef<TouchableElement, ToggleIconButtonProps>((props, ref) => {
+    return <InternalToggleIconButton ref={ref} {...props} />;
```
pladaria, line 419: I'd expect to set the `component-name` here (and the other dataAttributes)

src/icon-button.tsx
marcoskolodny: The onChange callback returns void|undefined|promise<void>, so that is not needed (if it's not void/undefined, then it has to be a promise).

marcoskolodny: renamed result to promise

src/icon-button.tsx
marcoskolodny: Agree. Updated

src/icon-button.tsx
marcoskolodny: updated

src/icon-button.tsx
marcoskolodny: good catch. Updated

yceballost (approved):

aweell (approved):

pladaria (approved):

PR #1040, opened 2024-03-01
feat(Naked & Small Naked Card): Inner padding 

yceballost:
Right padding in card content helps to avoid mixing information with naked cards in carousel
image

image


src/__screenshot_tests__/__image_snapshots__/naked-card-screenshot-test-tsx-naked-card-with-extra-1-snap.png
atabel, line 1: do you think it could be problematic that the extra isn't full width now?

atabel (approved):

src/__screenshot_tests__/__image_snapshots__/naked-card-screenshot-test-tsx-naked-card-with-extra-1-snap.png
yceballost: We discussed this, but we consider that makes sense if you include text in extra the behavior should be the same as the rest of the content.

It's true that if you include boxed components like inputs you can see a little misalignment with the media, but we can live with this detail (it is not very common to use boxed components in extra btw)

marcoskolodny, line 1: I think it may be "cleaner" if we add the same padding to the left too. In my opinion, only adding it to the right makes it look a bit weird.

marcoskolodny (approved):

src/__screenshot_tests__/__image_snapshots__/naked-card-screenshot-test-tsx-naked-card-with-extra-1-snap.png
yceballost: Adding padding on the left causes misalignment with the left side of the image. The alignment on the left side is more "obvious" because on the right side not all the lines reach the end of the line

pladaria (approved):

PR #1041, opened 2024-03-04
feat(dialogs): Remove history hacks in dialog

pladaria:
Remove history hacks from dialog components to avoid interfering with similar hacks in apps.

* Remove history push and back navigation from dialog components
* It still listens to 'popstate' event to force a dismiss

Verification of this branch against the webapp brach that enables related skipped tests:
https://github.com/Telefonica/webapp/actions/runs/8232821830 (multiple successful runs)

TODO:
* [x] Perform some manual testing inside the app before merging (just to make sure the native dialogs work as expected with the webapp hacks and the new implementation)

Discarded options:
* use a modal `<dialog>` element. Discarded because it interferes with other elements with overlays (like selects or tooltips), fixing that would require a large refactor and the previous implementation cannot be removed because `<dialog>` is not supported before safari 15.4. Additionally, the Close Watcher with back seems to be broken in chrome (it works using ESC).
* keep the hack but try to workaround issues with webapp implementations. That is a hack to workaround other hacks. Better avoid this kind of madness.


marcoskolodny (approved):

brtbrt:
weren't you experiencing some problem with our `restrictNavigation` hook? i.e. the three "back" actions to navigate back

brtbrt (approved):

pladaria:
> weren't you experiencing some problem with our `restrictNavigation` hook? i.e. the three "back" actions to navigate back

I guess my previous implementation had issues. I rewrote this from scratch and it works as expected :)

atabel (approved):

PR #1047, opened 2024-03-11
chore(Storybook): upgrade version and fix flash when changing story

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1752)


package.json
marcoskolodny, line 1: It's better to see changes when hiding whitespaces: https://github.com/Telefonica/mistica-web/pull/1047/files?diff=split&w=1

PR #1067, opened 2024-04-03
feat(Accordion,Callout,Cards,EmptyState,Header,Hero,Row,NavigationBar): added titleAs prop to allow configuring heading level

atabel:
WEB-1796 

I've intentionally not made it configurable in some components:
- Full screen components like feedback screen: because there shouldn't be any reason to not use `<h1>` there
- Some components like tooltip/popover. I think it's fine to always render a simple `<span>` in those cases

### DIFF
--- a/src/accordion.tsx
+++ b/src/accordion.tsx
@@ -35,6 +35,7 @@ export const useAccordionContext = (): AccordionContextType => React.useContext(
 interface AccordionItemContentProps {
     children?: void;
     title: string;
+    titleAs?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
     subtitle?: string;
     asset?: React.ReactNode;
     content: React.ReactNode;
--- a/src/accordion.tsx
+++ b/src/accordion.tsx
@@ -35,7 +35,7 @@ export const useAccordionContext = (): AccordionContextType => React.useContext(
 interface AccordionItemContentProps {
     children?: void;
     title: string;
-    titleAs?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
+    titleAs?: string;
     subtitle?: string;
     asset?: React.ReactNode;
     content: React.ReactNode;
--- a/src/header.tsx
+++ b/src/header.tsx
@@ -29,6 +29,7 @@ type RichText = string | ({text: string} & OverridableTextProps);
 type HeaderProps = {
     pretitle?: RichText;
     title?: string;
+    titleAs?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
     description?: string;
     small?: boolean;
     dataAttributes?: DataAttributes;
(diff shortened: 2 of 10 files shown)


marcoskolodny (approved):

src/header.tsx
```diff
@@ -138,19 +144,21 @@ export const Header: React.FC<HeaderProps> = ({
...
     title: string;
+    titleAs?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
     description?: string;
     button?: RendersNullableElement<typeof ButtonPrimary> | RendersNullableElement<typeof ButtonSecondary>;
 };
 
-export const MainSectionHeader: React.FC<MainSectionHeaderProps> = ({title, description, button}) => {
+export const MainSectionHeader: React.FC<MainSectionHeaderProps> = ({
+    title,
+    titleAs = 'h1',
```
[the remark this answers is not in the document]
atabel: yes, I did it on purpose. I think it's a good default for this component

src/list.tsx
```diff
@@ -37,6 +37,7 @@ interface CommonProps {
     children?: void; // no children allowed
     headline?: string | React.ReactNode;
     title: string;
+    titleAs?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
```
aweell, line 40: Should always be a heading? I'm thinking for example in:

https://tinyurl.com/24aw45hx

atabel: hmmm... good point. I guess it make sense to allow anything in rows. Still, do you think h3 is a good default or should we default to span?

aweell: My first thought was to keep span as default but i don't have a stronger opinion rather than the sake of continuity.

src/accordion.tsx
```diff
@@ -35,6 +35,7 @@ export const useAccordionContext = (): AccordionContextType => React.useContext(
 interface AccordionItemContentProps {
     children?: void;
     title: string;
+    titleAs?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
```
aweell, line 38: In the case of the accordion (when a heading tag is used) [what WAI-Aria proposes](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/) is that the whole accordion button should be wrapped by the heading tag. All of the cases I've seen following this practice have only a title. I'm not sure if having multiple texts (title + subtitle) makes a difference here.

Regarding the available tags in the `titleAs` prop I also have doubts that all cases using an accordion should be represented in the heading hierarchy of the page.

For example, a [FAQ's sections should ideally use a definition list](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-faq/):

```
<dt>
  <dt><!-- Header -->
    <button>Is this an FAQ question?</button>
  </dt>
  <dd><!-- Panel -->
    Yes, it is.
  </dd>
  <dt>
    <button>What about this one?</button>
  </dt>
  <dd>
    Yap, this too.
  </dd>
  <!-- ... more questions ... -->
</dl>
```

src/list.tsx
atabel: ok, I think it makes sense. Changed the default to span

aweell (approved):

pladaria (approved):

PR #1068, opened 2024-04-04
feat(RowList, BoxedRowList, Inline): support list a11y role

atabel:
WEB-1719

`RowList` and `BoxedRowList` will have `role="list"` by default, and the children `Row`/`BoxedRow` elements will have the `role="listitem"` by default.

`Inline` will allow to setup `role="list"` and their children will have the `role="listitem` set automatically, as we were [already doing with `Stack`](https://github.com/Telefonica/mistica-web/blob/master/src/stack.tsx#L64)

Playroom to test the PR with a screen reader:
https://mistica-hxdstsm8w-tuentisre.vercel.app/playroom/preview/#?code=N4Igxg9gJgpiBcIA8AlCB3AMgSwM4BcA+AHQDsACc1Dc-bfAGxgF5iQAVept82XMAE7YADnQilWIACIx+Q0dnE9xABQGzczYAAoAlOWaFywAL4nyAehIUqadLS4s2nRjHIAmHn0EixEtjJyvooUniDkquq4mjr6hsZmltZIFnY4BNZkSABCEAAeMFBpeERklDn5hXYOrpIu3OHe8n6SgT4KSuGRGlp6Bkam5lZlVLkFRTR0tc6OHl6y7S0BC80hc12kaj2x-QlDyRZjVRjppaRZABIQQgBe4vgAhgwAyoIQDAzW5QCSpAzYpDcAneThA-wIPFwwgeYBYwAAHCYvpQqABhAAWIkIGJE5AAjCkccJkeUidjMcIPISKSS0TSieQAMzUrEjUn0inkAAsLOJbLpWIZAFZeciUr9-oCDldbvcnq9gR9rCAADQgfDomAAWzgiGyDAeAFcQCYgA

### DIFF
--- a/src/list.tsx
+++ b/src/list.tsx
@@ -652,9 +652,16 @@ const RowContent = React.forwardRef<TouchableElement, RowContentProps>((props, r
     );
 });
 
-export const Row = React.forwardRef<TouchableElement, RowContentProps>(({dataAttributes, ...props}, ref) => (
-    <RowContent {...props} ref={ref} dataAttributes={{'component-name': 'Row', ...dataAttributes}} />
-));
+export const Row = React.forwardRef<TouchableElement, RowContentProps>(
+    ({dataAttributes, role = 'listitem', ...props}, ref) => (
+        <RowContent
+            role={role}
+            {...props}
+            ref={ref}
+            dataAttributes={{'component-name': 'Row', ...dataAttributes}}
+        />
+    )
+);
 
 type RowListProps = {
     children: React.ReactNode;
@@ -663,7 +670,12 @@ type RowListProps = {
     dataAttributes?: DataAttributes;
 };
 
-export const RowList: React.FC<RowListProps> = ({children, ariaLabelledby, role, dataAttributes}) => {
+export const RowList: React.FC<RowListProps> = ({
+    children,
+    ariaLabelledby,
+    role = 'list',
+    dataAttributes,
+}) => {
     const lastIndex = React.Children.count(children) - 1;
     return (
         <div
@@ -730,7 +742,7 @@ type BoxedRowListProps = {
 export const BoxedRowList: React.FC<BoxedRowListProps> = ({
     children,
     ariaLabelledby,
-    role,
+    role = 'list',
     dataAttributes,
 }) => (
     <Stack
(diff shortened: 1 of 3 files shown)


src/list.tsx
```diff
@@ -652,9 +652,16 @@ const RowContent = React.forwardRef<TouchableElement, RowContentProps>((props, r
     );
 });
 
-export const Row = React.forwardRef<TouchableElement, RowContentProps>(({dataAttributes, ...props}, ref) => (
-    <RowContent {...props} ref={ref} dataAttributes={{'component-name': 'Row', ...dataAttributes}} />
-));
+export const Row = React.forwardRef<TouchableElement, RowContentProps>(
+    ({dataAttributes, role = 'listitem', ...props}, ref) => (
```
atabel, line 656: note, it's not needed to do the same for `BoxedRow`, because `BoxedRowList` is using a `Stack` with the list role, which automatically sets the listitem role to their children

aweell: On of the drawbacks of usign the list role is that is not possible to make a difference between ordered or unordered lists, have you considered using as="ul | ol" in the rowList component instead?

atabel: hmmm... do you think that is a needed use case? do we need `<ol>` here?
Also, using `<ol>` removing the default styles has it's own problems too:
image

aweell: I was thinking in an invoices list as a use case for `<ol>`, but definitively based on Safari's behaviour the role="list" seems the right choice here yes. Thanks!

marcoskolodny (approved):

pladaria (approved):

src/list.tsx
```diff
@@ -730,7 +739,7 @@ type BoxedRowListProps = {
 export const BoxedRowList: React.FC<BoxedRowListProps> = ({
     children,
     ariaLabelledby,
-    role,
+    role = 'list',
     dataAttributes,
 }) => (
     <Stack
```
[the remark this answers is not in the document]
atabel: That would have other implications. I'd need to render Stack children items as li and reset their styles, and we'll probably end with the same problem discussed here:
https://github.com/Telefonica/mistica-web/pull/1068#discussion_r1551523359

yceballost (approved):

aweell (approved):

PR #1070, opened 2024-04-04
feat(Touchable, Buttons): Improve touchable and buttons accessibility

pladaria:
https://jira.tid.es/browse/WEB-1794

This PR adds the `onNavigate` and `role` props to:
- `Touchable`
- `ButtonPrimary`, `ButtonSecondary`, etc
- `ButtonLink`
- `TextLink`

There are other touchable-like components like `IconButton`, `Row`, `Card`, `Chip`... We can add these features when required but I think the scope of this PR is Ok for now


src/utils/browser.tsx
```diff
@@ -0,0 +1,9 @@
+export const redirect = (url: string, external = false, loadOnTop = false): void => {
```
pladaria, line 1: moved to an external file to facilitate testing

Winde (approved):

src/__tests__/button-test.tsx
```diff
@@ -1,12 +1,32 @@
...
+import type {ThemeConfig} from '../theme';
+
+jest.mock('../utils/browser', () => ({
+    ...jest.requireActual('../utils/browser'),
+    redirect: jest.fn(),
+}));
+
+beforeEach(() => {
+    (redirectSpy as any).mockReset();
+});
```
marcoskolodny, line 21: Out of curiosity: why is this needed?

src/button.tsx
```diff
@@ -377,7 +384,7 @@ const Button = React.forwardRef<TouchableElement, ButtonProps & {type: ButtonTyp
             EndIcon: props.EndIcon,
         }),
         disabled: props.disabled || showSpinner || isFormSending,
-        role: 'button',
+        role: props.role,
```
marcoskolodny, line 387: Shouldn't we default to button in case role was not provided?

src/touchable.tsx
```diff
@@ -52,56 +44,80 @@ interface CommonProps {
...
-export interface PropsMaybeTo extends CommonProps {
+
+interface PropsMaybeTo extends CommonProps {
     maybe: true;
     to?: string | Location;
     fullPageOnWebView?: boolean;
     replace?: boolean;
     href?: undefined;
     onPress?: undefined;
+    /** with "to", onNavigate will be executed in parallel to the navigation */
```
marcoskolodny, line 88: As far as I can see, it's not in parallel, right? You wait for the promise to be resolved, and after that you navigate

playroom/components/index.tsx
```diff
@@ -30,7 +30,7 @@ import IconCode from '../icons/icon-code';
 import * as styles from '../preview-tools.css';
 import {CSSTransition} from 'react-transition-group';
 
-import type {IconProps} from '../../dist';
+import type {IconProps} from '../../src/utils/types';
 import type {ThemeConfig, ColorScheme, KnownSkinName} from '../../src';
```
atabel, line 34: why don't you import `IconProps` form `'../../src'` too?

src/__tests__/button-test.tsx
```diff
@@ -44,18 +64,13 @@ test('<a> is rendered when using "to" prop', () => {
         </ThemeContextProvider>
     );
 
-    const anchor = screen.getByRole('button', {name: 'test'});
+    const anchor = screen.getByRole('link', {name: 'test'});
```
atabel, line 67: I know this is an expected change, but it will probably break many tests in webapp

src/touchable.tsx
atabel, line 1: Maybe we could have a `TouchableProps<T>` generic type to reuse all the `onPress`/`to`/`href` + friend props in any touchable component (Buttons, IconButton, Cards, Rows, etc).

I've created a ticket: https://jira.tid.es/browse/WEB-1817

atabel (approved):
Please, create a ticket to add `onNavigate` prop to cards, rows, etc

src/button.tsx
pladaria: the default is provided by the dom element (button for `<button>` and link for `<a>`)

src/touchable.tsx
pladaria: what you comment is true in the "href", not for the "to".

The "to" prop will use a `<Link>` element which produces a client side navigation, so it is ok to perform the onNavigate actions in parallel (this was already happening in current implementation)

playroom/components/index.tsx
pladaria: ok, will do

pladaria:
> Please, create a ticket to add `onNavigate` prop to cards, rows, etc

https://jira.tid.es/browse/WEB-1818

PR #1079, opened 2024-04-11
feat(Row): improve VoiceOver compatibility

atabel:
Fixes:
WEB-1724 
WEB-1740

The main issue that was causing the bugs is that Safari reads links with multiple lines like different links. The workaround for this is:

- Add an aria-label to the `<a>` element
- Add an aria-hidden to the `<a>` children
- The aria-label of the `<a>` should be computed with the content of the row, in this case we are concatenating `[title, headline, subtitle, description, extra]` props to create the aria-label. The problem is that some of these props aren't simple `string`s but any `ReactNode`, so we need to extract the inner text of the html elements on runtime to compute the `aria-label` for the `<a>`

Other changes in the PR:
- Fix some cases where the aria-label was not being set properly in rows with controls (checkbox/radio/switch)
- Change the dom order of `headline` to make screen readers read it after `title`

Still pending:
- Review badge a11y
- Review how should screen reader read `details` part. When the row is a link/button it is read as part of the row as expected. But when we use some control (switch/checkbox/radio) the details part is not read because it's outside the div referenced by `aria-labeledby`. Including `details` in the same div used for `headline`, `title`, `description`, etc would require a big refactor of this component layout


src/list.tsx
```diff
@@ -350,8 +351,12 @@ const RowContent = React.forwardRef<TouchableElement, RowContentProps>((props, r
         role,
         extra,
         dataAttributes,
+        'aria-label': ariaLabel,
     } = props;
 
+    // iOS voiceover reads links with multiple lines as separate links. By setting aria-label and marking content as aria-hidden, we can make it read the whole row as one link.
+    const iosVoiceOverLabelForLink = [title, subtitle, description, detail].filter(Boolean).join(' ');
```
atabel, line 358: caveat: headline is not read because it isn't included here. The problem is that `headline` prop is a `React.Node`, so I can't extract the text

src/__tests__/sheet-test.tsx
```diff
@@ -45,7 +45,7 @@ test('Sheet', async () => {
 
     await userEvent.click(closeButton);
     await waitForElementToBeRemoved(sheet);
-}, 20000);
+}, 30000);
```
marcoskolodny, line 48: Let's hope it doesn't fail again :(

src/list.tsx
```diff
@@ -468,8 +477,9 @@ const RowContent = React.forwardRef<TouchableElement, RowContentProps>((props, r
                 role={role}
                 dataAttributes={dataAttributes}
                 disabled={disabled}
+                aria-label={ariaLabel ?? iosVoiceOverLabelForLink}
```
marcoskolodny, line 480: What about Android if you don't use ariaLabel prop? Does it have the same issue as iOS? If not, why not letting the Android browser handle it natively instead of setting iosVoiceOverLabelForLink?

src/list.tsx
marcoskolodny: Have you tried something like this? https://stackoverflow.com/questions/34204975/react-is-there-something-similar-to-node-textcontent

I know it may be a bit "hacky", but usually the headline prop is used for rendering very simple elements, so I would give it a try

pladaria (approved):

src/list.tsx
atabel: do you think it's worth doing platform detection for this?

marcoskolodny: I'm not sure it will affect that much if we leave it for every platform.

Small thing: if we decide not to do it, I wouldn't name this variable "ios..."

marcoskolodny (approved):

pladaria (approved):

PR #1084, opened 2024-04-16
feat(vivo-type): make screen readers read vivinho char as Vivo

atabel:
WEB-1826


marcoskolodny (approved):
As discussed offline, these changes may break unit/screenshot/acceptance tests. It would be nice to see how many of them are affected

pladaria (approved):

PR #1088, opened 2024-04-18
fix(Sheet): fix sheets in large desktop sizes

pladaria:


src/sheet.tsx
```diff
@@ -28,6 +28,7 @@ import {safeAreaInsetBottom} from './utils/css';
 
 import type {ExclusifyUnion} from './utils/utility-types';
```
pladaria, line 29: better review ignoring whitespace: https://github.com/Telefonica/mistica-web/pull/1088/files?diff=unified&w=1

src/test-utils/index.tsx
```diff
@@ -124,8 +124,8 @@ const DEVICES: DeviceCollection = {
     [LARGE_DESKTOP_DEVICE]: {
```
pladaria, line 124: the issue was reproduced above ~1600px width

atabel (approved):

marcoskolodny (approved):

yceballost (approved):

PR #1097, opened 2024-04-25
feat(Cards,IconButton,TextLink,Chip): add onNavigate prop

atabel:
WEB-1818


src/list.tsx
```diff
@@ -269,6 +271,7 @@ interface ToRowContentProps extends CommonProps {
     to: string | undefined;
     fullPageOnWebView?: boolean;
     replace?: boolean;
+    onNavigate?: () => void | Promise<void>;
```
marcoskolodny, line 274: Why not generalizing using the new Touchable type that you created?

src/icon-button.tsx
```diff
@@ -97,60 +97,16 @@ const getButtonStyle = (
  *
  */
 export const RawDeprecatedIconButton = React.forwardRef<TouchableElement, DeprecatedProps>((props, ref) => {
-    const {icon, children} = props;
```
marcoskolodny, line 100: Nice cleanup :100:

marcoskolodny (approved):

pladaria (approved):

src/list.tsx
atabel: This case is more complex because there are more cases (controls, iconButton, etc), although it may be possible to use the type util here too. Also, I have two other PRs changing this same component, and one of them changing types too. I don't want conflicts.

PR #1100, opened 2024-04-26
feat(ResponsiveLayout, ResetResponsiveLayout, Hero, Slideshow): refactor ResponsiveLayout and update components

marcoskolodny:
Issues: [WEB-1833](https://jira.tid.es/browse/WEB-1833) and [WEB-1829](https://jira.tid.es/browse/WEB-1829)

- [X] Refactor ResponsiveLayout in order to not use selectors and simplify logic
- [X] Create ResetResponsiveLayout
- [X] Remove layout hacks related to this in Hero
- [X] Remove layout hacks related to this in Slideshow
- [X] Verify with design Carousel/Slideshow behaviour
- [X] Verify that webapp tests pass correctly
- [x] Create tests to cover many scenarios related to ResponsiveLayout and ResetResponsiveLayout

**Note:** I've tested these changes in webapp through the test_mistica_branch action, and there were two failing tests ([result](https://github.com/Telefonica/webapp/actions/runs/8878728973)). The second one has nothing to do with this PR and it's intended. The first one is actually the result of fixing a bug we had in the previous implementation of ResponsiveLayout.

If you see the diff, you will notice that the content in the left screenshot has an empty space (margin) on the right that shouldn't be there:

image

The problem was that using ResponsiveLayout with fullWidth was reseting the sideMargin variable to 0px in the previous logic. This breaks the case in which you have two nested layouts, with the second one having fullWidth while the first one not (because if children content uses the value of the margin, like in carousel, it will read 0px instead of the real value). This can be reproduced [in playroom](https://mistica-web.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8AlGBnADhAduglgG4wAyAhgJ4QCuALgHwA6OABC6htnkaZTbSwBm1ADYiA6vii0AFk1Zt2AYTIAnGuhgjmixQHd8sgEKiRMWuh26WhmAFt0AXmABBVasoA6QersAKYBYzHABzWXgWADYWAF8AGhY-AH0EqQAPAEoWR3pEq2t2AFkYKHwyFVUofIKWGRgyKBF8HBhnJAAVMhCWWgpMVsYQTF8IQfoACXrG5pgkAHpOkPoY6oLaQzNnQYqoFkGWAGobKDSVhRqWWHQwVXxMddxHQYARDBu7h5xB1es7ErKtucLuwAJJ2LowH4XdCqMBPEAyWi0TDoeBzOboGiwmCeah4TAiMjoGSeSB2OYARkiAAZqWkAJy0uYAfj0ZFoYBk3yBwJYRP6YFoKHZ%2BAg8Kp8Hp3N5bDm8l5Z15ACM6LRHsAodYkCYkbgAAq3cGqCgsdDgsQsfWqDBOYB%2BLI5FjAGIxeUytguQWir48mrzHVqnAG-BGihui6K4Eq3U4EjNADWbQDuDjOHjlqD1vQtvt2Vyztdqfj-tVKYTy01cp%2BGQykdlbvmaCwuAIxHIVDo8kbnBbPHb-HkIDiIFk9gwCAA2iBChBCPh0LQ1CAALrDgzSGToScAZgATNTlzEgA).


src/__screenshot_tests__/__image_snapshots__/private-hero-nested-responsive-layouts-screenshot-test-tsx-hero-with-nested-responsive-layouts-mobile-ios-1-snap.png
marcoskolodny, line 1: This was actually incorrect. According to [specs](https://www.figma.com/file/M8t65SWdTUyM44YjxuzCyu/%F0%9F%94%B8-Hero-Specs?type=design&node-id=0-8183&mode=design&t=TUvg97zsmmeHAzlo-0), the content should have padding relative to the image

src/__screenshot_tests__/__image_snapshots__/private-hero-nested-responsive-layouts-screenshot-test-tsx-hero-with-nested-responsive-layouts-desktop-1-snap.png
marcoskolodny, line 1: This was caused by making the story to be rendered with fullScreen: true (it was accidentally not added before)

src/responsive-layout.tsx
```diff
@@ -73,4 +77,29 @@ const ResponsiveLayout: React.FC<Props> = ({children, ...props}) => (
     </InternalResponsiveLayout>
 );
 
+export const ResetResponsiveLayout: React.FC<{
```
marcoskolodny, line 80: For now, we keep this new component for internal use. We can export it outside Mistica if we see that there is a real need in the future

pladaria (commented):
nice!

src/responsive-layout.tsx
```diff
@@ -73,4 +77,29 @@ const ResponsiveLayout: React.FC<Props> = ({children, ...props}) => (
...
+    children: React.ReactNode;
+    mobile?: boolean;
+    desktop?: boolean;
```
pladaria, line 83: it is strange to have boolean props with `true` as default value. What about having `skipMobile` and `skipDesktop` props?

so, instead of `desktop={false}`, just `skipDesktop`

src/responsive-layout.tsx
```diff
@@ -73,4 +77,29 @@ const ResponsiveLayout: React.FC<Props> = ({children, ...props}) => (
...
+    className?: string;
+    innerDivClassName?: string;
```
atabel, line 85: are these props used? if not, I'd remove them

src/__screenshot_tests__/private-responsive-layout-scenarios-screenshot-test.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+test.each(DEVICES)('ResponsiveLayout scenarios (%s)', async (device) => {
+    const page = await openStoryPage({
+        id: 'private-responsivelayout-scenarios--default',
+        device,
+    });
+
+    const image = await page.screenshot({fullPage: true});
+    expect(image).toMatchImageSnapshot();
+});
```
atabel, line 13: I think I'd create a different screenshot for each scenario. Otherwise it's difficult to know what are we testing.
Also, I'd include some text in the screenshot describing the case under test

src/__private_stories__/private-responsive-layout-scenarios-story.tsx
```diff
@@ -0,0 +1,230 @@
...
+    return (
+        <Box paddingY={24}>
+            <Stack space={24}>
+                {/** Carousel with alternating fullWidth and not fullWidth layouts */}
+                <ResponsiveLayout>
+                    <ResponsiveLayout fullWidth>
+                        <ResponsiveLayout>
+                            <ResponsiveLayout fullWidth>
+                                <ResponsiveLayout>
+                                    <ResponsiveLayout fullWidth>
```
atabel, line 33: what are we testing here? do we need to repeat the nesting so many times?
it's the carousel relevant to the test or we could use a simpler component (Placeholder)?

src/__private_stories__/private-responsive-layout-scenarios-story.tsx
```diff
@@ -0,0 +1,230 @@
...
+                                            ))}
+                                        />
+                                    </ResponsiveLayout>
+                                </ResponsiveLayout>
+                            </ResponsiveLayout>
+                        </ResponsiveLayout>
+                    </ResponsiveLayout>
+                </ResponsiveLayout>
+
+                {/** Hero wrapped in multiple layouts */}
```
atabel, line 58: I'd include some text in the screenshot describing the test case. Something like "multiple nested ResposiveLayouts are ignored"

src/__private_stories__/private-responsive-layout-scenarios-story.tsx
```diff
@@ -0,0 +1,230 @@
...
+                                    </ButtonPrimary>
+                                }
+                                buttonLink={<ButtonLink onPress={() => {}}>Link</ButtonLink>}
+                            />
+                        ))}
+                    />
+                </ResponsiveLayout>
+
+                {/** Multiple layouts */}
+                <ResponsiveLayout>
```
atabel, line 107: Isn't this the same case you are testing with the Hero?

src/responsive-layout.tsx
marcoskolodny: Removed

src/responsive-layout.tsx
marcoskolodny: makes sense, updated

src/__private_stories__/private-responsive-layout-scenarios-story.tsx
marcoskolodny: The carousel is relevant because it uses the margin var for some calculations. I could remove some of the nested layouts, but I also don't see any harm in leaving them as they are

src/__private_stories__/private-responsive-layout-scenarios-story.tsx
marcoskolodny: Similar. In mobile, Hero uses a reset as wrapper, so it's not exactly the same

src/__private_stories__/private-responsive-layout-scenarios-story.tsx
marcoskolodny: done

src/__screenshot_tests__/private-responsive-layout-scenarios-screenshot-test.tsx
marcoskolodny: Done

atabel (approved):

yceballost (approved):

pladaria (approved):

PR #1102, opened 2024-04-29
fix(AdvancedDataCard): footer extra space

nicholasbrutti-tlf:
Fix advanced data card extra space at bottom

Original:
original

Fixed:
fixed


src/community/advanced-data-card.tsx
```diff
@@ -315,7 +315,10 @@ export const AdvancedDataCard = React.forwardRef<HTMLDivElement, AdvancedDataCar
                                 !hasFooter && !hasExtras ? styles.minHeight : ''
                             )}
                         >
-                            <Box paddingTop={8}>
+                            <Box
+                                paddingTop={8}
+                                className={!hasFooter && !hasExtras ? '' : styles.minHeightBox}
```
atabel, line 320: why don't you add this class to the parent div instead?

nicholasbrutti-tlf: because this would affect the spacing of the box, making it smaller than it should be to complete the minimum height of the component, increasing the height of the footer

Captura de tela 2024-04-29 165810

atabel (approved):

PR #1104, opened 2024-04-30
feat(Theme): expose `colorValues` in theme

pladaria:
https://jira.tid.es/browse/WEB-1781

This way we can access color information without using `getCssVarValue`. This is useful when we need to use the color outside a CSS context, for example to send it to the bridge, use it as an API param...

### DIFF
--- a/src/theme-context-provider.tsx
+++ b/src/theme-context-provider.tsx
@@ -90,13 +90,12 @@ const ThemeContextProvider: React.FC<Props> = ({theme, children, as}) => {
     const isDarkModeEnabled = (colorScheme === 'auto' && isOsDarkModeEnabled) || colorScheme === 'dark';
     const colors: Colors = isDarkModeEnabled ? darkColors : lightColors;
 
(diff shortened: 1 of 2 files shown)


src/theme-context-provider.tsx
```diff
@@ -90,13 +90,12 @@ const ThemeContextProvider: React.FC<Props> = ({theme, children, as}) => {
     const isDarkModeEnabled = (colorScheme === 'auto' && isOsDarkModeEnabled) || colorScheme === 'dark';
     const colors: Colors = isDarkModeEnabled ? darkColors : lightColors;
 
-    const contextTheme = React.useMemo<Theme>(() => {
+    const contextTheme = React.useMemo((): Theme => {
```
pladaria, line 93: This way the function return type is stricter. Otherwise it accepts an object with unexpected properties.

marcoskolodny (approved):

atabel (approved):

PR #1105, opened 2024-04-30
feat(o2-new): dark mode improvements

yceballost:
This PR contains: 
1. Update Tokens from Mistica-design repo 
	- Include new two tokens textBrand & textActivated
	- Update o2-new dark mode colors
2. New tokens replace `brand` and `controlActivated` tokens usage in some components to ensure color contrast accessibility for future brand improvements


src/skins/o2-new.tsx
```diff
@@ -36,7 +36,7 @@ export const palette = {
     o2Red40: '#FF8B8B',
```
pladaria, line 36: Some elements have little contrast (the "3" in the stepper or the "check" in the checkbox):
image
image

pladaria (approved):

atabel (approved):

src/skins/o2-new.tsx
yceballost: Yea, we know. Probably we need to change these elements to `background` token to ensure that. But with this PR I would like to avoid design changes

PR #1106, opened 2024-04-30
fix(Hero): add bottom padding inside slideshow only when it has bullets

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1850)


src/__private_stories__/private-hero-nested-responsive-layouts-story.tsx
```diff
@@ -12,10 +12,10 @@ export const Default: StoryComponent = () => {
             <ResponsiveLayout>
                 <Slideshow
                     dataAttributes={{testid: 'slideshow'}}
-                    withBullets
                     inverseBullets={false}
                     items={Array.from({length: 3}, (_, idx) => (
                         <Hero
+                            noPaddingY
```
marcoskolodny, line 18: Updated this private story so it also tests this scenario

src/hero.tsx
```diff
@@ -138,7 +138,9 @@ const Hero = React.forwardRef<HTMLDivElement, HeroProps>(
         ref
     ) => {
         const {isTabletOrSmaller} = useScreenSize();
-        const isInsideSlideShow = useIsInsideSlideshowContext();
+        const slideshowContextValue = useSlideshowContext();
+        const isInsideSlideShow = !!slideshowContextValue;
```
marcoskolodny, line 142: We won't need this variable `isInsideSlideshow` after we merge https://github.com/Telefonica/mistica-web/pull/1100

src/hero.tsx
```diff
@@ -225,7 +227,8 @@ const Hero = React.forwardRef<HTMLDivElement, HeroProps>(
                             template="6+6"
                             left={
                                 <Box
-                                    paddingY={noPaddingY ? 0 : 56}
+                                    paddingTop={noPaddingY ? 0 : 56}
+                                    paddingBottom={hasSlideshowBullets ? 48 : noPaddingY ? 0 : 56}
```
aweell, line 231: In desktop the padding bottom has always 56px with bullets and without them when noPaddingY is false

aweell (commented):
https://tinyurl.com/259bhpzj

A slideshow of heros with different content height is causing the slideshow bullets to become invisible (Tried to fix that setting a fixed height, but it seems that is not working?)

src/hero.tsx
marcoskolodny: Fixed

marcoskolodny:
> https://tinyurl.com/259bhpzj
> 
> A slideshow of heros with different content height is causing the slideshow bullets to become invisible (Tried to fix that setting a fixed height, but it seems that is not working?)

This happened because Hero was not accepting numbers for the height prop (and Playroom was not complaining about it). I've changed the logic so that using numbers is also fine

src/hero.tsx
```diff
@@ -147,41 +149,41 @@ const Hero = React.forwardRef<HTMLDivElement, HeroProps>(
 
         if (isTabletOrSmaller) {
             return (
-                <div style={applyCssVars({[mediaStyles.vars.mediaBorderRadius]: '0px'})}>
```
marcoskolodny, line 150: See this changes hiding whitespaces (https://github.com/Telefonica/mistica-web/pull/1106/files?diff=split&w=1).

Wrapping the content with this div was causing `height: "100%"` to not work. I've moved this css var a bit down to avoid the unnecessary extra div and now it works as expected

pladaria (approved):

src/carousel.tsx
```diff
@@ -680,13 +680,18 @@ type SlideshowProps = {
...
+const SlideshowContext = React.createContext<{withBullets: boolean} | undefined>(undefined);
 
-export const useIsInsideSlideshowContext = (): boolean => React.useContext(IsInsideSlideshowContext);
+export const useSlideshowContext = (): {withBullets: boolean} | undefined =>
+    React.useContext(SlideshowContext);
 
-export const IsInsideSlideshowProvider = ({children}: {children: React.ReactNode}): JSX.Element => (
-    <IsInsideSlideshowContext.Provider value>{children}</IsInsideSlideshowContext.Provider>
-);
+export const SlideshowProvider = ({
```
atabel, line 688: why is this exported?

also, I think you don't need this component, just use `SlideshowContext.Provider`

atabel (approved):

src/carousel.tsx
marcoskolodny: Not sure why it was being exported. I've removed it

aweell (approved):

PR #1107, opened 2024-04-30
fix(Video): avoid duplicated state update in Video's state machine

marcoskolodny:
Issue: [link](https://jira.tid.es/browse/WEB-1846)

### DIFF
--- a/src/video.tsx
+++ b/src/video.tsx
@@ -188,7 +188,6 @@ const Video = React.forwardRef<VideoElement, VideoProps>(
             dispatch('finishLoad');
             if (video && shouldAutoPlay && video.paused) {
                 video.play();
-                dispatch('play');
             }
         };
 


src/video.tsx
```diff
@@ -188,7 +188,6 @@ const Video = React.forwardRef<VideoElement, VideoProps>(
             dispatch('finishLoad');
             if (video && shouldAutoPlay && video.paused) {
                 video.play();
-                dispatch('play');
```
marcoskolodny, line 191: This is already done [here](https://github.com/Telefonica/mistica-web/blob/master/src/video.tsx#L223). If we do it twice, it will cause the if statement in `onTimeUpdate` callback to evaluate to false (and therefore the `onPlay` callback passed as prop won't be executed).

pladaria (approved):

yceballost (approved):

atabel (approved):

PR #1108, opened 2024-04-30
fix(Chip): expand chips when inside a grid

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1849)


src/chip.tsx
```diff
@@ -139,6 +141,8 @@ const Chip: React.FC<ChipProps> = (props: ChipProps) => {
         <Box
             className={classnames(
                 styles.chipVariants[active ? 'active' : overAlternative ? 'overAlternative' : 'default'],
+                // If the chip is wrapped inside a BaseTouchable, we set inline-flex to the Touchable instead
```
marcoskolodny, line 144: The issue was that when the chip is wrapped inside a touchable, we were not setting inline-flex to the wrapper

pladaria (approved):

Winde (approved):

atabel (approved):

PR #1113, opened 2024-05-06
feat(Accordion): improve accessibility

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1816)

I've tried to use details/summary HTML markup, but there are some things that were not great about it:

- Behavior in browsers differs (specially Safari), making it trickier to implement.
- Screen reader doesn't consider <summary/> as interactive. When you press on the Accordion's header, it jumps to elements inside it directly.
- AFAIK, there are no simple ways to animate the opening/closing of the accordion's content. I've found only hacks that require handling JS events and other tricky things.

I believe that the current implementation is way more clean. In terms of accessibility, I've compared summary/details with it, and our implementation reads everything that the summary one reads.

Changes on this PR:

- [X] Disable animations if the user's device has reduced motion accessibility setting turned on
- [X] Allow passing role to Accordion/AccordionItem
- [X] Fix logic for rendering last divider in a list of Accordion items
- [X] Code cleanup


src/accordion.tsx
```diff
@@ -24,13 +25,15 @@ const ACCORDION_TRANSITION_DURATION_IN_MS = 400;
 
 type AccordionContextType = {
     index: ReadonlyArray<number>;
-    toogle: (item: number) => void;
+    toggle: (item: number) => void;
```
marcoskolodny, line 28: typo: toogle -> toggle

src/accordion.tsx
```diff
@@ -39,9 +42,11 @@ interface AccordionItemContentProps {
     subtitle?: string;
     asset?: React.ReactNode;
     content: React.ReactNode;
-    onToogle?: (value: boolean) => void;
     dataAttributes?: DataAttributes;
     trackingEvent?: TrackingEvent | ReadonlyArray<TrackingEvent>;
+    role?: string;
+    /** @deprecated Use onChange Accordion's onChange callback instead */
+    onToogle?: (value: boolean) => void;
```
marcoskolodny, line 49: This callback had no effect and it's not used anywhere. Given that there are no usages of accordion in webapp, I propose deleting it to avoid confusions. For now, I deprecated it. WDYT?

src/accordion.tsx
```diff
@@ -194,19 +199,22 @@ const AccordionItemContent = React.forwardRef<TouchableElement, AccordionItemCon
 );
 
 export const AccordionItem = React.forwardRef<TouchableElement, AccordionItemContentProps>(
-    ({dataAttributes, ...props}, ref) => (
-        <AccordionItemContent
-            {...props}
-            ref={ref}
-            dataAttributes={{'component-name': 'AccordionItem', ...dataAttributes}}
-        />
+    ({dataAttributes, role = 'listitem', ...props}, ref) => (
```
marcoskolodny, line 202: Adding list/listitem roles by default (similar to Rows)

src/accordion.tsx
```diff
@@ -263,14 +275,13 @@ interface BoxedAccordionItemProps extends AccordionItemContentProps {
     isInverse?: boolean;
 }
 
-export const BoxedAccordionItem = React.forwardRef<HTMLDivElement, BoxedAccordionItemProps>(
+export const BoxedAccordionItem = React.forwardRef<TouchableElement, BoxedAccordionItemProps>(
     ({dataAttributes, isInverse, ...props}, ref) => (
         <Boxed
             isInverse={isInverse}
-            ref={ref}
```
marcoskolodny, line 270: This was wrong. The ref should be passed to the content. Otherwise, calling ref.current.click() won't affect the Accordion

src/accordion.tsx
aweell: This solves the relationship between the accordion group and each of the accordions in case they have relation between them, since they could also be only regions of content with no relation between them.

How the relationship between the header and the content can be defined?

I see two cases:

* They are a group of controls: The accordion item acts as a `fieldset` where the header title is the legend and the controls are inside the collapsable panel.
* They are a definition list (usually used for FAQs): The list of accordions acs as a definition list so the accordion header should have the role: "term" and the content of the collapsable section, the role: "definition"

marcoskolodny: I'm not sure I understood you. Currently, the relation between the header and the content is made by using aria-controls in the header and aria-labelledby + role=region in the panel.

Is that what you meant?

src/accordion.tsx
```diff
@@ -194,19 +199,22 @@ const AccordionItemContent = React.forwardRef<TouchableElement, AccordionItemCon
...
+            />
+        </div>
     )
 );
 
 type AccordionBaseProps = {
     children: React.ReactNode;
     dataAttributes?: DataAttributes;
     onChange?: (index: number, value: boolean) => void;
+    role?: string;
```
marcoskolodny, line 217: Initially, I added "list" and "listitem" as default roles for Accordion and AccordionItem, but after discussing with @aweell, he pointed out that it may not be a good idea.

For example, if the accordions are being used to divide some content of a document in different sections, it's not the best idea to use a list for them.

We've decided to make it customizable, but without setting any default values

src/accordion.tsx
marcoskolodny: Discussed offline. Alex will investigate about best practices regarding this and will come back to us with some updates. In case some changes are necessary, we can iterate on it in a different PR.

src/accordion.tsx
aweell: Specs are updated accordingly

aweell (approved):

src/accordion.tsx
atabel: please, don't remove it, just deprecate it. It would be a breaking change. Even if it isn't used in Novum webapp it could be used by other clients

src/accordion.tsx
```diff
@@ -194,19 +199,22 @@ const AccordionItemContent = React.forwardRef<TouchableElement, AccordionItemCon
...
 
 export const AccordionItem = React.forwardRef<TouchableElement, AccordionItemContentProps>(
-    ({dataAttributes, ...props}, ref) => (
-        <AccordionItemContent
-            {...props}
-            ref={ref}
-            dataAttributes={{'component-name': 'AccordionItem', ...dataAttributes}}
-        />
+    ({dataAttributes, role, ...props}, ref) => (
+        <div role={role} className={sprinkles({width: '100%'})}>
```
atabel, line 203: instead of adding a new div just for the role, why don't you forward the `role` prop to `AccordionItemContent`?

src/accordion.tsx
atabel: Is this a real use case? I mean, it feels weird to me to control the open/close of the Accordion using a ref.

atabel (commented):
Please, comment in the PR description a summary of the a11y improvements made here. Something like "made role customizable and removed the open/close transition when the user has the reduced motion setting enabled"

src/accordion.tsx
marcoskolodny: I'm not sure. But even if there is no real use case, the ref logic was already implemented, so removing it would be a breaking change. I've just fixed it so that it actually works as expected

src/accordion.tsx
marcoskolodny: Discussed offline. In non-boxed rows, we need to set the listitem to the wrapper. Otherwise, we would have a11y problems

src/accordion.tsx
marcoskolodny: Discussed offline. We will put the ref in the topmost element of the component. We believe it should be like this because refs are usually used for other things instead of clicking the element (like checking component's visibility)

atabel (approved):

PR #1118, opened 2024-05-08
fix(AdvancedDataCard): minHeight issues caused by #1102

atabel:
fix [issue caused in Novum webapp](https://github.com/Telefonica/webapp/pull/4549/files/74159dfb3316a524b7ea9f971b0e37f99d83e9d7#r1592260448) because of changes in #1102
image


src/community/advanced-data-card.css.ts
```diff
@@ -25,7 +25,6 @@ export const actions = style([
     paddingX,
     sprinkles({
         display: 'flex',
-        flex: 1,
```
atabel, line 28: this was causing the footer to grow to fit parent minHeight

src/community/advanced-data-card.css.ts
```diff
@@ -49,6 +48,7 @@ export const touchableArea = style({
 
 export const touchableContainer = style({
     display: 'flex',
+    flex: 1,
```
atabel, line 51: this is the part that should grow instead

pladaria (approved):

PR #1127, opened 2024-05-22
fix(IconButton): add borderRadius to button's overlay

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1879)

### DIFF
--- a/src/icon-button.css.ts
+++ b/src/icon-button.css.ts
@@ -417,6 +417,7 @@ export const overlay = style({
     width: '100%',
     height: '100%',
     backgroundColor: 'transparent',
+    borderRadius: '50%',
     transition: `background-color ${colorTransitionTiming}`,
 
     selectors: {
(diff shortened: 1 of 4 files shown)


src/icon-button.css.ts
```diff
@@ -417,6 +417,7 @@ export const overlay = style({
     width: '100%',
     height: '100%',
     backgroundColor: 'transparent',
+    borderRadius: '50%',
```
marcoskolodny, line 420: I have no idea why the overlay was rounded in most cases, but not in Popover (only in iOS).

This fixes the issue

src/tooltip.tsx
```diff
@@ -569,7 +569,7 @@ export const BaseTooltip: React.FC<BaseTooltipProps> = ({
 
                                                 {onClose && (
                                                     <div className={styles.closeButtonIcon}>
-                                                        <InternalIconButton
+                                                        <IconButton
```
marcoskolodny, line 572: Not related. We can use IconButton here, we are not using hasOverlay/isOverMedia in here

yceballost (approved):
🫶

atabel (approved):

PR #1128, opened 2024-05-22
Storybook a11y POC

marcoskolodny:


src/icons/__stories__/mistica-icons-story.tsx
```diff
@@ -27,7 +27,7 @@ export default {
             control: {type: 'range', min: 24, max: 48, step: 4},
         },
     },
-    parameters: {fullScreen: true},
+    parameters: {fullScreen: true, a11y: {disable: true}},
```
marcoskolodny, line 30: Removing the a11y addon in the icons story because it's painfully slow and makes the browser freeze.

.storybook/preview.tsx
```diff
@@ -81,6 +81,16 @@ const getTheme = (
     };
 };
 
+const findAccessibilityAddonButton = (): HTMLButtonElement | undefined => {
```
marcoskolodny, line 84: Search for the button to run the a11y tests in the storybook UI and click it.

test

.storybook/main.js
yceballost, line 1: is it possible to reorder sidebar tabs to keep "code" tab first? I have doubt if the first tab should be code or controls, but being a developer oriented site, I think "code" tab should be the first one.

wdyt?

although maybe we will lost the "automatic check" from accessibility panel

marcoskolodny: If we reorder tabs, we lose the automatic a11y check :(

marcoskolodny: I've changed the logic in order for this to work. It should be okay now

yceballost (approved):
love it 🥰

atabel (commented):
I'd like to discuss the function we need to cover with this addon.

- We already run automated a11y tests for all the stories using Axe. Isn't this addon redundant?
- The failing rules in this addon won't make CI fail. I've seen we have some reported issues, are we supposed to fix all them? how do we make sure all the rules pass if CI doesn't fail?
- What does this addon uses under the hood? it also uses Axe or a different tool?
- Can we silence rules we don't want to fix (color contrast, for example)?
- The highlight results feature doesn't work for me: 
image

yceballost:
> I'd like to discuss the function we need to cover with this addon.
> 
> * We already run automated a11y tests for all the stories using Axe. Isn't this addon redundant?

IMO, this addon helps to detect misuse in some brands for developer and designers from other teams.

For example, an inverse header in Movistar should not be used because the background colour does not contrast with inverse text, however, in Vivo this use is valid.

pladaria (approved):
As commented in Teams. I would remove this option from the production build and keep it for development or PR previews.

PR #1129, opened 2024-05-24
feat(Table): new component

atabel:
jira: WEB-1870


src/__acceptance_tests__/__ssr_pages__/main-navigation-bar-large.tsx
```diff
@@ -23,7 +23,7 @@ const NavigationBarTest = (): JSX.Element => (
                     </Badge>
                 </NavigationBarAction>
                 <NavigationBarAction onPress={() => {}} aria-label="Open profile">
-                    <Avatar size={24} initials="ML" src="https://source.unsplash.com/600x600/?face" />
```
atabel, line 26: unrelated change, fixed this test because it was using an external image and it was failing to load

yceballost (changes requested):

pladaria (commented):
nice!
As commented in Teams, in mobile, when fullwidth, it would be great to have a way to bleed outside the responsive layout and set the padding/margin to the table. This way it would scroll over the whole viewport width

src/table.tsx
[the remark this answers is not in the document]
atabel: I think I've included a code comment explaining why this is impossible. The non-boxed table doesn't have a background, so you can't have a fixed header, because the header would be transparent and you'd see the rows going under it

src/table.tsx
```diff
@@ -0,0 +1,184 @@
...
+import {Text1, Text2} from './text';
+import {InternalBoxed} from './boxed';
+import classNames from 'classnames';
+import Box from './box';
+import {applyCssVars} from './utils/css';
+
+import type {DataAttributes} from './utils/types';
+
+type TableProps = {
+    heading: Array<string>;
```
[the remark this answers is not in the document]
atabel: not according the spec, and I think it could be problematic, how would you show them in mobile (collapse-rows mode)?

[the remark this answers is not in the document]
yceballost: yep, I remember talking to Alex about this but I don't remember what the outcome was. I thought that the custom item would be included and that in mobile it is the product's responsibility... that's why I wanted Alex's confirmation.

The product could use strings in mobile and custom composition in desktop, for example.

src/table.tsx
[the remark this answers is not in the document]
yceballost: Talked in teams, we mantain the current solution and we can include a prop (headerBackground) to set the color manually in the future (If needed)

Marcosld (approved):

src/table.tsx
```diff
@@ -51,11 +56,23 @@ export const Table = React.forwardRef(
...
             columnTextAlign,
+            rowVerticalAlign = 'middle',
             columnWidth,
             ...otherProps
         }: TableProps,
         ref: React.Ref<HTMLDivElement>
     ) => {
+        const getColumnTextAlign = (column: number) => {
+            const defaultTextAlign = 'left';
+            if (columnTextAlign) {
```
Marcosld, line 67: Perhaps you can use a default parameter to simplify this to: `return Array.isArray(columnTextAlign) ? columnTextAlign[column] : columnTextAlign`

Marcosld (approved):

yceballost (changes requested):

src/table.tsx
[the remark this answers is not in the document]
atabel: It works fine in my iPhone :/

yceballost (approved):
osom 🤩

src/table.css.ts
```diff
@@ -0,0 +1,243 @@
...
+export const collapsedRowsInMobile = style({});
+
+const BOXED_PADDING_Y_DESKTOP = 8;
+
+export const boxed = style({
+    padding: '0 16px',
+    paddingBottom: BOXED_PADDING_Y_DESKTOP,
+    '@media': {
+        [mq.tabletOrSmaller]: {
+            padding: '0 8px',
```
aweell, line 30: This padding left/right is also 16px in the mobile view

atabel: fixed

src/table.tsx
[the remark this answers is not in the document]
atabel: done

aweell (approved):

src/__screenshot_tests__/__image_snapshots__/table-screenshot-test-tsx-table-desktop-boxed-1-snap.png
marcoskolodny, line 1: I see a jump in the header items horizontal alignment when switching between 0 and 1 items in the story. Is that expected?

image

vs

image

marcoskolodny, line 1: When scrollOverResponsiveLayout is set to true, I see the left padding in the table, but I don't see one in the right side. Is that expected?

Left:

image

Right:

image

atabel: Yes, it's expected. It's how tables work, the column width is determined by the content, when there isn't any row, the width of the table is distributed between the columns taking into account only the width of the headings

atabel: I can't reproduce it, which set of args are you using apart from scrollOverResponsiveLayout=true ?

atabel: oh, I see this happens in FF, not in Chrome

atabel: should be fixed

marcoskolodny (approved):

src/__screenshot_tests__/__image_snapshots__/table-screenshot-test-tsx-table-desktop-boxed-1-snap.png
marcoskolodny: Looks good now!

pladaria (approved):

PR #1130, opened 2024-05-24
feat(Timer): create component

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1869)
Specs: [Link](https://www.figma.com/design/MU9VZyeyzTSDPLWmZMGYkK/%F0%9F%94%B8-[READY]-Timer?node-id=0-1&t=8aWjF0vZCDVBIhXq-0)


src/utils/helpers.tsx
```diff
@@ -141,6 +141,10 @@ export const isEqual = (a: unknown, b: unknown): boolean => {
         return true;
     }
 
+    if (typeof a === 'number' && typeof b === 'number' && isNaN(a) && isNaN(b)) {
```
marcoskolodny, line 144: According to [lodash documentation](https://lodash.com/docs/4.17.15#isEqual), `isEqual(NaN, NaN)` should return `true`:

image

But TS returns false when comparing NaN with itself:

image

pladaria: not TS, this is JS and it is expected to be false:

> NaN compares unequal (via [==](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Equality), [!=](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Inequality), [===](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality), and [!==](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_inequality)) to any other value — including to another NaN value.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN

marcoskolodny: Yes, It's expected in JS, but we don't want that for this helper

pladaria (approved):
nice!

playroom/snippets.tsx
```diff
@@ -2833,11 +2833,72 @@ const ProgressBlockSnippets = [
...
+    },
+    {
+        group: 'Timer',
+        name: 'Short Label',
+        code: `
+          <Timer
+            endTimestamp={Date.now() + 1000 * 60 * 60 * 24}
+            minTimeUnit="seconds"
+            maxTimeUnit="days"
+            labelType="short"
```
atabel, line 2856: does the `Timer` have `labelType`? I only see it in the props of `TextTimer`

playroom/snippets.tsx
```diff
@@ -2833,11 +2833,72 @@ const ProgressBlockSnippets = [
...
+            maxTimeUnit="days"
+            labelType="long"
+          />
+        `,
+    },
+    {
+        group: 'Timer',
+        name: 'TimerDisplay',
+        code: `
+          <TimerDisplay
```
atabel, line 2876: wrong component name?

src/__stories__/timer-story.tsx
atabel, line 1: when I set the `minTimeUnit` to seconds I can see days counter although I have the `maxTimeUnit` set to `undefined`, why is that?
image

src/theme.tsx
```diff
@@ -149,6 +179,21 @@ const TEXTS_DE: ThemeTexts = {
     counterQuantity: 'menge',
     counterMinValue: 'minimal',
     counterMaxValue: 'maximal',
+    timerDaysShortLabel: 'Tg.',
```
atabel, line 182: Are all these translations in the spec? Have them been defined by Localization team?

src/timer.tsx
```diff
@@ -0,0 +1,438 @@
...
+const SECOND_IN_MS = 1000;
+
+const MINUTE_IN_MS = MINUTE_IN_SECONDS * SECOND_IN_MS;
+const HOUR_IN_MS = HOUR_IN_MINUTES * MINUTE_IN_MS;
+const DAY_IN_MS = DAY_IN_HOURS * HOUR_IN_MS;
+
+export type TimeUnit = 'days' | 'hours' | 'minutes' | 'seconds';
+type Label = 'none' | 'short' | 'long';
+
+interface Timestamp {
```
atabel, line 29: this is a bad name, a timestamp is the number of seconds since 1970, so it's weird to be represented with this data structure.

I'd rename this to something like `RemainingTime` or similar

src/timer.tsx
```diff
@@ -0,0 +1,438 @@
...
+    seconds?: number;
+}
+
+interface BaseProps {
+    endTimestamp: Date | number;
+    minTimeUnit?: TimeUnit;
+    maxTimeUnit?: TimeUnit;
+    children?: void;
+    dataAttributes?: DataAttributes;
+    onProgress?: (value: Timestamp) => void;
```
atabel, line 42: should we have a example story using `onProgress`?

src/timer.tsx
```diff
@@ -0,0 +1,438 @@
...
+    hours?: number;
+    minutes?: number;
+    seconds?: number;
+}
+
+interface BaseProps {
+    endTimestamp: Date | number;
+    minTimeUnit?: TimeUnit;
+    maxTimeUnit?: TimeUnit;
+    children?: void;
```
atabel, line 40: not needed

src/timer.tsx
```diff
@@ -0,0 +1,438 @@
...
+    labelType?: Label;
+    minTimeUnit?: TimeUnit;
+    maxTimeUnit?: TimeUnit;
+    onProgress?: (value: Timestamp) => void;
+}) => {
+    const remainingTime = getRemainingTime(endTimestamp);
+    const [currentDays, setCurrentDays] = React.useState(remainingTime.days);
+    const [currentHours, setCurrentHours] = React.useState(remainingTime.hours);
+    const [currentMinutes, setCurrentMinutes] = React.useState(remainingTime.minutes);
+    const [currentSeconds, setCurrentSeconds] = React.useState(remainingTime.seconds);
```
atabel, line 137: why do we need all these states? I think a single state with the remaining time should be enough, the other states can be derived from it at render time

src/timer.tsx
```diff
@@ -0,0 +1,438 @@
...
+            labelType,
+            minTimeUnit,
+            maxTimeUnit
+        );
+
+        if (!isEqual(currentTimerValue, timerValue)) {
+            setTimerValue(currentTimerValue);
+            const timestampValue: Timestamp = {};
+            currentTimerValue.forEach((item) => (timestampValue[item.unit] = item.value));
+            onProgress?.(timestampValue);
```
atabel, line 213: why don't we call onProgress when we set the remaining time state?

src/timer.tsx
```diff
@@ -0,0 +1,438 @@
...
+                        : index === timerValue.length - 2
+                        ? ` ${texts.timerAnd} `
+                        : ', '
+                }`
+        )
+        .join('');
+
+    return (
+        <div
+            role="timer"
```
atabel, line 316: how is this announced by screen reader?
should we allow configuring `aria-live` prop? https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/timer_role#associated_wai-aria_roles_states_and_properties

src/utils/helpers.tsx
atabel: please include a unit test

src/__tests__/timer-test.tsx
```diff
@@ -0,0 +1,142 @@
...
+    await screen.findByText('0 días, 0 horas y 1 minuto');
+
+    act(() => jest.advanceTimersByTime(SECOND));
+    await screen.findByText('0 días, 0 horas y 0 minutos');
+
+    act(() => jest.runAllTimers());
+    await screen.findByText('0 días, 0 horas y 0 minutos');
+});
+
+test('Timer - renders only minimum unit if minTimeUnit is bigger than maxTimeUnit', async () => {
```
atabel, line 69: I miss a test for `<Timer endTimestamp={Date.now() + DAY} maxTimeUnit="seconds" />` finding text `"86400 segundos"`

src/timer.tsx
yceballost, line 1: TextTimer is not changing automatically to inverse context
image

[example](https://mistica-k136r440x-flows-projects-65bb050e.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8ARAlgZwA4BsCGAningC54DCeATlADoB2ABI5gJL0BuMVGMDzJNCRwwAvLRAAVISIn9GsDGCposgiPXEgUMJSrVoNcpoxgAPElTyjgSSeZLSAtt1P0oz3WSdZGAegA%2BAF8GQIYQIKA)

I talked days agowith Marcos about use the text with no format, but I think would be perfect if we can use an internal <Text> component to cover this detail avoiding wrap manually

If not, the solution is wrap textTimer over <Text> component but.. is it necessary? meh. Should be automatic.

```
<Text>
	<TextTimer labelType endTimestamp></TextTimer>
</Text>
```

yceballost (changes requested):

src/__stories__/timer-story.tsx
marcoskolodny: This is intended. Setting ONLY minTimeUnit sets max=days. Similarly, setting ONLY maxTimeUnit sets min=seconds.

playroom/snippets.tsx
marcoskolodny: Yeah. I forgot to update snippets after the name changes. my bad

src/theme.tsx
marcoskolodny: Yes

src/timer.tsx
marcoskolodny: Not needed. When setting the role="timer", the aria-live is automatically set to "off". And this is expected, we don't want it to announce changes every second

src/timer.tsx
marcoskolodny: Discussed offline. We call setState every second to update timer value internally and recompute everything. But if minTimeUnit is "minutes", we don't want to call onProgress every second

src/timer.tsx
marcoskolodny: The black text is expected. It's like when you put "hello" without wrapping inside a Text component.

Adding a Text wrapper as default may override some styles from the parent component. For example, adding an internal Text would break this:

<Text wordBreak={false}>
    <TextTimer/>
</Text>

Because wordBreak is set to true in Text component by default.

playroom/snippets.tsx
marcoskolodny: Old snippets. Changed

playroom/snippets.tsx
marcoskolodny: fixed

src/timer.tsx
marcoskolodny: fixed

src/timer.tsx
marcoskolodny: fixed

src/timer.tsx
marcoskolodny: Updated

src/timer.css.ts
```diff
@@ -0,0 +1,66 @@
...
+            [mq.tabletOrSmaller]: {
+                minWidth: pxToRem(56),
+            },
+        },
+    },
+]);
+
+const baseBoxedTimerValueContainer = style([
+    sprinkles({
+        padding: 8,
```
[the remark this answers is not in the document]
marcoskolodny: fixed

src/__screenshot_tests__/__image_snapshots__/private-skin-components-screenshot-test-tsx-components-in-o-2-alternative-1-snap.png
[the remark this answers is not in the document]
marcoskolodny: done

src/__tests__/timer-test.tsx
marcoskolodny: added

src/utils/helpers.tsx
marcoskolodny: done

src/timer.tsx
marcoskolodny: Done. Imitated carousel story

src/timer.tsx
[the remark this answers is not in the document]
marcoskolodny: Given that timer is not a specific part of the cards (it will be located in the extra slot), we don't have a "clean" way to force the overMedia behaviour. We could implement something like using react contexts inside the cards and using custom logic internally for the timer, but given that we will have to remove all of it when we create the "overMedia" variant, I don't think it's the best idea.

It will of course depend on how urgently we need this behaviour. Up to you @yceballost :)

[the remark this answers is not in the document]
marcoskolodny: Discussed offline. As mentioned by Yayo, we will iterate on this after we add the overMedia variant to Mistica

yceballost (approved):
top 🤩

atabel (approved):

src/__stories__/timer-story.tsx
atabel: This is surprising to me, why changing `minTimeUnit` affects `maxTimeUnit`?

marcoskolodny: If you don't set any of the props, min is Seconds and max is Hours.

If you set minTimeUnit="hours", but you don't specify max, it sounds better in my head to show every unit until the maximum one (instead of just showing hours)

PR #1133, opened 2024-05-30
fix(Spinner): use controlActivatedInverse token as default when used inside inverse variant

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1882)


src/spinner.tsx
```diff
@@ -5,6 +5,7 @@ import FadeIn from './fade-in';
 import {useAriaId, useTheme} from './hooks';
 import * as styles from './spinner.css';
 import {vars} from './skins/skin-contract.css';
+import {useIsInverseVariant} from './theme-variant-context';
 
 type Props = {
     color?: string;
```
yceballost, line 11: imo, this prop should be removed😛. should we show it in story controls?

src/spinner.tsx
yceballost, line 1: screenshot for inverse context or is it not necessary?

yceballost (approved):

src/spinner.tsx
atabel: If we don't want to support custom colors in spinner, we should deprecate the prop and remove it in the next major

atabel (approved):

src/spinner.tsx
marcoskolodny: We don't have screenshots of this component because of animations

src/spinner.tsx
marcoskolodny: Discussed offline. This is tricky, there are scenarios where you actually need to set a color. For example, in ButtonLinkDanger the spinner is red.

I'm afraid that this may break other scenarios, and I don't see as something bad that the color is customizable.

Given all this, we decided to align design specs in order to formally accept the customizable color: https://github.com/Telefonica/mistica-design/issues/1791

src/spinner.tsx
marcoskolodny: Actually, we never had screenshots of this component (at all). I've added them

src/spinner.tsx
atabel: lgtm

PR #1135, opened 2024-05-31
feat(HighlightedCard): support for alt for image

atabel:
WEB-1877


src/highlighted-card.tsx
```diff
@@ -119,6 +121,12 @@ const Content = React.forwardRef<HTMLDivElement, Props>((props, ref) => {
             </div>
             {imageUrl && (
                 <div
+                    {...(imageAlt
```
marcoskolodny, line 124: Maybe we could update this in order to use a Mistica Image component instead of a div

atabel: I woudn't touch this component too much. The idea is to deprecate/remove it mid-term

marcoskolodny (approved):

PR #1136, opened 2024-05-31
feat(Circle): custom background support

pladaria:


.github/workflows/ci.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         container: telefonica/web-builder:chromium93.0.4577-1.2.4
```
pladaria, line 21: This removes the warnings about node16 deprecation

src/__screenshot_tests__/circle-screenshot-test.tsx
```diff
@@ -12,3 +12,19 @@ test.each(CONTENT_OPTIONS)('Circle - %s', async (content) => {
...
+
+test('Circle with custom background', async () => {
+    await openStoryPage({
+        id: 'components-primitives-circle--default',
+        skin: 'O2-new',
+        args: {
+            content: 'none',
+            border: true,
+            size: 100,
+            customBackground: true,
```
pladaria, line 24: I'd prefer to allow setting the background style in the arg but storybook makes this impossible because it removes values with characters like "(". So you can't set a background style with an image

src/circle.tsx
```diff
@@ -24,15 +29,17 @@ const Circle: React.FC<Props> = ({
...
         <div
             className={styles.circle}
             style={{
                 width: size,
                 height: size,
-                backgroundColor,
-                backgroundImage: backgroundImage ? `url(${backgroundImage})` : 'none',
+                background: background
+                    ? background
+                    : `center / cover no-repeat ${imageStyle} ${backgroundColor || ''}`,
```
pladaria, line 42: react has issues when mixing shorthand properties with the full style

src/circle.css.ts
```diff
@@ -2,14 +2,13 @@ import {style} from '@vanilla-extract/css';
 import {sprinkles} from './sprinkles.css';
 
 export const circle = style([
+    {
+        boxSizing: 'border-box',
+    },
```
pladaria, line 7: the border was adding 2px to the size, we don't want this. Hopefully this doesn't break many screenshots in webapp

Winde (approved):

atabel (approved):

marcoskolodny (approved):

PR #1140, opened 2024-06-05
feat(EmptyState): allow using only ButtonLink as action

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1899)


src/__stories__/empty-state-story.tsx
```diff
@@ -9,29 +9,44 @@ export default {
...
         },
     },
+    args: {
+        actions: 'button and link',
+    },
 };
 
-type Args = {asset: string};
+interface Args {
+    actions: 'button and link' | 'button' | 'link';
```
atabel, line 23: Is having an action mandatory?
- If it's mandatory, I'd review the component props types
- If it isn't mandatory, I'd add a `'none'` option here and conditionally render the `<ButtonGroup/>` in the component

marcoskolodny: Added `none` as option

pladaria (approved):

atabel (approved):

PR #1141, opened 2024-06-05
fix(Select): set text color in native version

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-717)


src/__screenshot_tests__/__image_snapshots__/form-select-screenshot-test-tsx-select-display-all-options-with-native-select-1-snap.png
marcoskolodny, line 1: It looks like chrome has issues when setting the text color for native select. I've seen users setting color to `initial` in order to fix the problem. Doing this fixes the problem, but it also slightly changes the text color in light mode. I believe it's not a big deal, but maybe @yceballost can confirm if we can live with this.

src/select.tsx
```diff
@@ -365,7 +365,9 @@ const Select: React.FC<SelectProps> = ({
                             <option value="" style={{display: 'none'}} />
                         )}
                         {options.map(({value: val, text}) => (
-                            <option key={val} value={val}>
+                            // Set color: 'initial' to avoid wrong text color in some browsers when using dark mode.
```
marcoskolodny, line 368: Styling `<option/>` element is really tricky, most browsers completely ignore custom styles. The only browser that had the issue reported in the ticket is (afaik) Chrome. I've verified that this change fixes the problem in that browser and that none of the other ones are affected:

- FF / Chrome in Linux :heavy_check_mark:  
- Safari in a VM :heavy_check_mark: 
- Safari / Chrome in iOS :heavy_check_mark: 
- Chrome in Android :heavy_check_mark: 
- Chrome in Windows: pending

src/__screenshot_tests__/__image_snapshots__/form-select-screenshot-test-tsx-select-display-all-options-with-native-select-1-snap.png
yceballost: we can live with this absolutely :)

yceballost (approved):

pladaria (approved):

atabel (approved):

PR #1148, opened 2024-06-11
fix(Dialog, FixedFooterLayout, LoadingBar, NavigationBar, Overlay, Portal, Select, Sheet, Snackbar): avoid z-index issues

marcoskolodny:
And render provider's content inside its own stacking context to prevent it from covering the portals' content.

Issues:
- [1521](https://jira.tid.es/browse/WEB-1521)
- [1642](https://jira.tid.es/browse/WEB-1642)
- [1901](https://jira.tid.es/browse/WEB-1901)
- [1905](https://jira.tid.es/browse/WEB-1905)
- [1912](https://jira.tid.es/browse/WEB-1912)


src/sheet.tsx
```diff
@@ -253,18 +253,18 @@ const Sheet = React.forwardRef<HTMLDivElement, SheetProps>(({onClose, children,
                                 {typeof children === 'function'
                                     ? children({closeModal, modalTitleId})
                                     : children}
+                                <div className={styles.modalCloseButton}>
```
marcoskolodny, line 256: Moving close button inside the container with the role="dialog" (like it's done in other modals)

src/loading-bar.tsx
```diff
@@ -4,19 +4,29 @@ import {Portal} from './portal';
...
+import {isRunningAcceptanceTest} from './utils/platform';
 
 import type {DataAttributes} from './utils/types';
 
 type Props = {visible: boolean; children?: void; dataAttributes?: DataAttributes};
 
 const LoadingBar: React.FC<Props> = ({visible, dataAttributes}) => {
     return (
-        <Portal className={classnames(styles.portal, {[styles.hidden]: !visible})}>
+        <Portal
```
marcoskolodny, line 15: disable LoadingBar animations in acceptance tests

src/loading-bar.css.ts
```diff
@@ -14,17 +14,21 @@ const enterAnimation = keyframes({
...
         top: 0,
         left: 0,
         right: 0,
+        isolation: 'isolate',
     }),
     {
-        zIndex: 9999,
-        animation: `${enterAnimation} ${TRANSITION_DURATION_MS}ms ease-out`,
-        transition: `transform ${TRANSITION_DURATION_MS}ms ease-out`,
+        zIndex: 1, // Render above everything, including other portals
```
marcoskolodny, line 31: LoadingBar requires zIndex despite being rendered inside a portal, because we don't want it to be covered by other portals

src/snackbar.css.ts
```diff
@@ -18,6 +18,16 @@ export const snackbar = style([
     },
 ]);
 
+export const snackbarContainer = style([
+    sprinkles({
+        isolation: 'isolate',
+        position: 'relative',
+    }),
+    {
+        zIndex: 1, // Render above everything, including other portals
```
marcoskolodny, line 27: Snackbar requires zIndex despite being rendered inside a portal, because we don't want it to be covered by other portals

src/__screenshot_tests__/__image_snapshots__/private-components-in-portals-screenshot-test-tsx-components-with-portals-render-properly-mobile-ios-2-snap.png
yceballost, line 1: In my opinion, loadingBars should be placed under the modals because they are typically used to load content in the background. But it seems quite unusual to load something in the background within a sheet.

wdyt? @marcoskolodny @aweell

yceballost:
what do you thing about the idea to export tokens for zIndex? What is the way to work with all of these zindex when you build custom elements? How to ensure the good behavior through all of these layers?

src/theme-context-provider.tsx
```diff
@@ -165,44 +176,51 @@ const ThemeContextProvider: React.FC<Props> = ({theme, children, as, withoutStyl
...
+                                            </ScreenSizeContextProvider>
+                                        </AriaIdGetterContext.Provider>
+                                    </DocumentVisibilityProvider>
+                                </AspectRatioSupportProvider>
+                            </TrackingConfig>
+                        </ThemeContext.Provider>
+                    </TooltipContextProvider>
+                </ModalContextProvider>
+            </TabFocus>
+            <div ref={ref} style={{display: 'none'}} />
```
pladaria, line 222: consider this:
```suggestion
            {process.env.NODE_ENV !== 'test' && <div ref={ref} style={{display: 'none'}} />}
```
This way it won't be inserted in unit tests and won't pollute possible snapshots (like in button-test.tsx)

src/menu.tsx
```diff
@@ -394,7 +394,7 @@ export const Menu: React.FC<MenuProps> = ({
         <div {...getPrefixedDataAttributes(dataAttributes, 'Menu')}>
             {renderTarget({...targetProps, isMenuOpen})}
 
-            <Portal>
+            <Portal className={styles.wrapper}>
```
pladaria, line 397: I think it would be better if `Portal` adds this style by default. This way the isolation feature comes for free and you don't need to set the style in all usages

```
// portal.tsx
// ...
const newContainer = document.createElement('div');
// the style could be added here
```

src/__screenshot_tests__/__image_snapshots__/private-components-in-portals-screenshot-test-tsx-components-with-portals-render-properly-mobile-ios-2-snap.png
marcoskolodny: That is not what is happening nowadays in Master. There was a specific comment stating that loading bar should be rendered on top of everything.

src/loading-bar.tsx
atabel: hmmm... I think we already disabled all css animations with an injected css in acceptance tests

src/snackbar.css.ts
atabel: can you explain this case? how can this happen?
In the case of multiple portals, the last opened one will be over the others. Isn't this enough?

src/theme-context-provider.tsx
atabel: also, this div could be removed from the dom once we have done the job

src/snackbar.css.ts
marcoskolodny: Open a snackbar, and then open a modal or whatever while the snackbar is still there. You can test it in the private story I've created in this PR.

I mentioned this to design, and they agreed that the snackbar should not be covered by the  new portals

marcoskolodny:
> what do you thing about the idea to export tokens for zIndex? What is the way to work with all of these zindex when you build custom elements? How to ensure the good behavior through all of these layers?

Discussed offline. This was a misunderstanding, zIndex are not affecting any content rendered in portals.

yceballost (approved):
ameisin ✨

src/loading-bar.tsx
marcoskolodny: Oh, I didn't know this. That's nice, I'll remove this logic then

src/theme-context-provider.tsx
marcoskolodny: I've removed the div after applying the hack. This fixes also the issue with unit tests, so there is no need for the `NODE_ENV === 'test'` thing

src/menu.tsx
marcoskolodny: done

src/fixed-footer-layout.css.ts
```diff
@@ -49,6 +48,19 @@ export const container = style({
     },
 });
 
+export const portal = style([
```
marcoskolodny, line 51: Given that portals are now rendered with `isolation: isolate`, I had to add this to the FixedFooterLayout portal that is used to render the background (we want it to be rendered under all the other portals)

atabel: Please include code comments for these special cases. Otherwise, in two weeks you won't remember why we needed a zindex -1 here

atabel (approved):

src/theme-context-provider.tsx
```diff
@@ -91,6 +91,19 @@ const ThemeContextProvider: React.FC<Props> = ({theme, children, as, withoutStyl
...
 
+    const ref = React.useRef<HTMLDivElement>(null);
+    const [hasContentIsolation, setHasContentIsolation] = React.useState(false);
+
+    useIsomorphicLayoutEffect(() => {
+        // Set isolation: isolate to the parent of the provider. This way, we avoid content inside portals
+        // from being rendered under content that is inside the provider (in case it has z-index defined).
+        const root = ref.current?.parentElement;
+        if (root) {
+            root.style.isolation = 'isolate';
```
atabel, line 102: I think that, when the ThemeContextProvider has the `as` prop, the `isolation: isolate` style should be applied to the create dom node, instead of applying it to the parent

pladaria (approved):
nice! thanks :)

src/fixed-footer-layout.css.ts
marcoskolodny: done

src/theme-context-provider.tsx
marcoskolodny: done

aweell (approved):
Nice work ❤️

PR #1150, opened 2024-06-14
fix(ButtonLayout): add bleed when using only link

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1801)


src/button-layout.tsx
```diff
@@ -59,7 +59,7 @@ const ButtonLayout: React.FC<ButtonLayoutProps> = ({
 
     const linkContainer = link ? (
         <div
-            className={classnames(numberOfButtons > 1 ? styles.linkWithTwoButtons : styles.link)}
+            className={classnames(numberOfButtons !== 1 ? styles.linkWithTwoButtons : styles.link)}
```
atabel, line 62: the `styles.linkWithTwoButtons` name doesn't make sense now. I'd rename it to `styles.linkInNewLine` or something similar

pladaria (approved):

src/button-layout.tsx
marcoskolodny: fixed

PR #1152, opened 2024-06-19
fix(Cards): wrong headline padding in cards without subtitle/description

atabel:
WEB-1916


src/card.css.ts
```diff
@@ -395,7 +395,3 @@ export const flexColumn = style({
     display: 'flex',
     flexDirection: 'column',
 });
-
-globalStyle(`${flexColumn} > :last-child`, {
```
atabel, line 399: the problem with this rule is that `:last-child` uses dom order, not display order, so it was applying the styles to headline because was the last one in dom order

src/__screenshot_tests__/__image_snapshots__/data-card-screenshot-test-tsx-data-card-only-with-headline-and-title-1-snap.png
atabel, line 1: this screenshot covers the case

marcoskolodny (approved):
Good catch!

PR #1157, opened 2024-06-26
feat(Text): avoid ignoring line breaks

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1919)

[Failed screenshots in webapp](https://github.com/Telefonica/webapp/actions/runs/9679506648): all of them are expected, except this one:

modular-pages--modular-screen--modular-screen-tariff-details-screenshot-test-tsx-tariff-details-in-mobile-ios-1-diff

The problem in there is that the rendered component (`<Paragraphs/>`) wraps the text inside a `<p/>` that has `white-space: pre-wrap` in its styles. It can be fixed by simply swapping the order between `<p/>` and `<Text2/>` [in this line](https://github.com/Telefonica/webapp/blob/master/web/src/modular-pages/modules/components/paragraphs.tsx#L47).

We consider that this is a rare scenario, so when this PR is merged, we will release a new version of Mistica and upgrade webapp in order to fix it in there.


src/__screenshot_tests__/text-screenshot-test.tsx
```diff
@@ -27,3 +27,18 @@ test('Text wrapping', async () => {
     const image = await element.screenshot();
     expect(image).toMatchImageSnapshot();
 });
+
+test('Text wrapping - multiple whitespaces', async () => {
+    await openStoryPage({
+        id: 'components-text--text-wrapping',
+        device: 'DESKTOP',
+        args: {
+            text: '   a text with     spaces    everywhere   Pneumonoultramicroscopicsilicovolcanoconiosis test   ',
```
marcoskolodny, line 36: Sadly, I can't test an example where I use \n, because storybook has some weird behavior that removes all the args from the URL if I put the \ character

marcoskolodny: Good idea. Done :)

pladaria (approved):

PR #1158, opened 2024-06-27
fix(Carousel): when using renderBullets, don't hide the bullets automatically when there is only one page

atabel:


src/__screenshot_tests__/__image_snapshots__/carousel-screenshot-test-tsx-centered-carousel-with-only-one-item-1-snap.png
atabel, line 1: The "Page 0" text is [returned by `renderBullets`](https://github.com/Telefonica/mistica-web/blob/master/src/__stories__/centered-carousel-story.tsx#L41-L49)
This is the before/after diff: 
carousel-screenshot-test-tsx-centered-carousel-with-only-one-item-1-diff

sdiazserena (approved):

marcoskolodny (approved):

pladaria (approved):

PR #1161, opened 2024-07-01
WEB-1703 feat(css): export mistica tokens as public css custom properties

atabel:


examples/css/style.css
```diff
@@ -0,0 +1,123 @@
...
+    min-width: 80px;
+}
+
+button:hover {
+    background: var(--mistica-color-buttonPrimaryBackgroundHover);
+}
+
+button:active {
+    background: var(--mistica-color-buttonPrimaryBackgroundSelected);
+}
```
atabel, line 123: it could be relatively simple to include some basic classes for some components, like buttons or tags inside `mistica-common.css`

examples/css/index.html
atabel, line 1: this is an example of use

packages/generate-design-tokens/css-generator.js
```diff
@@ -0,0 +1,357 @@
...
+
+    const borderRadii = Object.entries(designTokens.radius)
+        .map(
+            ([radiusName, radiusDescription]) =>
+                `${buildBorderRadiusVarName(radiusName)}: ${buildRadius(radiusDescription)};`
+        )
+        .join('\n');
+
+    const textPresets = {
+        1: {
```
atabel, line 150: I've extracted these values from `text.tsx`, but I think these tokens should be in https://github.com/Telefonica/mistica-design/blob/production/tokens/movistar.json#L1384

packages/generate-design-tokens/css-generator.js
```diff
@@ -0,0 +1,357 @@
...
+    .${prefix}responsive-layout {
+        --${prefix}responsive-layout-margin: 40px;
+    }
+}
+
+${largeDesktopMediaQuery} {
+    .${prefix}responsive-layout {
+        --${prefix}responsive-layout-margin: calc((100vw - 1224px) / 2);
+    }
+}
```
atabel, line 355: does it make sense to include some utility classes like `.mistica-responsive-layout` here?

css/mistica-common.css
```diff
@@ -0,0 +1,158 @@
...
+    .mistica-responsive-layout {
+        --mistica-responsive-layout-margin: 40px;
+    }
+}
+
+@media (min-width: 1368px) {
+    .mistica-responsive-layout {
+        --mistica-responsive-layout-margin: calc((100vw - 1224px) / 2);
+    }
+}
```
atabel, line 158: Does it make sense to include some basic utility classes like this here?

css/mistica-common.css
```diff
@@ -0,0 +1,158 @@
+/* Default text color */
+[data-mistica-skin] {
+    color: var(--mistica-color-textPrimary);
+    background: var(--mistica-color-background);
+}
+
+/* border-radius utility classes */
+.mistica-border-radius-avatar {
```
atabel, line 8: I'm not sure if border-radius utility classes are worth, maybe having the css vars is enough

src/skins/tu.tsx
```diff
@@ -260,7 +260,7 @@ export const getTuSkin: GetKnownSkin = () => {
             divider: applyAlpha(palette.white, 0.05),
             dividerInverse: applyAlpha(palette.white, 0.05),
             navigationBarDivider: palette.darkModeBlack,
-            badge: palette.red70,
+            badge: palette.blue30,
```
atabel, line 263: is this change intended? I think the is't enough color contrast to read the number inside the badge now

css/mistica-common.css
yceballost: What is the use case? It seems like the CSS var could be enough

examples/css/style.css
yceballost: Nice, also export Boxed style could be nice

I have doubts with this mistica-common file
1. Need maintenance?
2. If we only include "some" could be open the door to force us to complete more this file

atabel: 1. Yes
2. Good point. I think, in case we include some basic components, we need to be clear about the scope and limitations of this. We can provide classes for basic stuff (components made by a single html element), but not for complex components like cards, because they are built by different pieces (title, subtitle, actions, etc) and we'd need classes for all the internal elements and the author would also need to replicate the correct html structure, so this doesn't scale. The most we can/should do in this cases is provide some examples, like the card I built here

yceballost: I like it!

You can also include `.mistica-boxed-inverse` but this is specifically what I was saying in point 2. xDD you can feed this document to infinity

src/skins/tu.tsx
aweell: It is, badges are blue in TU brand by design. They have a 6+:1 contrast, so it souldn't be a problem.

css/mistica-common.css
```diff
@@ -0,0 +1,165 @@
...
+.mistica-text-title1 {
+    font-size: var(--mistica-font-size-title1);
+    line-height: var(--mistica-line-height-title1);
+    font-weight: var(--mistica-font-weight-title1);
+}
+.mistica-text-title2 {
+    font-size: var(--mistica-font-size-title2);
+    line-height: var(--mistica-line-height-title2);
+    font-weight: var(--mistica-font-weight-title2);
+}
```
aweell, line 122: I wondering if we should add title3 to tokens.json so users can have all title sizes available. WDYT @yceballost

yceballost: Fine by me! and we will need to include also `title4` soon, right?

examples/css/style.css
```diff
@@ -0,0 +1,120 @@
...
+    align-items: center;
+    justify-content: center;
+    min-width: 56px;
+    padding: 4px 12px;
+    vertical-align: middle;
+}
+
+button {
+    color: var(--mistica-color-textButtonPrimary);
+    background: var(--mistica-color-buttonPrimaryBackground);
```
pladaria, line 97: you could define button as a colorless component and over that, build button-primary, button-secondary

pladaria (approved):

packages/generate-design-tokens/css-generator.js
aweell: Also, reviewing the values there's something weird, text4 and text5 have the same size and line height, and it seems that it starts in text 3, that should be size: 16px mobile and 18 desktop

src/skins/tu.tsx
yceballost: This is dark mode 🤔 I think Abel is right
https://camo.githubusercontent.com/daa3628439047ee6f4e6ede87c2fb3a069051d80a2aca7b8c83f73619275c1c4/68747470733a2f2f6465766d6973746963617765622e626c6f622e636f72652e77696e646f77732e6e65742f6d6973746963612d7765622d313731393939343934393638382f707269766174652d736b696e2d636f6d706f6e656e74732d73637265656e73686f742d746573742d7473782d636f6d706f6e656e74732d696e2d74752d696e76657273652d616e642d6461726b2d6d6f64652d312d646966662e706e67

packages/generate-design-tokens/css-generator.js
aweell: https://github.com/Telefonica/mistica-design/pull/1817

css/mistica-common.css
marcoskolodny: agree. If we need the class in the future, we can add it, but for now I think we can live with the vars

css/mistica-common.css
marcoskolodny: At least a basic one like responsive-layout, yes. But I wouldn't add many of them

packages/generate-design-tokens/css-generator.js
```diff
@@ -0,0 +1,364 @@
...
+            size: {mobile: 18, desktop: 20},
+            lineHeight: {mobile: 24, desktop: 28},
+        },
+        4: {
+            size: {mobile: 20, desktop: 28},
+            lineHeight: {mobile: 24, desktop: 32},
+        },
+        5: {
+            size: {mobile: 20, desktop: 28},
+            lineHeight: {mobile: 24, desktop: 32},
```
marcoskolodny, line 168: I think there is a typo in here. Text4 and Text5 have same values. Copy paste error?

marcoskolodny (approved):

packages/generate-design-tokens/css-generator.js
```diff
@@ -0,0 +1,364 @@
...
+            size: {mobile: 12, desktop: 14},
+            lineHeight: {mobile: 16, desktop: 20},
+        },
+        2: {
+            size: {mobile: 14, desktop: 16},
+            lineHeight: {mobile: 20, desktop: 24},
+        },
+        3: {
+            size: {mobile: 18, desktop: 20},
+            lineHeight: {mobile: 24, desktop: 28},
```
yceballost, line 160: size: {mobile: 16, desktop: 18},
lineHeight: {mobile: 24, desktop: 24},

packages/generate-design-tokens/css-generator.js
```diff
@@ -0,0 +1,364 @@
...
+        },
+        4: {
+            size: {mobile: 20, desktop: 28},
+            lineHeight: {mobile: 24, desktop: 32},
```
yceballost, line 164: size: {mobile: 18, desktop: 20},
lineHeight: {mobile: 24, desktop: 28},

packages/generate-design-tokens/css-generator.js
yceballost: Yes, and there is also a typo in [text story](https://mistica-web.vercel.app/?path=/story/components-text--text-components)

image

Text2 medium should show Text2 Medium 14/20 (Mobile) | 16/24 (Desktop)


according to specs
https://www.figma.com/design/aNHyvXe1mqjaelE8XWVQD5/%F0%9F%94%B9-Text-presets-definition?node-id=0-1

yceballost (changes requested):

examples/css/README.md
```diff
@@ -0,0 +1,168 @@
...
+For example:
+
+- `--mistica-color-textPrimary`
+- `--mistica-color-backgroundContainer`, etc.
+
+### Border radius tokens
+
+Naming convention:
+
+- `--mistica-border-radius-{radiusName}`
```
marcoskolodny, line 99: I think we decided not to create these tokens, right?

atabel: hmmm... no, we decided to remove the border-radius utility classes, but we keep the custom properties

yceballost (approved):

aweell (approved):

PR #1165, opened 2024-07-03
fix(ThemeContextProvider): avoid rendering isolation extra div in server and first client render

marcoskolodny:


src/theme-context-provider.tsx
```diff
@@ -222,7 +232,7 @@ const ThemeContextProvider: React.FC<Props> = ({theme, children, as, withoutStyl
                     </TooltipContextProvider>
                 </ModalContextProvider>
             </TabFocus>
-            {!hasContentIsolation && !as && <div ref={ref} style={{display: 'none'}} />}
+            {shouldRenderIsolationExtraDiv && <div ref={ref} style={{display: 'none'}} />}
```
atabel, line 235: I think you could encapsulate all this logic inside a component. This way you'd avoid re-renders to affect the whole app:
```tsx
{!as && <SetupStackingContext />}
```

```tsx
const SetupStackingContext = () => {
    const ref = React.useRef<HTMLDivElement>(null);
    const [hasContentIsolation, setHasContentIsolation] = React.useState(false);
    const [isFirstRender, setIsFirstRender] = React.useState(true);
    
    useIsomorphicLayoutEffect(() => {
        if (isFirstRender) {
            // Given that we don't render the extra div in server side, we skip creating it in the first client
            // render in order to avoid hydration issues
            setIsFirstRender(false);
        } else {
            // Set isolation: isolate to the parent of the provider. This way, we avoid content inside portals
            // from being rendered under content that is inside the provider (in case it has z-index defined).
            const root = ref.current?.parentElement;
            if (root) {
                root.style.isolation = 'isolate';
                setHasContentIsolation(true);
            }
        }
    }, [isFirstRender]);

    // Don't render the div in server side, because effects are not executed in there and it makes the div useless
    if (hasContentIsolation || !isClientSide() || isFirstRender) {
        return null
    }
    return <div ref={ref} style={{display: 'none'}} />
}
```

pladaria (approved):

src/theme-context-provider.tsx
marcoskolodny: Oh, nice one! Updated :)

atabel (approved):

PR #1171, opened 2024-07-09
feat(browser support): set min Safari version to 14

atabel:
Novum app no longer supports iOS 13: https://confluence.tid.es/display/CTO/%5BAPPS%5D+Novum+App+Information

### DIFF
--- a/src/screen-size-context-provider.tsx
+++ b/src/screen-size-context-provider.tsx
@@ -55,9 +55,11 @@ const ScreenSizeContextProvider: React.FC<Props> = ({children}) => {
             const listener = () => {
                 setState(mq.matches);
             };
-            mq.addListener(listener);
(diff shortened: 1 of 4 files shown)


src/screen-size-context-provider.tsx
```diff
@@ -55,9 +55,11 @@ const ScreenSizeContextProvider: React.FC<Props> = ({children}) => {
             const listener = () => {
                 setState(mq.matches);
             };
-            mq.addListener(listener);
+            mq.addEventListener('change', listener);
```
atabel, line 58: this is supported by iOS Safari >= 14

marcoskolodny (approved):

pladaria (approved):
nice! we should also update in webapp

PR #1172, opened 2024-07-09
fix(overscroll color): background color not changing when switching dark/light mode

atabel:
WEB-1943

### DIFF
--- a/src/overscroll-color-context.tsx
+++ b/src/overscroll-color-context.tsx
@@ -51,15 +51,44 @@ const TopOverscrollColor = () => {
     return null;
 };
 
+const useBodyBackgroundColor = () => {
+    const [bodyBackgroundColor, setBodyBackgroundColor] = React.useState<string>('transparent');
+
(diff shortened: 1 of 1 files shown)


src/overscroll-color-context.tsx
```diff
@@ -51,15 +51,44 @@ const TopOverscrollColor = () => {
     return null;
 };
 
+const useBodyBackgroundColor = () => {
+    const [bodyBackgroundColor, setBodyBackgroundColor] = React.useState<string>('transparent');
+
```
marcoskolodny, line 56: Why transparent? Until now we were using the body's backgroundColor as default.

pladaria (approved):

src/overscroll-color-context.tsx
atabel: yes, and we also avoid hydration issues, we can't read body background in SSR

marcoskolodny (approved):

PR #1175, opened 2024-07-18
feat(Table): allow setting any column as row header

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1948)
Specs: [Link](https://www.figma.com/design/WtF16bfvE9kKS4yx3VtLuE/%F0%9F%94%B8-Table-Specs?m=auto&t=mcZHVNAwa2CuBrCz-1)


src/table.tsx
marcoskolodny, line 1: Better review [with hidden whitespaces](https://github.com/Telefonica/mistica-web/pull/1175/files?diff=split&w=1)

Marcosld (approved):

jpastor-tuenti (approved):

src/table.tsx
```diff
@@ -115,6 +116,7 @@ export const Table = React.forwardRef(
             columnWidth,
             hideHeaders,
             scrollOverResponsiveLayout,
+            rowHeaderIndex = 0,
```
atabel, line 119: Does this mean we'll always have a `<th scope="row">`? I think we should support the case of not having one, and I'd say it should be the default case

marcoskolodny: Discussed offline. We decided to remove default header index and don't show a title in collapsed-rows mode if row's header index has not been specified

atabel (approved):

pladaria (approved):

aweell (approved):

PR #1176, opened 2024-07-18
feat(Callout): fix elements reading order and add role prop

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1952)


src/callout.tsx
atabel, line 1: please, move the inline styles to class names

marcoskolodny: done

aweell (approved):

PR #1178, opened 2024-07-19
feat(ThemeContextProvider): expose TextPreset size/lineHeight tokens as css vars

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1929)


src/__screenshot_tests__/private-text-preset-vars-screenshot-test.tsx
```diff
@@ -0,0 +1,25 @@
+import {openStoryPage, screen} from '../test-utils';
+
+const SKINS = ['Movistar', 'Vivo-new', 'O2-new'] as const;
```
marcoskolodny, line 3: Tested only in these skins because I don't see any value in adding all the possible ones here

pladaria (approved):

src/theme-context-provider.tsx
```diff
@@ -180,28 +190,49 @@ const ThemeContextProvider: React.FC<Props> = ({theme, children, as, withoutStyl
...
+    const themeVarsValues = {
         textPresets: textPresetsVars,
         colors,
         rawColors,
         borderRadii: theme.skin.borderRadii ?? defaultBorderRadiiConfig,
-    });
+    };
+
+    const responsiveThemeVarsValues = {
+        ...themeVarsValues,
```
atabel, line 233: does this mean you are repeating all the css vars, even those that aren't responsive (colors, radii, etc)? That would duplicate the size of the inline style

src/__private_stories__/text-preset-vars-story.tsx
```diff
@@ -0,0 +1,60 @@
...
+    children,
+}: {
+    weight?: string;
+    size?: string;
+    lineHeight?: string;
+    children: React.ReactNode;
+}) => {
+    return (
+        <Text>
+            <span style={{fontWeight: weight, fontSize: size, lineHeight}}>{children}</span>
```
atabel, line 22: why don't you pass these styles to `Text` using props?

src/theme-context-provider.tsx
marcoskolodny: Yep, good catch. I've fixed that

src/__private_stories__/text-preset-vars-story.tsx
marcoskolodny: Discussed offline. We would need a "reverse mapping" from number to weight strings (bold/light/regular/medium). Given that this is a private story, I don't see any benefit. If there is a need in the future, we can reiterate on this

atabel (approved):

PR #1179, opened 2024-07-23
fix(PreviewTools): render component as portal to avoid it from being covered by content

marcoskolodny:


yceballost (approved):

atabel (approved):
please, change the PR title to chore. We don't need this to appear in release notes

PR #1183, opened 2024-07-26
fix(Dialogs): avoid double execution of native dialogs

pladaria:
[WEB-1962](https://jira.tid.es/browse/WEB-1962)


src/dialog.tsx
```diff
@@ -213,9 +213,16 @@ const NativeModalDialog = ({
         message,
         title,
     });
+    const isOpenRef = React.useRef(false);
 
     React.useEffect(() => {
-        showNativeDialog(paramsRef.current);
+        if (isOpenRef.current) {
```
marcoskolodny, line 219: Maybe you can add a comment here explaining why is this needed

marcoskolodny (approved):

src/dialog.tsx
pladaria: ok, added

brtbrt (approved):

atabel (approved):

PR #1185, opened 2024-07-29
fix(Select): onChange was called twice and there was an empty option in iOS

atabel:
WEB-1953 WEB-1954

We want to support the case of a Select without a selected initial option. To support this we are including an additional default empty `<option>` in the `<select>`.

We style this empty option with `display: none` to avoid this option to appear in the select list when we open it. This works as expected in Chrome:
Screenshot_20240729_093422_Chrome
But it doesn't work as expected in Safari, and the empty option is shown when the select is open:
image

### The fix:
The best fix I've found is using the select's label for that empty option:
image
And once one of the options is selected, we remove the empty one. So, if we open the select again, the empty option isn't shown anymore (like in Chrome):
image

### DIFF
--- a/src/select.tsx
+++ b/src/select.tsx
@@ -301,6 +301,8 @@ const Select: React.FC<SelectProps> = ({
 
     const iconSize = pxToRem(20);
 
+    const selectedValue = valueState ?? value;
+
     return (
         <ThemeVariant isInverse={false}>
             {shouldUseNative || isServerSide ? (
@@ -317,7 +319,7 @@ const Select: React.FC<SelectProps> = ({
                             inputState={
                                 isFocused
                                     ? 'focused'
-                                    : value ?? valueState ?? inputRef.current?.value
+                                    : selectedValue ?? inputRef.current?.value
                                       ? 'filled'
                                       : 'default'
                             }
@@ -333,14 +335,11 @@ const Select: React.FC<SelectProps> = ({
                         )}
                         id={inputId}
                         aria-invalid={!!error}
-                        value={value ?? valueState}
+                        value={selectedValue}
                         required={!optional}
                         disabled={disabled}
                         onChange={(e) => {
                             setValue(e.target.value);
-                            if (onChangeValue) {
-                                onChangeValue(e.target.value);
-                            }
                         }}
                         onFocus={() => setIsFocused(true)}
                         onBlur={(e) => {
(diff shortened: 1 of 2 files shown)


src/select.tsx
```diff
@@ -333,14 +335,11 @@ const Select: React.FC<SelectProps> = ({
...
                         aria-invalid={!!error}
-                        value={value ?? valueState}
+                        value={selectedValue}
                         required={!optional}
                         disabled={disabled}
                         onChange={(e) => {
                             setValue(e.target.value);
-                            if (onChangeValue) {
-                                onChangeValue(e.target.value);
-                            }
```
atabel, line 343: this fixes https://jira.tid.es/browse/WEB-1953

src/select.tsx
```diff
@@ -357,12 +356,20 @@ const Select: React.FC<SelectProps> = ({
...
-                            // this is needed to allow a native select with no selected option
-                            <option value="" style={{display: 'none'}} />
+                        {!selectedValue && options.every(({value}) => !!value) && (
+                            // If no "empty" option exists, insert a dummy empty option. Once an option is selected,
+                            // this empty option is removed. This is needed to allow a native select without a selected
+                            // default option.
+                            // Chrome doesn't show this option when the select menu is open (because of display: none),
+                            // but Safari does. So we use the select's label for this option, otherwise it would be shown
+                            // as an empty option, which is weirder.
+                            <option value="" style={{display: 'none'}} aria-label="">
```
atabel, line 370: the empty `aria-label` is needed to avoid screen reader to read the select label twice

src/select.tsx
```diff
@@ -357,12 +356,20 @@ const Select: React.FC<SelectProps> = ({
...
                             opacity: 1,
+                            // Use transparent color for the empty option (show below) to avoid showing it when the menu is closed
+                            color: selectedValue ? undefined : 'transparent',
                         }}
                     >
-                        {options.every(({value}) => !!value) && (
-                            // if no "empty" option exists, insert a dummy empty option
-                            // this is needed to allow a native select with no selected option
-                            <option value="" style={{display: 'none'}} />
+                        {!selectedValue && options.every(({value}) => !!value) && (
```
pladaria, line 363: maybe microoptimization but you could cache this using useMemo

yceballost (approved):

src/select.tsx
marcoskolodny, line 1: Should we screenshot tests to cover the new functionality?

marcoskolodny (approved):

src/select.tsx
atabel: that's not possible, the new behavior only happens in iOS/Safari

PR #1186, opened 2024-07-29
fix(SheetRoot): Reduce transition duration to 0ms in test envs

Marcosld:


pladaria (approved):

src/sheet.css.ts
```diff
@@ -6,7 +6,7 @@ import {sprinkles} from './sprinkles.css';
 const insideSheetDialog = createVar();
 export const vars = {insideSheetDialog};
 
-export const transitionDuration = process.env.NODE_ENV === 'test' ? 0 : 400;
+export const transitionDuration = 400;
```
atabel, line 9: the problem with this change is now you have a time discrepancy between css and js. This `transitionDuration` is used in this file too. Can't this cause issues too?

atabel (approved):

src/sheet.css.ts
Marcosld: At first glance it should not cause any problems in unit test environment

marcoskolodny (approved):

PR #1190, opened 2024-07-30
chore(PreviewTools): remove deprecated brands from previewTools

yceballost:


marcoskolodny (approved):

atabel (approved):

atabel:
I think this could be confusing, because we still have o2/o2-new and vivo/vivo-new in playroom side pannel and in Storybook. Shouldn't we wait to deprecate/remove these skins and then remove them entirely from mistica?

yceballost:
> I think this could be confusing, because we still have o2/o2-new and vivo/[vivo-new](https://jira.tid.es/browse/VIVO-new) in playroom side pannel and in Storybook. Shouldn't we wait to deprecate/remove these skins and then remove them entirely from mistica?

all the old documentation about o2 "classic" was removed by the brand team in brand guidelines. This is just to keep the documentation portal (Brand Factory) aligned with brand guidelines. I think we can keep old brands in the rest of resources til remove them completely (storybook and playroom). 

Old brands in Mística only exist to support old products (until this product make the migration) but are not brands that should be used today.. so the documentation in brand factory should not show it

PR #1193, opened 2024-08-05
feat(Popover, Snackbar, Callout, Chip, Dialog, Cards): allow customizable close button label

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1966)

### DIFF
--- a/src/callout.tsx
+++ b/src/callout.tsx
@@ -25,7 +25,7 @@ type Props = {
     titleAs?: HeadingType;
     description: string;
     onClose?: () => void;
-    closeButtonAccessibilityLabel?: string;
+    closeButtonLabel?: string;
     icon?: React.ReactElement;
     button?: RendersNullableElement<typeof ButtonPrimary>;
     secondaryButton?: RendersNullableElement<typeof ButtonSecondary>;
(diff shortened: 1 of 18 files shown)


marcoskolodny:
> Can we add this to cards since they have an onClose and it's not possible to change that label?
> 
> [Specs are updated accordingly](https://www.figma.com/design/tKdPOfcUALzVIh5oizFbm7/branch/pArEzcgxyEMvdvg5YhCpvg/%F0%9F%94%B8-Cards-Specs?m=auto&node-id=9507-6893&t=Ez6NHQarD1yWaYQb-1)
> 
> I've seen that the same text `modalClose` is being used in sheet, guess is for desktop since is equal to a dialog?

@aweell I've updated all the cards. About the sheet: yes, it's because of desktop mode. But I'm not sure adding the prop only because of this makes sense, given that sheet is commonly used in mobile devices

aweell (approved):
LGTM!

src/callout.tsx
```diff
@@ -25,6 +25,7 @@ type Props = {
     titleAs?: HeadingType;
     description: string;
     onClose?: () => void;
+    closeButtonAccessibilityLabel?: string;
```
atabel, line 28: why so verbose? I'd call it just `closeButtonLabel`

atabel (approved):

PR #1195, opened 2024-08-08
fix(FeedbackScreen): avoid double responsive margins in buttons when they are not fixed

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1843)

### DIFF
--- a/src/feedback.tsx
+++ b/src/feedback.tsx
@@ -16,7 +16,7 @@ import {isRunningAcceptanceTest} from './utils/platform';
 import {Text6, Text2, Text3} from './text';
 import Box from './box';
 import {InternalBoxed} from './boxed';
-import ResponsiveLayout from './responsive-layout';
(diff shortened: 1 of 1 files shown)


src/feedback.tsx
marcoskolodny, line 1: Better review with [hidden whitespaces](https://github.com/Telefonica/mistica-web/pull/1195/files?diff=split&w=1)

yceballost (approved):
It is a production bug or just a super corner case?

marcoskolodny:
> It is a production bug or just a super corner case?

@yceballost it happens when you use a feedbackScreen and the height of the viewport is not big enough, so the button is not positioned as fixed:

image

It's a corner case, but it's good to fix it

atabel (approved):

atabel:
please, run a webapp preview build before merging this

PR #1199, opened 2024-08-19
fix(Inline): wrap content if required when space equals between/around/evenly

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1980)


yceballost (approved):
noiceee

atabel (approved):

PR #1201, opened 2024-08-20
chore(Storybook): fix story control naming inconsistencies

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1891)


pladaria (approved):

src/__stories__/callout-story.tsx
```diff
@@ -16,19 +16,12 @@ type Args = {
     title: string;
     description: string;
     actions: string;
-    withIcon: boolean;
-    isClosable: boolean;
+    icon: boolean;
```
yceballost, line 19: should we use "Asset" in this case? We have pending to change "icon" prop to "asset". Maybe we can go ahead with asset in stories? 

https://github.com/Telefonica/mistica-web/issues/626
image

marcoskolodny: Fixed in Callout, Dialog and Feedback

atabel (approved):

src/__stories__/feedback-screen-story.tsx
```diff
@@ -86,18 +86,18 @@ export const FeedbackScreenStory: StoryComponent<FeedbackScreenArgs> = ({
     multipleParagraphs,
     animateText,
     showIcon,
```
yceballost, line 88: icon instead?

yceballost (approved):

src/__stories__/feedback-screen-story.tsx
marcoskolodny: good catch. Fixed

PR #1202, opened 2024-08-21
feat(Titles): add new Title2, replace Title2 and Title3 with Title3 and Title4 respectively

marcoskolodny:
BREAKING CHANGE: renamed `Title2` to `Title3` and `Title3` to `Title4`.

Issue: [Link](https://jira.tid.es/browse/WEB-1982)


playroom/snippets.tsx
yceballost, line 1: "Title4 is not defined" in playroom. Missing to import in playroom maybe?

yceballost (changes requested):

src/__stories__/title-story.tsx
```diff
@@ -89,6 +89,18 @@ export const Title3Story: StoryComponent<Args> = ({title, right, linkText}) => (
     />
 );
 
+export const Title4Story: StoryComponent<Args> = ({title, right, linkText}) => (
+    <TitleComponent
+        as="h4"
```
atabel, line 94: I think these `as` are wrong, bigger tittle (Title4) should have the h1

atabel (approved):

pladaria (approved):

src/__stories__/title-story.tsx
marcoskolodny: Good catch. Fixed

PR #1204, opened 2024-08-21
feat(i18n): improve texts sizes

pladaria:
* Texts moved to `text-tokens.tsx`. Once the text tokens defined by design are ready we could autogenerate this file
* A `Token` is an object like: `{es: 'Patata', en: 'Potato', de: 'Kartofel', ...}`
* Each component imports (by name) the required tokens
* The provider still allows to override texts


src/callout.tsx
```diff
@@ -16,6 +16,7 @@ import {sprinkles} from './sprinkles.css';
 import {vars} from './skins/skin-contract.css';
 import {getPrefixedDataAttributes} from './utils/dom';
 import {applyCssVars} from './utils/css';
+import {translate, closeButtonLabel as closeButtonLabelText} from './text-tokens';
```
pladaria, line 19: Current token names are too generic and cause collisions. For example: `clearButton` or `loading`. 
We would define a naming convention... For example `TextTokenLoading` or use uppercase names `TEXT_LOADING`

I'm open to suggestions

src/callout.tsx
```diff
@@ -51,7 +52,10 @@ const Callout: React.FC<Props> = ({
     role,
 }) => {
     const variant = useThemeVariant();
-    const {texts} = useTheme();
+    const {
+        texts,
+        i18n: {locale},
+    } = useTheme();
```
pladaria, line 58: As you can see we now require to get the `locale` and import a `translate` function to get the translation. 
Also, the `locale` must be converted to language inside the `translate` function.

This can be improved.

`useTheme` could provide the `translate` function and the language can be precalculated in the context creation, so no extra imports (apart of the tokens) would be needed. 

Unless you think a better approach, I will make this change

src/callout.tsx
```diff
@@ -107,7 +111,11 @@ const Callout: React.FC<Props> = ({
                                 bleedRight
                                 Icon={IconCloseRegular}
                                 onPress={onClose}
-                                aria-label={closeButtonLabel ?? texts.closeButtonLabel}
+                                aria-label={
+                                    closeButtonLabel ||
+                                    texts.closeButtonLabel ||
+                                    translate(closeButtonLabelText, locale)
```
pladaria, line 117: Note that I'm using `||` instead of `??`. I think we can consider empty strings as missing

src/callout.tsx
pladaria: choose one?

```js
const {t} = useTheme(); // I prefer this one. `t` is commonly used for translations
const {translate} = useTheme(); 
const {i18n: {t}} = useTheme();
const {i18n: {translate}} = useTheme();
```

src/callout.tsx
pladaria, line 117: You will see this pattern a lot in following files:

```js
texts.patata || translate(patata)
```
I'm unable to think way to avoid having to write this in each token usage (unless we disable text customization)

src/callout.tsx
atabel: I think you can do `import * as texts from './text-tokens'`
the bundler should be smart enough to detect `texts.something` usages

src/callout.tsx
atabel: I like `const {translate} = useTheme();` but `t` is ok too
You could have a different hook just for translations too `const translate = useTranslate()`

pladaria: I prefer to keep it in the same useTheme hook. This way we avoid an additional import and the useTheme is required anyway to get the text overrides

atabel (approved):

src/callout.tsx
pladaria: It worked :D

src/callout.tsx
marcoskolodny: Maybe we could have some kind of utility function that does this?

const getTranslation = (token) => texts[token] || translate(token, locale)

marcoskolodny (approved):

src/callout.tsx
pladaria: that wouldn't work.

Imagine you want to translate the text "cancel"

```
const getTranslation = (token) => { // ----> is this token value the "cancel" string?
    if (texts.token) { // --->  this would be `texts['cancel']`
        return token;
    }
    translate(???, locale);  // ---> the translate function accepts as parameter a `Token` type, which is an object
    // like {es: 'cancelar', en: 'cancel'...}. We can't use the "cancel" string here as parameter
}

marcoskolodny: Hmm, I see :(

I think we can proceed with what we have now, and if we come up with some ideas in the future, we can iterate on this

PR #1205, opened 2024-08-21
feat(Feedback, FixedFooterLayout, Dialog, Accordion, Touchable, Header, IconButton, NavigationBar, OverscrollColor): remove deprecated props

marcoskolodny:
BREAKING CHANGE: deprecated props are removed in several components

Issue: [Link](https://jira.tid.es/browse/WEB-1991)


src/button-fixed-footer-layout.tsx
```diff
@@ -12,8 +10,6 @@ import type {RendersNullableElement} from './utils/types';
 type Props = {
     isFooterVisible?: boolean;
     button?: NullableButtonElement;
-    /** @deprecated */
-    desktopButtonAlign?: 'left' | 'center';
```
marcoskolodny, line 16: I'm not sure what's the expected behaviour in desktop if we don't allow this. As of now, it will be equivalent to mobile (`full-width`). I've checked in webapp and there are only two cases in which it's being used with `center`, and I don't know why

src/__acceptance_tests__/__ssr_pages__/header.tsx
```diff
@@ -6,19 +6,27 @@ import {
...
+                <Stack space={16}>
+                    <Stack space={8}>
+                        <Text3 regular color={skinVars.colors.textPrimary}>
+                            Cuota mensual (IVA incluido)
+                        </Text3>
+                        <Text8 color={skinVars.colors.textPrimary}>60,44 €</Text8>
+                    </Stack>
+
+                    <ButtonPrimary href="asdf">Descargar factura</ButtonPrimary>
+                    <Text3 regular>Y esto es un subtitulo</Text3>
```
pladaria, line 28: I think you could change this by a placeholder. We're testing the Header SSR, the extra content is unrelated to this test

src/__type_tests__/button-type-test.tsx
```diff
@@ -42,7 +37,7 @@ import {ButtonPrimary} from '..';
 
 <ButtonPrimary to="/foo">Text</ButtonPrimary>;
 
-<ButtonPrimary to="/foo" fullPageOnWebView>
+<ButtonPrimary to="/foo" newTab>
     Text
 </ButtonPrimary>;
```
pladaria, line 42: can you explain this? I don't find a deprecation message for `fullPageOnWebView`

atabel: Yes, I think it's still needed, and used in webapp

src/button-fixed-footer-layout.tsx
atabel: I wouldn't mind. Those are not production cases, so not a big deal if the buttons change alignment in desktop

src/__type_tests__/button-type-test.tsx
marcoskolodny: Restored

src/__acceptance_tests__/__ssr_pages__/header.tsx
marcoskolodny: done

atabel (approved):

src/dialog-context.tsx
```diff
@@ -98,7 +61,6 @@ export const DialogRoot = ({children}: DialogRootProps): JSX.Element => {
 
     return (
         <DialogContext.Provider value={value}>
-            <ExposeDialogFunctions />
```
pladaria, line 101: :+1:

pladaria (approved):

src/index.tsx
```diff
@@ -10,6 +10,7 @@ export {default as FadeIn} from './fade-in';
 
 export {AccordionItem, BoxedAccordionItem, Accordion, BoxedAccordion} from './accordion';
 export {ButtonPrimary, ButtonSecondary, ButtonDanger, ButtonLink, ButtonLinkDanger} from './button';
+export {default as ButtonGroup} from './button-group';
```
marcoskolodny, line 13: This is necessary because we are deprecating button/secondaryButton props in `Header` component (can be replaced by `extra` prop from `HeaderLayout).` The issue is that the only way to replicate the usage of button/secondaryButton with extra prop is by using ButtonGroup component. Unless we export it in Mistica, that's not possible (and there are at least 5 scenarios like this in webapp).

I've discussed with design and they believe it's okay to expose this component, because they already have it in the design docs (and it's even exposed in the Mistica storybook nowadays).

@pladaria  @atabel I re-requested your review in order to validate this.

pladaria (approved):

atabel (approved):

PR #1206, opened 2024-08-21
feat(Callout, Cards, Dialog, EmptyState, EmptyStateCard, FeedbackScreen, Tabs): remove icon prop and use only Icon/asset

marcoskolodny:
BREAKING CHANGE: icon/Icon/asset props are renamed in several components

Issue: [Link](https://jira.tid.es/browse/WEB-1992)

## :warning: Remark

The only component that is not following the convention is the `Sheet`, because changing this would create inconsistencies with the bridge.

## :spiral_notepad: Note

I've reviewed with @yceballost component by component and we concluded that we will proceed as follows (components in bold are affected by this PR):

#### <ins>Components with asset prop:</ins>

- **Callout**
- **Cards**
- **Dialogs**
- **EmptyState**
- **EmptyStateCard**
- **FeedbackScreen** (we want to allow Lottie animations in here for example, so Icon is not enough)
- Accordion (no changes needed)
- Lists (no changes needed)
- Popover (no changes needed)

#### <ins>Components with Icon prop:</ins>
- **Tabs**
- Chips (no changes needed)
- Avatar (no changes needed)
- Menu (no changes needed)
- Tag (no changes needed)
- InfoFeedbackScreen (no changes needed)


src/tabs.tsx
```diff
@@ -119,8 +119,12 @@ const Tabs: React.FC<TabsProps> = ({selectedIndex, onChange, tabs, dataAttribute
                                         aria-controls={ariaControls}
                                         aria-selected={isSelected ? 'true' : 'false'}
                                     >
-                                        <Inline space={!!icon && !!text ? 8 : 0} alignItems="center">
-                                            {icon && <div className={styles.icon}>{icon}</div>}
+                                        <Inline space={!!Icon && !!text ? 8 : 0} alignItems="center">
+                                            {Icon && (
+                                                <div className={styles.icon}>
```
atabel, line 124: I think we could remove this className and just add the needed dimensions to Icon

atabel (approved):

src/tabs.tsx
marcoskolodny: No, we can't. Svg elements by default have `display: inline-block`, and this would create a space below the icon unless we wrap it inside a div with fixed size (or we set `display: block` in the icon's className).

After discussing offline, we decided to leave the wrapper untouched.

Screenshot from 2024-08-22 12-08-15

image

src/tabs.tsx
```diff
@@ -33,7 +33,7 @@ export type TabsProps = {
     tabs: ReadonlyArray<{
         readonly text: string;
         readonly trackingEvent?: TrackingEvent | ReadonlyArray<TrackingEvent>;
-        readonly icon?: React.ReactNode;
+        readonly Icon?: React.FC<IconProps>;
```
pladaria, line 36: Consider `React.FC` as deprecated. You can use `React.ComponentType<IconProps>`

pladaria (approved):

PR #1209, opened 2024-08-23
fix(Vivinho char): vivinho char in headings being read as a separated heading

atabel:
WEB-1941

### DIFF
--- a/src/text.tsx
+++ b/src/text.tsx
@@ -60,7 +60,7 @@ const makeVivinhoCharReadableForScreenReaders = ({
         children.includes(VIVINHO_CHAR)
     ) {
         return {
-            ariaLabel: children.replaceAll(VIVINHO_CHAR, 'Vivo'),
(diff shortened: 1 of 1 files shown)


src/text.tsx
```diff
@@ -41,25 +41,51 @@ const vivinhoForScreenReaders = (
...
+    ariaLabel,
+    as,
+}: {
+    children: React.ReactNode;
+    ariaLabel?: string;
+    as?: React.ComponentType<any> | string;
+}) => {
+    // When the Text is a heading (<hx>), we set an aria-label replacing the vivinho char with "Vivo" for
+    // screen readers and hide the original text.
+    // If we used the generic solution, Safari/iOS VoiceOver would read "Vivo" as a separated heading
```
atabel, line 55: this is the same Safari/VoiceOver bug we faced with links. Screen reader reads every element as a separated one, giving the impression there are multiple headings

pladaria (approved):

marcoskolodny (approved):

PR #1211, opened 2024-08-23
feat(ButtonLink): add default variant

marcoskolodny:
BREAKING CHANGE: ButtonLink won't be small as default anymore

Issue:  [Link](https://jira.tid.es/browse/WEB-1625)

## :warning: Remark
The height of a small ButtonLink **(only in desktop)** was not equivalent to the one in any other small Button:

image

This was because according to design specs, the ButtonLink used `<Text3/>` in desktop, while other buttons use harcoded values for the text component. I've reviewed this with design and they agreed on updating the ButtonLink so it uses the same text sizes as the other buttons.

This change causes small visual changes in the small version of ButtonLink too, and that is expected.


yceballost:
remark note don't apply anymore, right?

marcoskolodny:
> remark note doesn't apply anymore, right?

@yceballost well, given that it's written in past tense, I don't see anything wrong about it. It's just a clarification that explains why there are visual changes in the small buttonLink too

src/button-group.tsx
```diff
@@ -21,6 +24,15 @@ const ButtonGroup: React.FC<ButtonGroupProps> = ({
     align = 'left',
     dataAttributes,
 }) => {
+    const linkContainerRef = React.useRef<HTMLDivElement>(null);
+    const [hasSmallLink, setHasSmallLink] = React.useState(false);
+
+    React.useLayoutEffect(() => {
+        if (linkContainerRef.current?.querySelector('[data-small-link=true]')) {
```
marcoskolodny, line 31: After discussing with @pladaria, this is the only reasonable solution we could come up with. We loose the RSC for ButtonGroup, but we don't think this is serious. The code is very small and it shouldn't have a huge impact

pladaria (approved):

src/button-group.tsx
marcoskolodny: Discussed offline. I've updated the logic to avoid these issues. Now it should be ok

atabel (approved):

PR #1213, opened 2024-08-26
feat(React): require react18 and replace useAriaId with React.useId

marcoskolodny:
BREAKING CHANGE: react17 is no longer supported by Mistica and useAriaId hook is removed

Issue: [Link](https://jira.tid.es/browse/WEB-2002)


src/form.tsx
```diff
@@ -47,7 +47,8 @@ const Form: React.FC<FormProps> = ({
     const fieldRegistrations = React.useRef(new Map<string, FieldRegistration>());
     const formRef = React.useRef<HTMLFormElement | null>(null);
     const {texts, t} = useTheme();
-    const id = useAriaId(idProp);
+    const reactId = React.useId();
+    const id = idProp ?? reactId;
```
pladaria, line 51: I would use `||` here and consider an empty id string as undefined

src/radio-button.tsx
```diff
@@ -64,7 +64,8 @@ const RadioButton: React.FC<PropsRender | PropsChildren> = ({
     ...rest
 }) => {
     const {disabled, selectedValue, focusableValue, select, selectNext, selectPrev} = useRadioContext();
-    const labelId = useAriaId(ariaLabelledby);
+    const reactId = React.useId();
+    const labelId = ariaLabelledby ?? reactId;
```
pladaria, line 68: same here and other places

src/form.tsx
marcoskolodny: done

src/radio-button.tsx
marcoskolodny: done

pladaria (approved):
nice!

PR #1215, opened 2024-08-27
feat(Package): convert to strict ESM

pladaria:
BREAKING CHANGE:
* This library is now a strict ESM package
* import icons from '@telefonica/mistica/icons'

---
Tasks in this PR
* [x] New package.json `exports` configuration
* [x] simplified `dist` folder
    * exported `css` files are now located in `dist/css` (physical location)
    * no temporary `dist-ts` folder for type definitions generation
* [x] `commonjs` scripts now use the `.cjs` extension
* [x] Icons export using `export * from`


package.json
```diff
@@ -7,37 +7,51 @@
...
+            "default": "./dist/css/*.css"
+        }
+    },
     "scripts": {
-        "test": "jest --config jest.unit.config.js",
-        "test-acceptance": "test-acceptance",
-        "test-ssr": "test-acceptance --config jest.ssr.config.js",
+        "test": "jest --config jest.unit.config.cjs",
+        "test-acceptance": "JEST_PUPPETEER_CONFIG=jest-puppeteer.config.cjs test-acceptance --config jest.acceptance.config.cjs",
+        "test-ssr": "JEST_PUPPETEER_CONFIG=jest-puppeteer.config.cjs test-acceptance --config jest.ssr.config.cjs",
```
pladaria, line 36: The `JEST_PUPPETEER_CONFIG` env var tells `jest-environment-puppeteer` where to find the `puppeteer` config.
This is needed because we're using an old version of the `jest-environment-puppeteer` library (v6) and it is unaware of `.cjs` files. Newer versions don't have this issue but we cannot upgrade because this requires a `puppeteer` upgrade.

packages/.eslintrc.json
```diff
@@ -0,0 +1,5 @@
+{
+    "env": {
+        "es6": true
+    }
+}
```
pladaria, line 5: We should update our jest configuration to understand `.cjs` files.

vite.config.js
```diff
@@ -44,6 +41,18 @@ export default defineConfig({
...
+                ...Object.keys(packageJson.dependencies),
+                ...Object.keys(packageJson.peerDependencies),
+                'react/jsx-runtime',
+                'moment/locale/de.js',
+                'moment/locale/es.js',
+                'moment/locale/en-gb.js',
+                'moment/locale/pt-br.js',
+                'moment/moment.js',
+                '@vanilla-extract/sprinkles/createRuntimeSprinkles',
+            ],
```
pladaria, line 55: For some reason, without this, a `node_modules` folder is created inside `dist` containing copies of all external packages, including `react` and `react-dom`. That produces a broken build because using mistica introduces two copies of react.

src/index.tsx
```diff
@@ -216,1753 +216,5 @@ export type {Locale} from './utils/locale';
...
-export {default as CommunityExampleComponent} from './community/example-component';
-export {default as CommunityAdvancedDataCard} from './community/advanced-data-card';
-export {
-    RowBlock as CommunityRowBlock,
-    SimpleBlock as CommunitySimpleBlock,
-    InformationBlock as CommunityInformationBlock,
-    HighlightedValueBlock as CommunityHighlightedValueBlock,
-    ValueBlock as CommunityValueBlock,
-    ProgressBlock as CommunityProgressBlock,
-} from './community/blocks';
```
pladaria, line 233: see `package.json`. There is now a dedicated export in `/community` for these components

src/index.tsx
```diff
@@ -216,1753 +216,5 @@ export type {Locale} from './utils/locale';
...
-export {default as iconKeywords} from './generated/mistica-icons/icons-keywords';
-
-// Exported this way to facilitate tree-shaking
```
pladaria, line 236: no more copy paste needed. The icons are exported as `export * from './generated/mistica-icons';` (see the end of this file)

PR #1228, opened 2024-08-28
feat(i18n): refactor texts to improve bundle size

pladaria:
BREAKING CHANGE: `texts` from `useTheme` are now empty by default. If a mistica defined text is required, it should be imported from `text-tokens` and must be translated using the `t` function from `useTheme`.

* Texts moved to `text-tokens.tsx`. Once the text tokens defined by design are ready we could autogenerate this file.
* A `Token` is an object like: `{es: 'Patata', en: 'Potato', de: 'Kartofel', ...}`
* Each component imports (by name) the required tokens
* The provider still allows to override texts


marcoskolodny (approved):

scripts/size-stats/app-test-lib-overhead/src/index.js
```diff
@@ -27,7 +27,10 @@ root.render(
                     i18n: {locale: 'es-ES', phoneNumberFormattingRegionCode: 'ES'},
                 },
             },
-            React.createElement(IconTruckFilled, {})
+            React.createElement(IconTruckFilled, {
+                // adding a single text token to verify that a single token is included, not the whole dictionary
+                key: textTokens.formCreditCardCvvLabel.en,
```
pladaria, line 32: Verified that this works as expected

atabel (approved):

PR #1229, opened 2024-08-29
fix(Buttons): avoid warnings related to change in order of react hooks

marcoskolodny:
If loadingText was empty and then it's changed to a non-empty string, the browser will complain because of a change in the order of hooks.

This happens because the `TextContentRenderer` function calls `useTheme()`. The issue is that one of the calls to `TextContentRenderer` was conditional: it was called only when loadingText is truthy. This causes the hook to be called more times if loadingText was falsy and suddenly changes to a truthy value.

### DIFF
--- a/src/button.tsx
+++ b/src/button.tsx
@@ -137,6 +137,20 @@ const renderButtonContent = ({
     const defaultIconSize = small ? styles.iconSize.small : styles.iconSize.default;
     const spinnerSizeRem = small ? styles.spinnerSize.small : styles.spinnerSize.default;
 
+    const buttonElement = renderButtonElement({
+        small,
+        content: children,
+        defaultIconSize,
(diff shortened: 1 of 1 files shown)


src/button.tsx
```diff
@@ -231,16 +235,7 @@ const renderButtonContent = ({
...
-                    <Box paddingLeft={8}>
-                        {renderButtonElement({
-                            small,
-                            content: loadingText,
-                            defaultIconSize,
-                            TextContentRenderer,
-                        })}
-                    </Box>
-                ) : null}
+                {loadingText ? <Box paddingLeft={8}>{loadingButtonElement}</Box> : null}
```
marcoskolodny, line 238: This line was causing the issue that I explain in the PR description

pladaria (approved):

PR #1231, opened 2024-09-02
fix(Touchable): add maybe types to touchable props

pladaria:
https://jira.tid.es/browse/WEB-1936

### DIFF
--- a/src/__type_tests__/chip-type-test.tsx
+++ b/src/__type_tests__/chip-type-test.tsx
@@ -1,6 +1,8 @@
 import * as React from 'react';
 import Chip from '../chip';
 
+import type {PressHandler} from '../touchable';
+
 <Chip>hello</Chip>;
 <Chip badge={3}>hello</Chip>;
 <Chip badge>hello</Chip>;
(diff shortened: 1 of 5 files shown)


Winde (approved):

src/__type_tests__/chip-type-test.tsx
```diff
@@ -62,3 +64,9 @@ import Chip from '../chip';
 <Chip to="/" onClose={() => {}}>
     hello
 </Chip>;
+
+const ChipMaybeOnPress = ({text, onPress}: {text: string; onPress?: PressHandler}): JSX.Element => (
+    <Chip maybe onPress={onPress}>
```
marcoskolodny, line 69: It feels weird to see something like `<Chip maybe/>`. Perhaps we could consider something like `maybeTouchable` or `maybeInteractive`.

Also, this would require to add `maybe` in all the problematic scenarios. I'm not sure this really solves the issue. I think leaving it as it is won't be a big issue.

Let's discuss this a bit better in the next weekly

pladaria:
Discarded for now

PR #1235, opened 2024-09-06
feat(BrandLoadingScreen): improve lottie animations

marcoskolodny:
In order to verify the improvements, I've modified locally the size-stats example and used bundle analizer to compare the results.

1. Used an online tool to compress the lottie JSON files for Vivo-new logo. This improved in 20% the total size of the lotties:
    - Before: 
MicrosoftTeams-image
 
   - After: 
MicrosoftTeams-image (1)

2. I noticed that the main bottleneck was using `lottie-react` to render the lotties. This lib imports `lottie-web`, and that library is quiet heavy. [Some people recommended](https://github.com/airbnb/lottie-web/issues/1184#issuecomment-411586909) importing `lottie-light` from `lottie-web` and using it as replacement for the default one.

    What I've done is copying a minimized version of ([lottie-react](https://github.com/Gamote/lottie-react/blob/main/src/components/Lottie.ts) into our repository and adapted it to use `lottie-light` instead of `lottie-web`. This caused a pretty big improvement (over 50% of the total size of the lib).

    - Before: 
MicrosoftTeams-image (2)

    - After: 
MicrosoftTeams-image (3)

In total, by using the exact same example in size-stats, we can see that the bundles size has been reduced by over 160kb:

- Before:
MicrosoftTeams-image (4)

- After:
MicrosoftTeams-image (5)

## :spiral_notepad: Note
The size stats comment in this PR only shows the improvements caused by compressing the vivo lotties. The ones obtained by replacing the lottie library are not visible because the example page in the size-stats doesn't use a BrandLoadingScreen component (and the lotties are imported lazily). Refer to the images I've attached in this description to see the impact.


src/lottie/index.tsx
marcoskolodny, line 1: Copied from https://github.com/Gamote/lottie-react/blob/main/src/components/Lottie.ts and modified in order to fix some errors and adapt to our use cases

src/vivinho-loading-animation/in-lottie.json
marcoskolodny, line 1: Used https://www.lottiemizer.com/ to compress the JSON files

src/lottie/hooks/use-lottie.tsx
```diff
@@ -0,0 +1,333 @@
...
+
+        const listeners = partialListeners.filter(
+            (listener: PartialListener): listener is Listener => listener.handler !== null
+        );
+
+        if (!listeners.length) {
+            return;
+        }
+
+        const animationEvents: Array<AnimationEventName> = ['complete', 'loopComplete'];
```
marcoskolodny, line 270: I had to add the listeners conditionally only for these two events because I was getting runtime (and TS) errors otherwise.

I've analyzed this and we should be fine by adding only these two events, because we never use the other ones in Mistica

src/lottie/hooks/use-lottie.tsx
```diff
@@ -0,0 +1,328 @@
...
+        );
+
+        if (!listeners.length) {
+            return;
+        }
+
+        const deregisterList = listeners.map((listener: Listener) => {
+            // In Mistica we only use these two events. If we try to add the event listener for all the other
+            // events from the listeners array, some runtime errors appear (and TS complains about wrong types).
+            const misticaAnimationEvents: Array<AnimationEventName> = ['complete', 'loopComplete'];
```
marcoskolodny, line 273: I had to add the listeners conditionally only for these two events because I was getting runtime (and TS) errors otherwise.

I've analyzed this and we should be fine by adding only these two events, because we never use the other ones in Mistica

marcoskolodny (commented):
I've tested rendering the BrandLoadingScreen in the Mistica storybook, and I can verify the improvements I've measured locally with bundle analyzer. As you can see, the total size of mistica bundles appears to have been reduced by around 170kb when using the BrandLoadingScreen (over a 10% improvement in the total bundle size for this page).

Also, notice how the bundle that contains the `main.` prefix is 200ms faster now (567ms vs 323ms)

Before:

image

After:
image

src/lottie/index.tsx
```diff
@@ -0,0 +1,80 @@
+'use client';
```
pladaria, line 1: Please, put a copy of the license inside the lottie folder
https://github.com/Gamote/lottie-react/blob/main/LICENSE

```
The above copyright notice and this permission notice shall be
included in all copies or substantial portions of the Software.
```

And a README.md explaining a bit why did we include these sources and where the original repository is

marcoskolodny: Done :)

yceballost (approved):

.yarn/patches/lottie-react-npm-2.4.0-d4d95c55aa.patch
marcoskolodny, line 1: Using `yarn patch` to import the light version of `lottie-web` instead of the default one that the library uses. This is the main improvement

atabel (approved):

pladaria (approved):

atabel (approved):

PR #1244, opened 2024-09-17
feat(PhoneNumberField): Custom formatter support + lazy load libphonenumber on demand

pladaria:
* load libphonenumber on demand
* allow custom formatter which allows to not use libphonenumber at all

### DIFF
--- a/src/phone-number-field.tsx
+++ b/src/phone-number-field.tsx
@@ -35,7 +35,7 @@ const PhoneInput = ({
     inputRef,
     value,
     defaultValue,
-    onChange,
+    onChange: onChangeFromProps,
     prefix,
     e164,
     format: formatFromProps,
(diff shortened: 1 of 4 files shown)


src/phone-number-field.tsx
```diff
@@ -21,14 +26,39 @@ type InputProps = Omit<React.InputHTMLAttributes<HTMLInputElement>, 'value' | 'o
...
+    const [isLibphonenumberLoaded, setIsLibphonenumberloaded] = React.useState(false);
+
+    React.useEffect(() => {
+        if (formatFromProps) {
+            formatRef.current = formatFromProps;
+        } else {
+            import('@telefonica/libphonenumber' /* webpackChunkName: "libphonenumber" */).then((lib) => {
+                libphonenumber = lib;
+                formatRef.current = formatPhoneUsingLibphonenumber;
+                setIsLibphonenumberloaded(true);
```
atabel, line 58: what happens if the user has introduced some text before libphonenumber has loaded? is the text formatted once the lib has loaded? Can we have a test of this case?

atabel (approved):

src/phone-number-field.tsx
pladaria: added a unit test to verify this case

atabel (approved):

PR #1246, opened 2024-09-18
feat(Form fields): improve accesibility of errors

atabel:
WEB-1958

- Added an icon to error messages (this makes the error message more accesible over inverse and for color-blind users)
- Added an aria-live region to announce all the fields with error when the form is submitted and there is more than one field with error
- Make sure all mandatory form fields are validated
- Form fields are validated onBlur. Previously, we only validated fields with a special format, like email.
- Fixed an error making VoiceOver read fields label twice


src/credit-card-expiration-field.tsx
```diff
@@ -112,9 +113,6 @@ const CreditCardExpirationField = ({
     const {setFormError, jumpToNext} = useForm();
 
     const validate = (value: ExpirationDateValue, rawValue: string): string | undefined => {
-        if (!rawValue) {
-            return optional ? '' : texts.formFieldErrorIsMandatory || t(tokens.formFieldErrorIsMandatory);
-        }
```
atabel, line 117: I've find this code repeated in most input fields, I've moved it inside `useFieldProps`

src/form-context.tsx
```diff
@@ -151,15 +163,22 @@ export const useFieldProps = ({
...
+            if (!rawValues[name] && !optional) {
+                error = texts.formFieldErrorIsMandatory || t(tokens.formFieldErrorIsMandatory);
+            } else if (validate) {
+                error = validate(values[name], rawValues[name]);
+            }
+            setFormError({name, error});
             onBlur?.(e);
         },
-        inputRef: (input: HTMLInputElement | null) => register(name, {input, validator: validate}),
+        inputRef: (input: HTMLInputElement | null) => register(name, {input, validator: validate, label}),
```
atabel, line 181: I need the field label to be in the registrations map, to be able to use it when rendering the global form error: "Revisa los errores en los siguientes campos: nombre, email"

src/text-field-base.tsx
```diff
@@ -336,6 +336,18 @@ export const TextFieldBase = React.forwardRef<any, TextFieldBaseProps>(
...
+                            style={labelStyle}
+                            error={error}
+                            forId={id}
+                            inputState={inputState}
+                            shrinkLabel={shrinkLabel}
+                            optional={!rest.required}
+                        >
+                            {label}
+                        </Label>
+                    )}
```
atabel, line 350: Moved the label before the input. This avoids a bug (I think it's a bug) in Safari/VoiceOver which made the label being read twice by screen reader (when focusing the input, and then as an independent element). This doesn't happen if the label appears before the input in the html.

src/cvv-field.tsx
```diff
@@ -42,7 +42,6 @@ const TooltipContent = ({acceptedCards}: {acceptedCards: CardOptions}) => {
                             <Text2>
                                 {texts.formCreditCardCvvTooltipAmex || t(tokens.formCreditCardCvvTooltipAmex)}
                             </Text2>
-                            )
```
atabel, line 45: image

src/__screenshot_tests__/__image_snapshots__/form-select-screenshot-test-tsx-select-error-1-snap.png
yceballost, line 1: in desktop, the text is not aligned in the same way as mobile, @aweell is this expected? 

is it possible to align the text line to the middle of the icon? (compare with mobile screenshot to see the difference)
image

yceballost (changes requested):

src/__screenshot_tests__/__image_snapshots__/form-select-screenshot-test-tsx-select-error-1-snap.png
aweell: Not expected, but also not correctly documented in the specs


the size of the icon could be 1lh to match the line height in mobile (16) but also in desktop (20) so even if the font-size increases appears aligned to the first line of text? this approach seems correct?

src/__tests__/form-test.tsx
```diff
@@ -78,6 +78,29 @@ test('custom validator', async () => {
     );
 });
 
+test('when there are multiple errors on submit, the fields with error are anounced by screen reader', async () => {
```
pladaria, line 81: what happens when multiple fields contain errors? All errors gets announced?

If so, can you add a test for this case?

pladaria (approved):

src/text-field-components.tsx
```diff
@@ -94,15 +96,20 @@ export const HelperText = ({leftText, rightText, error, id}: HelperTextProps): J
     return (
         <>
             {leftText && (
-                <div className={classnames(styles.helperText, styles.leftText)}>
-                    <Text1 color={leftColor} regular as="p" id={id}>
+                <p className={classnames(styles.helperText, styles.leftHelperText)}>
+                    {error && (
+                        <Box as="span" paddingRight={4}>
+                            <IconInformationRegular size="1rem" color={leftColor} />
```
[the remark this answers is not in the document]
atabel: fixed

src/__screenshot_tests__/__image_snapshots__/form-select-screenshot-test-tsx-select-error-1-snap.png
atabel: changed size to `1lh` (fallback to 1rem for old browsers not supporting `lh` unit)

aweell (approved):
LGTM!

marcoskolodny (approved):

yceballost (approved):

PR #1247, opened 2024-09-25
fix(Sheet): avoid content from rendering on top of sticky title/buttons

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-2016)


src/__screenshot_tests__/input-fields-screenshot-test.tsx
```diff
@@ -498,6 +498,13 @@ test.each`
     const field = await screen.findByLabelText('Label');
 
     await field.click({clickCount: 3});
+
```
marcoskolodny, line 501: Unrelated. This test seems to be unstable, I've created a task to review it: https://jira.tid.es/browse/WEB-2047

atabel (approved):

pladaria (approved):

pladaria (approved):
not directly related with this PR but I found that when the screen is not large enough, the text zone becomes unreadable. In this case, the fixed footer shouldn't be fixed

image

can be reproduced here: https://mistica-jp8vz0d1w-flows-projects-65bb050e.vercel.app/?path=/story/components-modals-sheet--actions

yceballost (approved):
🔝

PR #1250, opened 2024-09-25
feat(Sheet): lazy load sheet implementations

pladaria:
* `sheet-root` now has the bare minimum to work, it lazy loads everything when needed, including the native implementation.
* Web implementations of native sheets (defined by the bridge) have been deprecated. The recommended way to use that predefined sheets is using the `showSheet` function.

## Pending tasks

* [x] Update documentation
* [x] Verify in webapp
    * https://webapp-sheet.movistar-es-dev.svc.dev.tuenti.io/pages/example-bottom-sheet
    * [x] web
    * [x] android native
    * [x] ios native

## Before 

Bundles from the webview entrypoint

image

## After

The arrow points to the sheet-root module

image
image


wyunreal (approved):

wyunreal:
> * `sheet-root` now has the bare minimum to work, it lazy loads everything when needed, including the native implementation.
> * Web implementations of native sheets (defined by the bridge) have been deprecated. The recommended way to use that predefined sheets is using the `showSheet` function.
> 
> ## Pending tasks
> * [ ]  Update documentation
> * [ ]  Verify in webapp using real device
> 
> ## Before
> Bundles from the webview entrypoint
> 
> image
> 
> ## After
> The arrow points to the sheet-root module
> 
> image image

After:

https://i.giphy.com/media/v1.Y2lkPTc5MGI3NjExOXluNG41ZWJoc3RtZXFmbGQ2OGd2MTB4M3RzMm1lbGN5YnN1a3I3aCZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/ALtzQ6CHfC7vO5nRz7/giphy.gif

doc/texts.md
```diff
@@ -2,39 +2,9 @@
```
pladaria, line 2: Unrelated

package.json
```diff
@@ -25,7 +25,7 @@
         "ts-check": "tsc --project tsconfig.json --noEmit",
         "build": "yarn clean && node scripts/build.js",
         "compile": "node scripts/compile.js",
-        "clean": "rimraf dist && rimraf dist-es && rimraf dist-ts && rimraf css/mistica.css && rimraf community.d.ts && rimraf community.js && rimraf node_modules/.cache",
+        "clean": "rimraf dist && rimraf dist-es && rimraf dist-ts && rimraf css/mistica.css && rimraf community.d.ts && rimraf community.js && rimraf node_modules/.cache && bash -c 'yarn jest --clearCache > /dev/null 2>&1'",
```
pladaria, line 28: the redirect to /dev/null is to mute some false warnings

doc/sheet.md
```diff
@@ -3,78 +3,21 @@
...
 
 export const App = () => {
   return (
-    <>
-      <SheetRoot />
-      <RestOfYourApp />
-    </>
+    <SheetRoot>
+      <MyApplication />
+    </SheetRoot>
```
atabel, line 20: SheetRoot doesn't create a context, it's a singleton, so this shouldn't be needed

doc/texts.md
atabel: I had already changed this in my PR, but I'll revert my changes

src/__tests__/sheet-test.tsx
```diff
@@ -43,7 +47,7 @@ test('Sheet', async () => {
 
     const closeButton = await within(sheet).findByRole('button', {name: 'Close'});
 
-    await act(() => userEvent.click(closeButton));
+    await act(async () => userEvent.click(closeButton));
```
marcoskolodny, line 50: I know it was there before this PR, but the act() shouldn't be needed. useEvent.click() already uses it in its internal implementation. Maybe you can remove them

src/sheet-common.tsx
```diff
@@ -0,0 +1,393 @@
+'use client';
+/*
+ * Common code shared by web sheet components.
+ *
+ * IMPORTANT: Do NOT import anything from here into `sheet-root.tsx`
```
marcoskolodny, line 5: Maybe you can explain why in this comment. Otherwise, in a few months someone may read this code and they won't understand the reason behind this

marcoskolodny (approved):
Nice work!

doc/sheet.md
pladaria: It can be used both ways, but  I find more "react-ish" using it wrapping the children.

src/__tests__/sheet-test.tsx
pladaria: If I remove this act, the test won't pass. I get a timeout in the next assertion, even if I increase the timeout time

src/sheet-common.tsx
pladaria: ok

src/__tests__/sheet-test.tsx
marcoskolodny: Discussed offline. It seems that there is something weird happening with waitForElementToBeRemoved(). Let's change this to something like:

MicrosoftTeams-image

pladaria: as talked. The act was removed and the next expectation just checks that the element does not exist in the dom

PR #1252, opened 2024-09-26
chore(doc, snippets): Exposing Mística CSS in readme and snippet fixes

yceballost:
I took the opportunity to fix some snippets that was using old props and fixing some urls that are broken


playroom/snippets.tsx
marcoskolodny, line 1: I think some snippets using Header may be broken too, right?

atabel (approved):

playroom/snippets.tsx
yceballost: button prop is available in MainSectionHeader, I think it is ok. 

But thanks anyway, because I made changes in header snippets for isInverse
https://github.com/Telefonica/mistica-web/pull/1252/commits/b3a636eb6dbf63b4a4c28f73e23d5fc4e5445a65

marcoskolodny (approved):

PR #1254, opened 2024-09-27
feat(Align): create component

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-2045)

Following a similar approach to https://lens.loom.dev/components/align

As an example, if we want to align vertically the right content in the Row component, we can do something like:

image

And the result will be:

image


src/align.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from 'react';
+import {getPrefixedDataAttributes} from './utils/dom';
+import * as styles from './align.css';
+
+import type {DataAttributes} from './utils/types';
+
+type Props = {
+    x?: 'start' | 'center' | 'end';
+    y?: 'start' | 'center' | 'end';
```
yceballost, line 9: It is weird to me having props like `y` or `x`. I would expect something like `alignItems` or `justifyContent` like in `<Grid />` component. I'm looking for feature parity between those components

`<Grid alignItems="center" justifyItems="center"></Grid>`

https://mistica-web.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8UCWA3ABAZwC4E8AbGAXmGEwCMBDMAawHMAnCAVwDsp5MAdEZmPj6YAviIB8PdpkxIA4k1RRM1QqgbsAkrhgBbbCT5gY7HU2EALGOou5DIAIwAGJ%2BgvCAVqzyoAZvgBhCFMTOyNQmHMQSWkZWQDUJjBiHFQAL1JgABYnESpaRhYOKCDCCCYySDKmbAA6dhhWXCZVAAkbPIB6GJkkToUlGL60dBiQABoQXCtdGGwEAG0QAFkIdFQ8aiYQAF1JgHclafn4BYBmADYnHZEgA

src/align.tsx
yceballost, line 1: What do you think about to have an additional property like `height` or maybe accept `style`?

for example, in this case this new component isn’t really helping me much because I need this extra div to set the height
https://tinyurl.com/23wyhrme

marcoskolodny: Yeah, sounds good. I've added it (fallback to 100% if not specified)

src/align.tsx
yceballost: got it!

src/__stories__/align-story.tsx
```diff
@@ -0,0 +1,37 @@
+import * as React from 'react';
+import {Align, Text2} from '..';
+
+export default {
+    title: 'Layout/Align',
+    argTypes: {
```
yceballost, line 6: missing `width` + `height` controls?

yceballost (approved):

pladaria (approved):
nice and clean :sparkles:

src/align.tsx
```diff
@@ -0,0 +1,35 @@
...
+    width = '100%',
+    height = '100%',
+    children,
+    dataAttributes,
+}: Props): JSX.Element => {
+    return (
+        <div
+            {...getPrefixedDataAttributes(dataAttributes, 'Align')}
+            className={styles.container}
+            style={{placeItems: `${y} ${x}`, width, height}}
```
atabel, line 28: Not a big deal, but I think default width/height could be in the css class. This would reduce the html size when using SSR.

atabel (approved):

src/align.tsx
marcoskolodny: ok, done

src/__stories__/align-story.tsx
marcoskolodny: added

PR #1255, opened 2024-09-27
fix(Touchable): make content focusable when using onPress + as="a"

marcoskolodny:


src/text-link.tsx
```diff
@@ -64,10 +64,10 @@ const TextLink = ({
         <BaseTouchable
             {...props}
             stopPropagation
-            as={props.onPress ? 'a' : undefined}
```
marcoskolodny, line 67: Using <a/> without setting href will cause the element to not be focusable. This is how HTML works and it's expected.

It makes more sense for the TextLink to be rendered as a button if we are passing onPress instead of href

pladaria (approved):

atabel (approved):

PR #1257, opened 2024-10-01
feat(vanilla-extract): remove sprinkles usages from TS files

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-2051)

Before:

before

After:

image


src/__screenshot_tests__/__image_snapshots__/private-image-srcset-screenshot-test-tsx-image-with-src-set-is-responsive-desktop-1-snap.png
marcoskolodny, line 1: These are small differences in rendering caused by removing the display: inline-block style that we were applying with sprinkles.

Removing sprinkles from the stories was not necessary because this is not included into bundles, but I've done it for consistency with other TS files

src/box.tsx
```diff
@@ -49,21 +50,36 @@ const Box = React.forwardRef<HTMLDivElement, Props>(
...
-            // if this fails, it's because the consumer passed in a value that is not a valid padding size
-            // fallback to inline styles in that case.
-            paddingStyles = {
-                paddingTop: typeof paddingTop === 'object' ? paddingTop.mobile : paddingTop,
-                paddingBottom: typeof paddingBottom === 'object' ? paddingBottom.mobile : paddingBottom,
-                paddingLeft: typeof paddingLeft === 'object' ? paddingLeft.mobile : paddingLeft,
-                paddingRight: typeof paddingRight === 'object' ? paddingRight.mobile : paddingRight,
-            };
-        }
+        const paddingTopValues =
```
marcoskolodny, line 53: I couldn't come up with a way to use sprinkles in Box and Grid components (because I don't know how to replicate the fallback logic with pure CSS).

The only way I was able to make this work is to use CSS vars and `style()` instead of `sprinkles()` in the corresponding CSS files.

src/tooltip.tsx
```diff
@@ -619,15 +618,15 @@ const Tooltip = ({
     return (
         <BaseTooltip
             content={
-                <Box className={styles.content}>
```
marcoskolodny, line 622: This was working just by luck. The Box component sets padding to 0 by default, and in this case the padding is being applied also through className prop.

When changing the logic in the Box CSS file to use vars instead of sprinkles, the padding from props took precedence over the one passed through className (I guess it depends on how vanilla determines the one that "wins").

In this case, given that we are not using padding props at all, it doesn't actually makes sense to use Box here. I've replaced it with a div to avoid the issue I've mentioned before

marcoskolodny: I've also detected some use cases in webapp where padding is being applied through className instead of props. These will fail because of the same reason, so we may need to fix them after merging this PR:

- web/src/pages/cancellation-with-retention-v3/v1/components/list-offers/index.tsx
- web/src/pages/mv-forms/forms/cancelamento-sva/commons/infos/index.tsx
- web/src/pages/mv-forms/forms/problemas-cobrancas/commons/infos/index.tsx
- web/src/pages/mv-forms/forms/suporte-tecnico/commons/infos/index.tsx
- web/src/pages/mv-forms/forms/suspensao-temporaria/commons/choice/index.tsx
- web/src/pages/mv-forms/forms/suspensao-temporaria/commons/infos/index.tsx
- web/src/pages/update-contacts-vivo-code/components/update-contact-parcial/index.tsx

src/form.tsx
```diff
@@ -225,7 +225,7 @@ const Form = ({
                 id={id}
                 onSubmit={handleSubmit}
                 ref={formRef}
-                className={classnames(sprinkles({width: '100%'}), className)}
+                className={classnames(styles.form, className)}
```
pladaria, line 228: hmm, not sure why was this 100% width needed. The `<form>` element is 100% width by default

src/tooltip.tsx
pladaria: As commented in teams, I'd deprecate the className prop in the Box component. If someone wants a div with styles, then they can use a div.

pladaria (approved):
Is is possible to add an eslint rule to forbid sprinkes imports in .tsx files?

src/form.tsx
marcoskolodny: Discussed offline. We are not really sure why this was added in the first place, and I'm afraid of breaking something if we remove it. After all, having that doesn't really break anything anyways

marcoskolodny:
> Is is possible to add an eslint rule to forbid sprinkes imports in .tsx files?

Yes, I've added it here :)

src/tooltip.tsx
marcoskolodny: Done, deprecated className

atabel (approved):

PR #1261, opened 2024-10-04
chore(Storybook): fix theme default values for stories opened in a new tab

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-2064)

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -63,16 +63,14 @@ const getTheme = (
     selectedPlatform?: Platform,
     selectedColorScheme?: ColorScheme
 ): ThemeConfig => {
-    const skin = selectedSkin ?? MOVISTAR_SKIN;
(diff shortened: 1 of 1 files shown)


.storybook/preview.tsx
```diff
@@ -63,16 +63,14 @@ const getTheme = (
     selectedPlatform?: Platform,
     selectedColorScheme?: ColorScheme
 ): ThemeConfig => {
-    const skin = selectedSkin ?? MOVISTAR_SKIN;
-    const platform = selectedPlatform ?? 'desktop';
```
marcoskolodny, line 67: When opening a story in a new tab without using any url params for the skin/platform/colorScheme, the platform was being set to desktop as fallback. This is wrong and it should have been fixed when we added the logic to avoid flashes between first and second render in storybook (https://github.com/Telefonica/mistica-web/pull/1048).

atabel (approved):

PR #1264, opened 2024-10-07
fix(Sheet): lock focus inside sheet container when content is not interactive

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-2018)


src/sheet-common.tsx
```diff
@@ -237,6 +237,7 @@ const Sheet = React.forwardRef<HTMLDivElement, SheetProps>(({onClose, children,
                     {...dragableSheetProps}
                     {...getPrefixedDataAttributes(dataAttributes, 'Sheet')}
                     ref={ref}
+                    tabIndex={-1}
```
atabel, line 240: I think the focus should be moved to the element with role=dialog

marcoskolodny: You're right. Done :)

package.json
```diff
@@ -79,7 +79,7 @@
         "@swc/cli": "^0.1.62",
         "@swc/core": "^1.3.95",
         "@swc/jest": "^0.2.29",
-        "@telefonica/acceptance-testing": "5.0.0",
+        "@telefonica/acceptance-testing": "5.1.0",
```
marcoskolodny, line 82: unrelated, just upgrading this lib to have the changes in https://github.com/Telefonica/js-toolbox/pull/107

pladaria (approved):

PR #1265, opened 2024-10-08
chore(Playroom): add support for using Lottie animations

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-1712)

### DIFF
--- a/playroom/components/animation.tsx
+++ b/playroom/components/animation.tsx
@@ -1,0 +1,32 @@
+import * as React from 'react';
+import Lottie from 'lottie-react';
+
+import type {LottieComponentProps} from 'lottie-react';
+import type {ExclusifyUnion} from '../../src/utils/utility-types';
+
+type Props = Omit<LottieComponentProps, 'animationData'> &
+    ExclusifyUnion<{animationData: unknown} | {animationUrl: string}>;
+
+const Animation = ({animationData: animationDataProp, animationUrl, ...props}: Props): JSX.Element | null => {
+    const [animationData, setAnimationData] = React.useState<unknown>();
+
+    React.useLayoutEffect(() => {
+        if (animationUrl) {
+            fetch(animationUrl)
+                .then((response) => {
+                    if (response.ok) {
+                        response.json().then((jsonData) => setAnimationData(jsonData));
+                    }
+                })
+                .catch(() => {
+                    setAnimationData(undefined);
+                });
+        } else {
+            setAnimationData(animationDataProp);
+        }
+    }, [animationUrl, animationDataProp]);
+
+    return animationData ? <Lottie animationData={animationData} {...props} /> : null;
+};
+
+export default Animation;
--- a/playroom/components/animation.tsx
+++ b/playroom/components/animation.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import Lottie from 'lottie-react';
+import Loader from './loader';
 
 import type {LottieComponentProps} from 'lottie-react';
 import type {ExclusifyUnion} from '../../src/utils/utility-types';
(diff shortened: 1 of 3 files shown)


playroom/components/index.tsx
```diff
@@ -36,7 +36,7 @@ import type {ThemeConfig, ColorScheme, KnownSkinName, IconProps} from '../../src
 export * from '../../src';
 export * from '../../src/community';
 export {default as Loader} from './loader';
-export {default as ButtonGroup} from '../../src/button-group';
```
marcoskolodny, line 39: This was old. ButtonGroup is exported by Mistica since the last major release

playroom/components/animation.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
+import Lottie from 'lottie-react';
+
+import type {LottieComponentProps} from 'lottie-react';
+import type {ExclusifyUnion} from '../../src/utils/utility-types';
+
+type Props = Omit<LottieComponentProps, 'animationData'> &
+    ExclusifyUnion<{animationData: unknown} | {animationUrl: string}>;
```
marcoskolodny, line 8: The component allows all the same props as `<Lottie />`, and also `animationUrl` for the cases in which you want to download the JSON data from an url

yceballost (approved):
incredibile Ɛ> 

can you include a simple snippet? 

`<Animation animationUrl="https://raw.githubusercontent.com/Telefonica/mistica-icons/refs/heads/production/animated-icons/Vivo/success.json" />`

playroom/components/animation.tsx
```diff
@@ -0,0 +1,32 @@
...
+                });
+        } else {
+            setAnimationData(animationDataProp);
+        }
+    }, [animationUrl, animationDataProp]);
+
+    return animationData ? <Lottie animationData={animationData} {...props} /> : null;
+};
+
+export default Animation;
```
atabel, line 32: why not something like this? we already have a Loader component in Playroom:
```suggestion
const Animation = ({animationData, animationUrl, ...props}: Props): JSX.Element | null => {
    if (animationData) {
        return <Lottie animationData={animationData} {...props} />;
    }
    return (
        <Loader
            load={animationUrl}
            render={(animationData) => <Lottie animationData={animationData} {...props} />}
        />
    );
};

export default Animation;
```

playroom/components/animation.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
+import Lottie from 'lottie-react';
+
+import type {LottieComponentProps} from 'lottie-react';
+import type {ExclusifyUnion} from '../../src/utils/utility-types';
+
+type Props = Omit<LottieComponentProps, 'animationData'> &
+    ExclusifyUnion<{animationData: unknown} | {animationUrl: string}>;
+
+const Animation = ({animationData: animationDataProp, animationUrl, ...props}: Props): JSX.Element | null => {
```
pladaria, line 10: Won't this create false expectations? If people see this component in Playroom, they may think that it is available in Mistica.

Perhaps we could give it an ugly name like: `PlayroomOnlyLottieAnimation`

playroom/components/animation.tsx
marcoskolodny: I forgot completely about the Loader component, this is a perfect use case for it. Good catch!

Done :)

playroom/components/animation.tsx
marcoskolodny: We have the same scenario with <Loader/> component nowadays and no one complained about it (at least for now). I don't think this is a serious issue.

If this issue arises in the future, we can rename these components if needed, but for now I think we can live with this :)

pladaria (approved):

atabel (approved):

PR #1269, opened 2024-10-16
feat(Row): allow aria-label in informative rows

atabel:
https://jira.tid.es/browse/WEB-2061
By default, the browser will read text content inside rows one by one (title, description, etc).
With the changes in this PR, if an aria-label is provided to the informative row, the whole row will be read as a single text (the one provided in the aria-label prop) 

This PR also fixes https://jira.tid.es/browse/WEB-1885 which is a basically revert of https://github.com/Telefonica/mistica-web/pull/942. As part of this PR we allow passing custom styles (via `className` prop) to `ScreenReaderOnly` to workaround possible style issues found (like the one described originally in https://github.com/Telefonica/mistica-web/pull/942).

### DIFF
--- a/src/screen-reader-only.css.ts
+++ b/src/screen-reader-only.css.ts
@@ -13,9 +13,8 @@ export const screenReaderOnly = style([
         width: 1,
         height: 1,
         clip: 'rect(0, 0, 0, 0)',
+        clipPath: 'inset(50%)',
+        whiteSpace: 'nowrap',
         userSelect: 'none',
-        // move the element out of the screen to avoid scrolling issues. See https://github.com/Telefonica/mistica-web/pull/942
-        left: -10000,
-        top: -10000,
     },
 ]);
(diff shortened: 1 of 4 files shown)


src/screen-reader-only.css.ts
```diff
@@ -13,9 +13,8 @@ export const screenReaderOnly = style([
         width: 1,
         height: 1,
         clip: 'rect(0, 0, 0, 0)',
+        clipPath: 'inset(50%)',
```
atabel, line 16: this seems to be the modern version of the deprecated `clip: rect(0, 0, 0, 0)`

src/screen-reader-only.css.ts
```diff
@@ -13,9 +13,8 @@ export const screenReaderOnly = style([
...
+        whiteSpace: 'nowrap',
```
atabel, line 17: I've found other libs include this too.
Some references:
https://kittygiraudel.com/snippets/sr-only-class/
https://tailwindcss.com/docs/screen-readers

marcoskolodny (approved):

pladaria (approved):

PR #1274, opened 2024-10-21
feat(Spinner): improve a11y

marcoskolodny:
- `role = "progressbar"`: I've found some examples online where spinner is represented as a indeterminate progress bar. For example, [Material does this](https://m2.material.io/components/progress-indicators):

image

image

- `aria-polite`: we don't need changes announcements to be assertive.

- `aria-busy`: to express that content is being loaded:
image

### DIFF
--- a/src/__tests__/spinner-test.tsx
+++ b/src/__tests__/spinner-test.tsx
@@ -1,0 +1,31 @@
+import * as React from 'react';
+import {Spinner} from '..';
+import {render, screen} from '@testing-library/react';
+import ThemeContextProvider from '../theme-context-provider';
+import {makeTheme} from './test-utils';
+
+test('spinner is accessible', () => {
+    render(
+        <ThemeContextProvider theme={makeTheme()}>
+            <Spinner />
+        </ThemeContextProvider>
+    );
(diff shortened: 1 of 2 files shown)


atabel (approved):
have you tried this in a page with multiple spinners?

marcoskolodny:
> have you tried this in a page with multiple spinners?

@atabel yes, it seems fine :)

src/__tests__/spinner-test.tsx
```diff
@@ -0,0 +1,31 @@
...
+    const spinner = screen.getByRole('progressbar', {name: 'Cargando'});
+
+    expect(spinner).toBeInTheDocument();
+    expect(spinner).toHaveAttribute('aria-busy');
+});
+
+test('spinner with role presentation is accessible', () => {
+    render(
+        <ThemeContextProvider theme={makeTheme()}>
+            <Spinner rolePresentation />
```
pladaria, line 23: as commented in Teams, lets deprecate this prop

marcoskolodny: done

PR #1276, opened 2024-10-22
feat(PhoneNumberFieldLite): Phone number field with simple formatting to reduce bundle size

pladaria:
WEB-2081


src/phone-number-field-lite.tsx
```diff
@@ -0,0 +1,186 @@
...
+            }
+            return formatPhoneLite(regionCode, number);
+        },
+        [formatFromProps, regionCode]
+    );
+
+    const rifm = useRifm({
+        format,
+        value: controlledValue,
+        accept: /[\d\+]+/g,
```
pladaria, line 115: @marcoskolodny the issues we saw in the custom formatter example and the `+` char were caused because it wasn't added in the `accept` regex

src/__screenshot_tests__/input-fields-screenshot-test.tsx
```diff
@@ -521,6 +521,28 @@ test.each`
     expect(await fieldWrapper.screenshot()).toMatchImageSnapshot();
 });
 
+test.only.each`
```
marcoskolodny, line 524: I think you forgot to remove `only` from here

atabel (approved):

src/__screenshot_tests__/input-fields-screenshot-test.tsx
pladaria: ouch!

src/phone-number-field-lite.tsx
marcoskolodny: Good catch!

marcoskolodny (approved):

atabel (approved):

PR #1279, opened 2024-10-23
fix(useDisableBodyScroll): avoid affecting body's height

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-2088)

### DIFF
--- a/src/hooks.tsx
+++ b/src/hooks.tsx
@@ -53,7 +53,6 @@ export const useDisableBodyScroll = (disable: boolean): void => {
                             `top: ${-bodyScrollTop}px;`,
                             'left: 0px;',
                             'right: 0px;',
-                            'bottom: 0px;',
                             'overscroll-behavior-y: contain;', // disable overscroll
                         ].join('');
                 }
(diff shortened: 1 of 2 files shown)


src/hooks.tsx
```diff
@@ -53,7 +53,6 @@ export const useDisableBodyScroll = (disable: boolean): void => {
                             `top: ${-bodyScrollTop}px;`,
                             'left: 0px;',
                             'right: 0px;',
-                            'bottom: 0px;',
```
marcoskolodny, line 56: No need for the `bottom: 0px;`, because:
- If the body's height is smaller than the screen, there is no scroll, so it shouldn't be an issue.
- If there is scroll, the hook already fixes the top position and scrolls to the corresponding vertical position. Setting bottom to 0px doesn't change anything.

The issue is that if we also set bottom to 0, children of the body that have `height:100%` will be affected by this hook (because when top and bottom values are set, the height of the container becomes well defined, and it will trigger changes in the child).

marcoskolodny: I've tried different scenarios, and everything seems to work fine:
- Body that has no scroll
- Body that has scroll
  - Verified case when scroll position is at the top, middle and bottom of the body
- Verified that resizing the window while blocking the scroll doesn't affect anything, even if the body didn't have scroll before the resize and I make the window very small.

atabel (approved):

pladaria (approved):

PR #1281, opened 2024-10-29
feat(Meter): New component

pladaria:
### Milestones

* [x] Angular
* [x] Circular
* [x] Linear
* [x] Automatic width support (for non numeric widths)
* [x] Forward animation
* [x] Reverse animation
* [x] Accessibility
* [x] prefers-reduced-motion support
* [x] Workaround for small segments
* [x] Workaround for zero segment at the end
* [x] Non rounded lines depending on theme
* [x] Tests
* [x] Playroom Snippets

### Implementation details
* Using `requestAnimationFrame` for animation
* Values array in the range [0..100] (like the progress bar)
* Creating a full circle using the `arc` path has issues. When the start end end point are the same, it draws the circle in an unexpected direction. To avoid this effect, a very small threshold value is used to clamp (not noticeable, a small fraction of a px).


src/meter.tsx
yceballost, line 1: With multiple values (stacked) the border is not visible during the animation

You can test it changing value 50 to 10 for example
https://tinyurl.com/2ysu72go

yceballost (changes requested):

src/meter.tsx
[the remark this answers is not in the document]
pladaria: Hmm... This is on purpose. If you have X values in your Meter, the animation shows all, as expected.

if the value, instead of 0 is 0.001. You should get the same result as now

atabel (approved):

src/meter.tsx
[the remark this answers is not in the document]
yceballost: understood!

If I wish to hide this values i can implementate a simple logic to use "undefined" value when is 0. But by default is true that the component is coherent as you did it.

pladaria: This is expected, to better visualize what's happening instead of 0s use values of 1: `[1, 50, 1, 1]` => `[1, 10, 1, 1]`.

The first segment is animated but it goes behind the large one, so it is not visible

src/theme.tsx
```diff
@@ -162,11 +162,12 @@ export type Theme = {
...
     dimensions: {headerMobileHeight: number; headerDesktopHeight: number};
     colorValues: Colors;
+    borderRadii: BorderRadiiConfig;
     textPresets: TextPresetsConfig;
     Link: LinkComponent;
     isDarkMode: boolean;
     isIos: boolean;
     useHrefDecorator: () => (href: string) => string;
-    t: (token: TextToken) => string;
+    t: (token: TextToken, ...params: Array<string | number>) => string;
```
marcoskolodny, line 171: Shouldn't params be optional?

src/meter.tsx
```diff
@@ -0,0 +1,494 @@
...
+const CENTER_Y = VIEW_BOX_WIDTH / 2;
+
+const STROKE_WIDTH_PX = 6;
+const SEPARATION_PX = 2;
+
+const ANIMATION_DELAY_MS = 200;
+const ANIMATION_DURATION_MS = 1000;
+const ANIMATION_EPSILON = 1000 / 60 / ANIMATION_DURATION_MS / 4;
+
+const SMALL_VALUE_THRESHOLD = Math.PI / 10000;
```
marcoskolodny, line 26: This works fine for semi-circles, but in circular meter, this value should be multiplied by 2 (360 degrees = 2*pi)

I don't think this is critical, but if you want the same threshold for both types of meter, you should take it into account

pladaria: ok, not super relevant, I just want a "small enough value". In fact this is a leftover from a previous implementation. At this moment all values are internally represented in the range [0..1], so it would be better to use a 0.0001

src/theme.tsx
pladaria: they are optional. If not passed any additional parameters, the params array will be empty (`[]`)

marcoskolodny (approved):

src/meter.tsx
marcoskolodny: If this is leftovers, maybe you can replace it with 0.0001?

Otherwise, anyone that will read this code in the future may think that using the calculation with math.PI is relevant

pladaria: done

yceballost (approved):
nice job 🥰

src/text-tokens.tsx
yceballost, line 1: we have copies reviewed by ux writing

https://www.figma.com/design/jWWCJ9kYl6I5uHLz3GcgRp/%F0%9F%94%B8-Data-Visualizations-Specs?node-id=2150-1858&t=GDst6sSOy8PdwMxE-4

pladaria: changed!

PR #1292, opened 2024-11-11
feat(Header, Cards, Hero, CoverHero): improve a11y

marcoskolodny:
Issues:
- [WEB-2087](https://jira.tid.es/browse/WEB-2087)
- [WEB-2107](https://jira.tid.es/browse/WEB-2107)


src/utils/types.tsx
```diff
@@ -14,3 +14,10 @@ export type IconProps = {
 export type ByBreakpoint<T> = T | {mobile: T; tablet?: T; desktop: T};
 
 export type HeadingType = 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'span';
+
+export const isBiggerHeading = (heading: HeadingType, otherHeading?: HeadingType): boolean => {
+    if (!otherHeading) {
+        return true;
+    }
+    return heading !== 'span' && (otherHeading === 'span' || heading < otherHeading);
+};
```
atabel, line 23: hmmm... I think this file isn't a good place for this utility function

src/utils/types.tsx
```diff
@@ -14,3 +14,10 @@ export type IconProps = {
 export type ByBreakpoint<T> = T | {mobile: T; tablet?: T; desktop: T};
 
 export type HeadingType = 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'span';
+
+export const isBiggerHeading = (heading: HeadingType, otherHeading?: HeadingType): boolean => {
+    if (!otherHeading) {
+        return true;
+    }
+    return heading !== 'span' && (otherHeading === 'span' || heading < otherHeading);
```
atabel, line 22: this means that if both headings are `'span'`, the second one wins?

atabel: I'd add some early returns to simplify this boolean expression

marcoskolodny: According to design specs, same heading level should never be used.

That being said, I think in case the headings are the same, it's better for the first one to "beat" the second one. I'll change it

marcoskolodny: Done

src/utils/types.tsx
marcoskolodny: Moved to utils/headings.tsx

atabel (approved):

src/community/advanced-data-card.tsx
```diff
@@ -87,44 +90,86 @@ const CardContent = ({
...
-                <Text2
-                    color={vars.colors.textPrimary}
-                    truncate={pretitleLinesMax}
-                    as={pretitleAs}
-                    regular
-                    hyphens="auto"
-                >
-                    {pretitle}
-                </Text2>
+        /** using flex instead of nested Stacks, this way we can rearrange texts so the DOM structure makes more sense for screen reader users */
```
marcoskolodny, line 93: AdvancedDataCard was missing the logic for the custom order of elements that other cards have

src/card.tsx
```diff
@@ -578,7 +638,7 @@ export const MediaCard = React.forwardRef<HTMLDivElement, MediaCardProps>(
                         aria-label={isTouchable ? ariaLabel : undefined}
                     >
                         {isTouchable && <div className={styles.touchableMediaCardOverlay} />}
-                        <div className={styles.mediaCard}>
+                        <div className={styles.mediaCard} aria-hidden={isTouchable}>
```
marcoskolodny, line 641: All cards except PosterCard were missing this, and it's important in order to avoid Safari from reading all the texts inside a link as separate links (same thing we did in rows)

pladaria (approved):

aweell:
When the pretitleAs is modified in the header component (although some cases would be not correct) is happening that when the pretitle is h2 and title is h1, the spacing between elements seems to change?


https://github.com/user-attachments/assets/89cb07e7-7d13-4a74-9206-69be464a5fad

marcoskolodny:
> When the pretitleAs is modified in the header component (although some cases would be not correct) is happening that when the pretitle is h2 and title is h1, the spacing between elements seems to change?

Good catch @aweell ! It should be fixed now :)

aweell (approved):
Nice job!

PR #1294, opened 2024-11-14
feat(InfoSheet): update styles and allow adding a button to the content

marcoskolodny:
Issue: [Link](https://jira.tid.es/browse/WEB-2095)


src/sheet-types.tsx
```diff
@@ -71,7 +74,7 @@ export type SheetType = keyof SheetPropsByType;
 export type SheetResultByType = {
     RADIO_LIST: {action: 'SUBMIT'; selectedId: string} | {action: 'DISMISS'};
     ACTIONS_LIST: {action: 'SUBMIT'; selectedId: string} | {action: 'DISMISS'};
-    INFO: void;
+    INFO: {action: 'DISMISS'};
```
marcoskolodny, line 77: I think it's reasonable to always return 'DISMISS' instead of nothing. If the users doesn't want to do anything with it, they can just ignore the result

src/sheet-native.tsx
```diff
@@ -88,25 +88,36 @@ const showActionsListNativeSheet = (
...
-                type: 'LIST',
-                id: 'list-0',
-                listType: 'INFORMATIVE',
-                autoSubmit: false,
-                selectedIds: [],
-                items,
-            },
-        ],
-    });
+        content: button
```
atabel, line 107: in the tests you and Pedro did in Android webviews, this makes old apps return an error (fix pending by Android team). Could we capture that error here and fallback to showing the sheet without button?

marcoskolodny: Discussed offline. We can't do such thing, because the Android error occurs when pressing the button, not when opening the sheet.

This should be handled by the app itself

atabel (approved):

pladaria (approved):

aweell (approved):
LGTM!

PR #1298, opened 2024-11-26
fix(Header, Cards, Hero): fix spacings logic

marcoskolodny:


src/card.tsx
```diff
@@ -424,10 +424,12 @@ const CardContent = ({
                             </div>
                         )}
                         {headline && (
-                            // assuming that the headline will always be followed by one of: pretitle, title, subtitle, description
```
marcoskolodny, line 427: All cards were assuming this before, except of AdvancedDataCard. In https://github.com/Telefonica/mistica-web/pull/1292, I've done the same for the AdvancedDataCard, and it broke one screenshot test in webapp (they use pretitle and extra prop to render everything else).

I've removed all these unnecesary assumptions from everywhere

src/header.tsx
```diff
@@ -107,7 +109,7 @@ export const Header = ({
                                 )}
                                 {headlineContent}
                                 {pretitleContent && (
-                                    <div style={{paddingBottom: pretitle || description ? 8 : 0, order: -1}}>
+                                    <div style={{paddingBottom: title || description ? 8 : 0, order: -1}}>
```
marcoskolodny, line 112: This was a typo :(

atabel (approved):

PR #1299, opened 2024-11-29
feat(VanillaExtract): Use custom identifier generation

pladaria:
Prefix Vanilla Extract identifier names with Mistica version to avoid name collision in a module federation context


marcoskolodny (approved):

vite.config.js
```diff
@@ -4,11 +4,24 @@ import react from '@vitejs/plugin-react';
...
+ * 1.0.0 => v1
+ */
+const version = 'v' + packageVersion.replace(/(\.0)*$/, '').replace(/\./g, '_');
 
 export default defineConfig({
     plugins: [
         react(),
-        vanillaExtractPlugin(),
+        vanillaExtractPlugin({
+            identifiers: ({hash}) => `${version}_${hash}`,
```
pladaria, line 23: another idea:

Create a map hash=>autoincrementCounter

and the resulting identifiers would be: `v1_2_3_i455` to make it even shorter

atabel: I'm not sure if that approach would work. If vite uses multiple processes to compile files, you'd have collisions

atabel (approved):

PR #1301, opened 2024-12-04
feat: add label property to simple-block

paulo-asilva:
OBVIVO-2814 (Simple Block): the label property was added to the simple block, and some adjustments were also included using the Inline component to align the position in the component.

OBVIVO-2618 (Chip): including the use of the inverse token (vars.colors.neutralLow) along with some validations to add it if inverse is true. There is also a small change in the text and icon on the chip to make it readable when using inverse.


src/community/blocks.tsx
```diff
@@ -84,22 +84,35 @@ interface SimpleBlockProps {
...
+                        })}
+                    >
+                        {image}
+                    </div>
+                    <Text2 regular color={vars.colors.textSecondary}>
+                        {description}
+                    </Text2>
+                </Inline>
+                <div className={styles.rightContent}>
+                    <Text2 regular color={vars.colors.brand}>
```
yceballost, line 112:
```suggestion
                    <Text2 regular color={vars.colors.textBrand}>
```

image

yceballost (changes requested):

PR #1306, opened 2024-12-16
feat(Drawer): new component

pladaria:
https://jira.tid.es/browse/WEB-2113


src/portal.tsx
```diff
@@ -20,29 +20,22 @@ export const Portal = ({children, className}: Props): JSX.Element | null => {
     const [container, setContainer] = React.useState<HTMLDivElement | null>(null);
```
pladaria, line 20: This was adding two nodes and removing one, leaving leftovers in the dom.

src/drawer.css.ts
```diff
@@ -0,0 +1,108 @@
...
+import * as mq from './media-queries.css';
+import {vars} from './skins/skin-contract.css';
+import {sprinkles} from './sprinkles.css';
+
+export const ANIMATION_DURATION_MS = 400; // review
+
+export const container = style({
+    position: 'fixed',
+    display: 'flex',
+    paddingBottom: 'env(safe-area-inset-bottom)',
```
pladaria, line 11: pending to verify inside an iphone

src/__tests__/testid-test.tsx
```diff
@@ -288,3 +289,12 @@ test('Meter test ids', () => {
         },
     ]);
 });
+
+test('Drawer test ids', () => {
+    checkTestIds(<Drawer title="Title" subtitle="Subtitle" description="Description" onClose={() => {}} />, [
+        {
+            componentName: 'Drawer',
+            internalTestIds: ['title', 'subtitle', 'description'],
```
atabel, line 297: are these defined in the spec?

src/drawer.css.ts
```diff
@@ -0,0 +1,112 @@
...
+    sprinkles({
+        paddingTop: 40,
+        display: 'flex',
+        flexDirection: 'column',
+        flexGrow: 1,
+    }),
+    {
+        '@media': {
+            [mq.tabletOrSmaller]: {
+                paddingBottom: 16 - 8,
```
atabel, line 44: what's the meaning of this `- 8`?

src/drawer.tsx
```diff
@@ -0,0 +1,306 @@
...
+
+        const close = React.useCallback(() => {
+            setIsOpen(false);
+            return new Promise((resolve) => {
+                setTimeout(resolve, styles.ANIMATION_DURATION_MS);
+            }).then(onClose);
+        }, [onClose]);
+
+        const dismiss = React.useCallback(() => {
+            return close().then(() => onDismiss?.());
```
atabel, line 87: what's the difference between `onDismiss` and `onClose`? I initially thought that `onDismiss` was called immediately and `onClose` when the animation ends, but it seems both are called when the animation ends :/

src/drawer.tsx
```diff
@@ -0,0 +1,306 @@
...
+                    <div
+                        data-testid="drawerOverlay"
+                        onClick={onDismiss ? dismiss : undefined}
+                        className={classnames(
+                            styles.overlay,
+                            isOpen ? styles.overlayOpen : styles.overlayClosed
+                        )}
+                    />
+                    <div
+                        data-testid="drawerLayout"
```
atabel, line 117: is this testid needed?

src/drawer.tsx
atabel: ok, I see the difference. `onDismiss` is only called when the drawer is closed by a dismiss action (esc key /close button, click outside). But can't we call `onDismiss` immediately without waiting for animation?

atabel (approved):

src/drawer.tsx
pladaria: Yes, we could call it immediately but I decided to not emit the close/dismiss event while the drawer is visible.

src/drawer.css.ts
pladaria: I've added this to avoid adding extra padding to the actions. But this doesn't look good when the link button is not visible. Instead of this, I've added the `bleedY` prop to the LinkButton

src/drawer.tsx
pladaria: Not really. I've added it to differentiate the boundaries of the different parts of this component. The `DrawerLayout` is a separate component as `Drawer`.

src/__tests__/testid-test.tsx
pladaria: Not in spec, but I was assuming the same testids for these elements that we were using in cards, etc

src/drawer.tsx
yceballost, line 1: Opening the width prop we are opening also problems like.. what happened if the setup width is more than the device viewport? Is it possible to avoid this problem?

[Example](https://mistica-1rof8b9a8-flows-projects-65bb050e.vercel.app/playroom/preview/#?code=N4Igxg9gJgpiBcIA8AhArgFwxAdgBQCcBLAWwEMCBPAAl0JgGcGBeYACgEprmA+ahmBgDKGMhhhsAOiAgAHGDgAiBMgHcYBaQBpqGAmhgcAvj0k5q1APLzzytRrNIA9Oix1i5KqZzAA5oJExCWk5BTt1TRAdADMyABsBLgAyJOopc2okcIcMi1UiKAwAC1YAVgAGSqMzCwsMIgw4mGZpABUGpuka2oY0ACN6xubpIX7BzpBui1gGMGJZetwWkEVGOaIFolwu3NocAGE4iAFWTm4+AWFRcSkZG2zImPjE6t2+zGwcVmBdGAAPDDwajSQikCiUbR7ehMIFnXjUYBGahGV61fgwSA4KDg1yfb6-AFAkYY3DYqiQ9yMBiwrjwxHI1G1d5uHAAGSIOAA1vjxITgSB2VyKfgCFSaecEUiUVM9ooiAwSPKWOxaXxEYzvGikIEwJz+LIyGBmsAAIwANhMMosSDwcUNMCKEDisAI1CKMCIviKGFYAGYqtQnJq0ZlbfbHc6NG6PV6fcB-eUkUGraG7UaIy7o57vX6A8ndtaw+mnZn3dm4wmk8HrU4dZzg84Ht5jGYoiBijASHBEABZCAAN3logIICMQA)

The width should never exceed the viewport's width. If the width is greater than the viewport, the drawer should occupy 100% - 40px (to ensure the overlay is displayed).

maybe something like
`max-width: calc(100vw - 40px);`

src/drawer.css.ts
```diff
@@ -0,0 +1,112 @@
+import {style} from '@vanilla-extract/css';
+import * as mq from './media-queries.css';
+import {vars} from './skins/skin-contract.css';
+import {sprinkles} from './sprinkles.css';
+
+export const ANIMATION_DURATION_MS = 400; // review
```
yceballost, line 6:
```suggestion
export const ANIMATION_DURATION_MS = 350;
```

src/drawer.css.ts
```diff
@@ -0,0 +1,112 @@
...
+        right: 0,
+        bottom: 0,
+        overflow: 'hidden',
+    }),
+    {
+        paddingBottom: 'env(safe-area-inset-bottom)',
+        '@media': {
+            [mq.mobile]: {
+                left: 0,
+                transition: `transform ${ANIMATION_DURATION_MS}ms cubic-bezier(0.32, 0.72, 0, 1)`,
```
yceballost, line 23:
```suggestion
                transition: `transform ${ANIMATION_DURATION_MS}ms cubic-bezier(.5,0,.5,1)`,
```

src/drawer.css.ts
```diff
@@ -0,0 +1,112 @@
...
+            },
+            [mq.tabletOrBigger]: {
+                borderTopLeftRadius: vars.borderRadii.container,
+                borderBottomLeftRadius: vars.borderRadii.container,
+                transition: `transform ${ANIMATION_DURATION_MS}ms cubic-bezier(0.65, 0, 0.35, 1)`,
```
yceballost, line 28:
```suggestion
                transition: `transform ${ANIMATION_DURATION_MS}ms cubic-bezier(.5,0,.5,1)`,
```

Tested in local, I feel this curve is more natural and smooth

src/drawer.css.ts
```diff
@@ -0,0 +1,112 @@
...
+        [mq.mobile]: {
+            transform: 'translateY(100%)',
+        },
+        [mq.tabletOrBigger]: {
+            transform: 'translateX(100%)',
+        },
+    },
+});
+
+export const overlay = style([
```
yceballost, line 85: @aweell I think the overlay transition is not needed in mobile. What do you think?

yceballost (changes requested):

src/drawer.css.ts
pladaria: ok

src/drawer.css.ts
pladaria: ok. Please update specs

src/drawer.css.ts
pladaria: I like it. And If the drawer is white over a white background, it adds contrast.

src/drawer.tsx
pladaria: done. We could also add a `min-width`.

Please update specs with this information

src/drawer.tsx
```diff
@@ -0,0 +1,307 @@
...
+import {ButtonLink, ButtonPrimary, ButtonSecondary} from './button';
+import Divider from './divider';
+import {getPrefixedDataAttributes} from './utils/dom';
+import * as tokens from './text-tokens';
+
+import type {DataAttributes, HeadingType, TrackingEvent} from './utils/types';
+
+const PADDING_X_DESKTOP = 40;
+const PADDING_X_MOBILE = 16;
+const PADDING_X_TABLET = 16;
```
aweell, line 26: Update to 32 to match the spacings in tablet, updated in specs also:

https://www.figma.com/design/NfM16IJ4ffPVEdiFbtU0Lu/%F0%9F%94%B8-%5BREADY%5D-Drawer-Specs?node-id=10-5397&t=ODR4pg1TRGAgvsF8-1

src/drawer.tsx
aweell: We have updated specs accordingly and also added the min width = default width in desktop & tablet for now.

If it's a need in the future to have smaller drawers we can define a different min-width

src/drawer.tsx
pladaria: done

yceballost (approved):

aweell (approved):
Nice work!

Just as a nice to have last comment... if we can set the min value of the storybook width slider to the min-width of the drawer would be awesome!

Screenshot 2025-01-08 at 14 26 26

pladaria:
> Nice work!
> 
> Just as a nice to have last comment... if we can set the min value of the storybook width slider to the min-width of the drawer would be awesome!
> 
> Screenshot 2025-01-08 at 14 26 26

set to 300, so the min-width style can be verified

PR #1307, opened 2025-01-02
feat(Form): Disable autofocus on error for some fields on iOS

pladaria:
Fields that are excluded from autofocus on error in iOS because the focus action opens the picker/selector

* date
* datetime-local
* month
* select

Note: scroll to element not affected

A simple unit test that covers these scenarios has been added but you can be manually test this using playroom with a snippet like:
```jsx
<Form>
  <Stack space={16}>
    <Select
      name="field"
      label="field"
      options={[{ value: "x", text: "x" }]}
    />
    <ButtonPrimary submit>Submit</ButtonPrimary>
  </Stack>
</Form>
```
and opening the preview in an iphone

### DIFF
--- a/src/utils/platform.tsx
+++ b/src/utils/platform.tsx
@@ -21,10 +21,19 @@ export const isRunningAcceptanceTest = (platformOverrides: Theme['platformOverri
 
 const isEdgeOrIE = Boolean(typeof self !== 'undefined' && (self as any).MSStream);
 
-export const isAndroid = (platformOverrides: Theme['platformOverrides']): boolean =>
-    getUserAgent(platformOverrides).toLowerCase().includes('android') && !isEdgeOrIE;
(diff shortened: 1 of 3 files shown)


src/utils/platform.tsx
```diff
@@ -21,10 +21,19 @@ export const isRunningAcceptanceTest = (platformOverrides: Theme['platformOverri
...
+        return true;
+    }
+
+    return getUserAgent(platformOverrides).toLowerCase().includes('android') && !isEdgeOrIE;
+};
 
 export const isIos = (platformOverrides: Theme['platformOverrides']): boolean => {
+    if (platformOverrides.platform === 'ios') {
+        return true;
+    }
```
pladaria, line 35: added to easily force the platform in tests

atabel (approved):

yceballost (approved):

PR #1308, opened 2025-01-08
feat: add label property to simple-block

wendellguimaraes-tlf:
OBVIVO-2814 (Simple Block): the label property was added to the simple block, and some adjustments were also included using the Inline component to align the position in the component.

https://www.figma.com/design/hTthBYJeDVzcpUhkQYlyFG/Blocks-specs?node-id=7304-2609&t=GTbvRZkMvRSJacnI-0


wendellguimaraes-tlf:
> Perfeito!
> 
> Só um pequeno detalhe, adicione o adereço `label`aos trechos da sala de jogos
> 
> https://github.com/Telefonica/mistica-web/blob/34bcc78ad0403527f6d9a9105fbf4ec70655df12/playroom/snippets.tsx#L3957-L3970

Done !

src/community/blocks.tsx
```diff
@@ -84,22 +84,35 @@ interface SimpleBlockProps {
...
     'aria-label'?: string;
+    label?: string;
 }
 
-export const SimpleBlock = ({image, description, 'aria-label': ariaLabel}: SimpleBlockProps): JSX.Element => {
+export const SimpleBlock = ({
+    image,
+    description,
+    'aria-label': ariaLabel,
+    label,
```
atabel, line 94: I don't like this prop name. `label` is usually associated to accesibility. For example in form inputs, or for `aria-labels` (which this component already has).
Can we change this prop name to something different (`rightText` for example)?

src/community/__screenshot_tests__/__image_snapshots__/blocks-screenshot-test-tsx-blocks-simple-block-1-snap.png
atabel, line 1: I don't link the right text to be blue (`textBrand`). It looks like it's clickable

src/community/blocks.tsx
wendellguimaraes-tlf: Done !

atabel (approved):

PR #1309, opened 2025-01-14
fix: change inverse token chip

wendellguimaraes-tlf:
OBVIVO-2618 (Chip): including the use of the inverse token (vars.colors.neutralLow) along with some validations to add it if inverse is true. There is also a small change in the text and icon on the chip to make it readable when using inverse.


yceballost (changes requested):
Hi @wendellguimaraes-tlf 

The PR is not aligned with the specs definition :( 

The file that you linked in PR description is not correct. 
https://www.figma.com/design/512a5Ke4EldL0GYffCXGCJ/%F0%9F%94%B8-Chips-Specs?node-id=0-1

This is de official specs for Chips and you can see with a yellow box what parts are pending to be implemented. Please, read the specs and make the changes according to the specs. 

If you have any doubts, please address them before starting work to avoid unnecessary effort or wasted time.

Thanks for your effort with this PR anyways 🙏

PR #1318, opened 2025-02-14
feat(Meter): added extra prop

wyunreal:


src/__stories__/meter-story.tsx
```diff
@@ -100,6 +110,21 @@ export const MeterStory: StoryComponent<MeterStoryArgs> = ({
                         reverse={reverse}
                         values={values}
                         width={fullWidth ? '100%' : width}
+                        {...(extraContent
+                            ? {
+                                  extra: (
+                                      <div
+                                          style={{
```
Winde, line 117: would it make sense to use <Align> of mistica here instead of css?

src/__screenshot_tests__/__image_snapshots__/meter-screenshot-test-tsx-meter-default-angular-100-1-snap.png
wyunreal, line 1: My eyes saw no difference here :)

src/__screenshot_tests__/__image_snapshots__/meter-screenshot-test-tsx-meter-default-circular-33-33-2-snap.png
wyunreal, line 1: I don't know why the screenshot is not taken including the full circle. It seems it is only taking the extra required space into account. I've tested adding 2 components in the story and there is no spacing issue.

src/__stories__/meter-story.tsx
wyunreal: As it is a story inside mistica, I prefer to have stories isolated from other components.

wyunreal: take into account this story enables screenshot tests, I think these tests should not fail when, for example, a box is changed in mistica.

src/meter.tsx
```diff
@@ -281,6 +283,23 @@ const MeterComponent = ({
         ' ' +
         values.map((v, i) => `${t(meterSectionLabel, i + 1, Math.round(v * 100))}`).join('. ');
 
+    const extraStyle = React.useMemo(() => {
+        if (type === TYPE_LINEAR) {
+            return {display: 'flex'};
+        }
+        const COS_45 = 0.707107;
+        const sizeFactor = (width / 2 - STROKE_WIDTH_PX) * COS_45;
```
vdedios, line 291: More accurate:

```suggestion
        const sizeFactor = (width / 2 - STROKE_WIDTH_PX) * Math.cos(Math.PI / 4);
```

vdedios (approved):

pladaria (approved):

src/__stories__/meter-story.tsx
```diff
@@ -100,6 +110,21 @@ export const MeterStory: StoryComponent<MeterStoryArgs> = ({
                         reverse={reverse}
                         values={values}
                         width={fullWidth ? '100%' : width}
+                        {...(extraContent
+                            ? {
+                                  extra: (
```
atabel, line 115: Why not just `extra={extraContent ?  {} : undefined}`

atabel (approved):

src/__stories__/meter-story.tsx
wyunreal: done !!

PR #1326, opened 2025-03-10
feat(Row): add tabIndex prop

ecirstea:
O2DE-7083

Related with accessibility, when expanding a list, the focus should move to the first element of the expanded list and that's why the tabIndex prop is needed.

Locally tested:
[Screen_recording_20250311_084555.webm](https://github.com/user-attachments/assets/c1d569c4-d2ad-4a6e-b72b-63e097e63ce8)


brtbrt:
1. have you tried on webapp if it fits our needs?
2. do the other components in mística have a test for a11y props? could you check it? in case not, we're good to go!

brtbrt (approved):

claramunt (approved):

atabel (approved):

PR #1337, opened 2025-04-01
feat(InputFields & Counter): use inputBorder token in border

yceballost:
> [!Important]  
> All OBs are communicated


src/text-field-base.css.ts
```diff
@@ -61,7 +61,7 @@ const commonInputStyles = style([
     }),
     {
         background: 'none',
-        outline: 0,
```
yceballost, line 64: would be nice have screenshots for focus?

src/pin-field.tsx
aweell, line 1: When using a keyboard the focus is visible in the first element but when writing following digits, the focus disappears leaving only the "visible" active state, I would expect the focus to move towards the active element also

yceballost: yay.. maybe this is out of the scope of this PR, the initial idea was just to use inputBorder token in input fields borderColor 😬 just to avoid blockages in this PR

aweell: I was trying to improve focus visualization in inputs and I think this will be an improvement:

https://mistica-8z6dhhgnr-flows-projects-65bb050e.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8BnALgTwDYwHwB0A7AAmOAAMjSBtKAQzToFpIBbABwkJkLScLqsYAXnwgAKjAAeaAGIBLGFihiAusVzEo8gG5kqpYhB0wATgDMsEAO7xiO%2BSnkAjHAG4DpCAFc0WedxMEObmKDBodgBM7FIeJMQAvgbkSYRESAD06Nh4RHmESABKMCichE4mADJ0GD5oxI4AkoQmpmEE8UgAQhBSxOx0UNqEAOYAmsLAkQAsCR2GxEiSMgpKUMQCQqIgmzBixFh0zkrbAHKCeyDEGfOLGT1SHZnFpVwVMNW1vh0gCUA

I would remove focus from the scope of this PR and create another one to deal with that topic.

aweell (approved):

src/pin-field.tsx
```diff
@@ -198,11 +199,10 @@ const PinInput = ({
                          * style than other inputs (less margin, fontSize and height).
                          */
                         style={{
-                            marginTop: `calc(${styles.pinInputLineHeight} - 1px)`,
-                            marginBottom: `calc(${styles.pinInputLineHeight} - 1px)`,
                             lineHeight: styles.pinInputLineHeight,
                             fontSize: styles.pinInputLineHeight,
-                            height: styles.pinInputLineHeight,
+                            height: `calc(${styles.pinInputLineHeight} * 3 - 2px)`,
```
atabel, line 204: can you explain this calc?

src/text-field-components.css.ts
```diff
@@ -102,13 +102,18 @@ export const fullWidth = sprinkles({
...
-    overflow: 'hidden',
-    border: 'regular',
-    display: 'flex',
-    borderRadius: vars.borderRadii.input,
-    position: 'relative',
-});
+export const field = style([
+    sprinkles({
+        overflow: 'hidden',
+        border: 'regular',
```
atabel, line 108: instead of setting border to regular and overriding the borderColor, I'd just set the border. Alternatively, we could define a new border type for inputs in [sprinkles](https://github.com/Telefonica/mistica-web/blob/4302740e75928802d0bf51e1995925f00eb9ab2e/src/sprinkles.css.ts#L64)

atabel (approved):

src/text-field-components.css.ts
yceballost: I included new type in sprinkles to manage input borders from one side

PR #1340, opened 2025-04-03
feat(Form): allow disabling onBlur validation, make '(optional)' label optional, allow to hide search icon in SearchField

atabel:
WEB-2235

- new `validateOnBlurInsideForm` prop to toggle field validation onBlur. By default this prop is true
- new `showOptionalLabel` prop to toggle the "(optional)" label suffix in optional fields. By default this prop is true
- new `withSearchIcon` prop to toggle the search icon in SearchField. By default this prop is true

Example disabling validation on blur:
https://mistica-rbb21fwi2-flows-projects-65bb050e.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8AxCAnAtgHQHYAJ8JcBlAVwCNMBLAFwF5gAKAMw0wBEBDWrgSnz0AfHkKEuAGxjpaTYKLGFadKfHzYQAFQAW1AM759%2BAJ4Qy6fFB5cNAGgWLMMPXq4BzGGoBSJAPIA5ADo9WnRqXDdqFmNWdm5eW3xcMgkJRIAmPnsCMQBfPgVcvBEcpAAhCAAPfAAHLigocLdGAEYANlySxSQSXjAAa3w9OrAYVo6uxXwkTRhK2hRqGAkoJK4neg1cdZgNfAkuCmXNkH8dvYA3SWorWhhfXDKJcwBJXD0bmDQsRhZJPRguXwAHpJt0AKKYLjUCSLZZQBxTbYbDQwKEwjSIxQHI4SE4wAC06IkmJyU3wVwkNx490ez3Qbw%2BsG%2BmF%2B-0BWJBYLE5TItFoxAAMlxTHzOYQamEoehjGU%2BQLcIxefziAAFKVcGVDSg0WhCEgwXBQJDAuUq3Dq6jS4xCIpksSgxEm3pcAaTE0VSolE0skogWwgWjaNHOBAAbRAAFkIBd9Lx0CAALoBgDuNyDenDAGY2gAGRO5IA


src/__stories__/select-story.tsx
```diff
@@ -15,6 +15,7 @@ type Args = {
     error: boolean;
     native: boolean;
     optional: boolean;
+    showOptionalLabel: boolean;
```
yceballost, line 18: mmm why not hideOptionalLabel? the default state is show the label when you use optional={true}, and using this prop you want to hide it.

However, if you use `<TextField name="name" label="Name" showOptionalLabel={true} />` the label is not shown (it is obviously) but I hope that the prop will show the label with your approach.

src/search-field.tsx
```diff
@@ -16,6 +16,7 @@ export interface SearchFieldProps extends CommonFormFieldProps {
     onChangeValue?: (value: string, rawValue: string) => void;
     getSuggestions?: (value: string) => ReadonlyArray<string>;
     inputMode?: React.HTMLAttributes<HTMLInputElement>['inputMode'];
+    withSearchIcon?: boolean;
```
yceballost, line 19: maybe is more scalable called "hideStartIcon"? This name could be the same in any field in the future if we decide to open this config

yceballost (approved):

carlosgarciaibanez (approved):

src/__stories__/select-story.tsx
aweell: If its required it shouldn't show the label no matter the prop value is IMO so current behaviour is expected.

aweell (approved):

src/__stories__/select-story.tsx
atabel: I think I prefer positive booleans, even if the default value is true.

Of course, if the field isn't optional the label is never shown, but when it is the label is shown unless you set `showOptionalLabel` to false

src/search-field.tsx
atabel: my initial idea was `startIcon` to allow customizing it (allowing null/undefined for removing it). But @aweell told me we don't want that customization.

Following your proposal I think I prefer something like `withStartIcon`. I prefer positive booleans, even if the default value is true

PR #1343, opened 2025-04-10
feat(TextLink): add underline to textLink

yceballost:
https://jira.tid.es/browse/WEB-2151

Specs: https://www.figma.com/design/81rPCqGchGf54aoh2kpF1o/%F0%9F%94%B8-TextLink-Specs?node-id=0-1


src/text-link.tsx
```diff
@@ -27,6 +27,7 @@ interface CommonProps {
     dataAttributes?: DataAttributes;
     /** IMPORTANT: try to avoid using role="link" with onPress and first consider other alternatives like to/href + onNavigate */
     role?: string;
+    bodyLink?: boolean;
```
yceballost, line 30: When links are within text there should be a way to identify them that is not based on color, by default all links will be underlined.

There are other cases where the context already communicates users that is a link and they are not surrounded by text, for example in navigation bars or menus. In these cases, the underline in the default state will be hidden, and they will only show the underline on hover states.

aweell (approved):

src/text-link.tsx
atabel: I think `bodyLink` isn't clear enough. I think I prefer something like:

`underline?: 'always' | 'on hover'` being `'always'` the default.

yceballost: better, yes! changed!

atabel (approved):

PR #1347, opened 2025-04-15
feat(Row,BoxedRow): allow overriding touchable role

wyunreal:
Changes:

- On row and boxed row components, a new prop is added to customize the touchable role.
- if this prop is NOT used, the touchable continues getting the role prop is passed, else, the touchable role is used.
- This prop is not used when the row contains a control.

### DIFF
--- a/src/__tests__/list-test.tsx
+++ b/src/__tests__/list-test.tsx
@@ -81,6 +81,27 @@ test('Row as a button', async () => {
     expect(spy).toHaveBeenCalled();
 });
 
+test('Row, keeping its listitem role, containing a button reporting link role', async () => {
+    const spy = jest.fn();
+    render(
+        <ThemeContextProvider theme={makeTheme()}>
+            <RowList>
+                <Row title="Title" onPress={spy} touchableRole="link" />
+            </RowList>
+        </ThemeContextProvider>
+    );
+
+    const rowDiv = screen.getByRole('listitem');
+    expect(rowDiv).toBeInTheDocument();
+
+    const button = screen.getByRole('link', {name: 'Title'});
+    expect(button).toBeInTheDocument();
+    expect(rowDiv).toContainElement(button);
+
+    await userEvent.click(button);
+    expect(spy).toHaveBeenCalled();
+});
+
 test('Row with switch', async () => {
     const spyOnChange = jest.fn();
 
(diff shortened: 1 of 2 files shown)


src/list.tsx
```diff
@@ -643,12 +646,12 @@ const RowContent = React.forwardRef<TouchableElement, RowContentProps>((props, r
     return (
         <div
             className={classNames(styles.rowContent, styles.rowContentPadding)}
-            role={role}
+            role={touchableRole ?? role}
```
atabel, line 649: why using touchableRole here? this isn't a touchable

src/list.tsx
atabel, line 1: I'd add some unit test to check the roles given different situations:
- default roles
- check item role & touchable role when item role is overriden
- check item role & touchable role when touchableRole is overriden
- check item role & touchable role when both role and touchableRole are overriden
- check roles in some special cases like switch, radio or nontouchable rows

atabel, line 1: I see current implementation sets `'listitem'` as default role for `Row` but not for `BoxedRow`. `BoxedRows` should have the default `'listitem'` role too.

src/list.tsx
```diff
@@ -475,14 +478,14 @@ const RowContent = React.forwardRef<TouchableElement, RowContentProps>((props, r
                     [styles.pointer]: !disabled,
                 })}
                 {...interactiveProps}
-                role={role}
+                role={touchableRole ?? role}
```
atabel, line 481: I'd say this is a bug.
If you override `role` but not `touchableRole`, we shouldn't set both roles to the same value.
I know you did this to preserve old behavior, but I'd consider old behavior a bug

wyunreal: So lets do a major version for this bug to be fixed, no ?

src/list.tsx
wyunreal: you totally right, Me Se fue la pinza XD

src/list.tsx
wyunreal: will do !!

wyunreal: Will do !!

src/list.tsx
```diff
@@ -732,17 +735,20 @@ type BoxedRowProps = ExclusifyUnion<
...
+            overflow="visible"
+            className={styles.boxed}
+            variant={props.isInverse ? 'inverse' : 'default'}
+            ref={ref}
+            role={role}
+            dataAttributes={{'component-name': 'BoxedRow', testid: 'BoxedRow', ...dataAttributes}}
+        >
+            <RowContent {...props} />
+        </InternalBoxed>
+    )
```
wyunreal, line 750: @atabel WDYT ?

src/list.tsx
atabel: I think this change doesn't deserve a major release.
I see this more as a bugfix than as breaking change. The possible breakage cases in product code are quite small and easy to fix if any. The alternative to preserve the old behavior would be generating invalid output when role is overriden but touchableRole isn't, I don't like that

d-sainz (approved):

Winde (approved):

src/list.tsx
wyunreal: I think all is covered now, wdyt @atabel ?
Do you think we need to add another cases ?

src/__tests__/list-test.tsx
```diff
@@ -81,6 +81,27 @@ test('Row as a button', async () => {
...
+            </RowList>
+        </ThemeContextProvider>
+    );
+
+    const rowDiv = screen.getByRole('listitem');
+    expect(rowDiv).toBeInTheDocument();
+
+    const button = screen.getByRole('link', {name: 'Title'});
+    expect(button).toBeInTheDocument();
+    expect(rowDiv).toContainElement(button);
```
atabel, line 99: I'd use `within` here
```suggestion
    const button = within(rowDiv).getByRole('link', {name: 'Title'});
    expect(button).toBeInTheDocument();
```

src/list.tsx
atabel: I think it's enough

atabel (approved):

PR #1351, opened 2025-04-22
feat(mistica-css): card asset

DavidNietoGonzalez:


examples/css/index.html
```diff
@@ -293,6 +293,14 @@ <h3 class="mistica-text-title2">Cards</h3>
                 <h4 class="mistica-text-title1">MediaCard / DataCard / SnapCard</h4>
                 <div class="cards-group">
                     <section class="mistica-card">
+                        <div class="mistica-card__asset-floating">
+                            <div class="mistica-card__asset-icon" style="background: #E6F5FD">
```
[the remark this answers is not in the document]
DavidNietoGonzalez: Not really. It was done this way to maintain the React structure. It gets reduced and adjusted. Two classes are left: one for each type of asset — one floating and the other embedded in the card.

examples/css/index.html
```diff
@@ -293,6 +293,12 @@ <h3 class="mistica-text-title2">Cards</h3>
                 <h4 class="mistica-text-title1">MediaCard / DataCard / SnapCard</h4>
                 <div class="cards-group">
                     <section class="mistica-card">
+                        <div class="mistica-card__asset-floating" style="background: #E6F5FD">
```
yceballost, line 296: why don't you use Mística color tokens?

You could use `brandLow` for circle background

examples/css/index.html
```diff
@@ -293,6 +293,12 @@ <h3 class="mistica-text-title2">Cards</h3>
                 <h4 class="mistica-text-title1">MediaCard / DataCard / SnapCard</h4>
                 <div class="cards-group">
                     <section class="mistica-card">
+                        <div class="mistica-card__asset-floating" style="background: #E6F5FD"> 
+                            <svg width="24" height="24" viewBox="0 0 24 24" role="presentation">
+                                <path 
+                                    fill="var(--_1vqcj1i1u)"
```
yceballost, line 299: just a doubt, could you use `brand` color token here?

css/mistica-common.css
```diff
@@ -817,3 +817,35 @@ button.mistica-display-card:has(.mistica-card__media):active:after {
 .mistica-accordion-item[open]::details-content {
     height: auto;
 }
+
+.mistica-card__asset {
```
yceballost, line 821: Mmm I have doubts about the asset approach. This container accept any type of content inside not only circle stuff. 

Should this class be used for Circle component like `.mistica__circle` ?
should we use border-radius 50% here? i don't think so 🤔

You could include a horizontal logo like
image

same for floating asset

wdyt @atabel ?

yceballost (changes requested):

examples/css/index.html
DavidNietoGonzalez: this is harcoded just for HTML example, on AEM there are a color picker to choose the color.

css/mistica-common.css
DavidNietoGonzalez: ok, this assest is only circle with icon. we can upgrade for another type of asset

examples/css/index.html
yceballost: yay! I know. I would try to avoid hardcoding colors in examples. For instance, these colors won’t work properly in dark mode because they are not using tokens.

To make the examples as compliant as possible.

DavidNietoGonzalez: Then i change it for a var to improve that.

css/mistica-common.css
atabel: If we want to support any content there yes, I'd create a specific class for the cirlcle style (border-radius: 50%) and only apply it conditionally.

DavidNietoGonzalez: update with:
Since it is required to include not only an icon but also images and border radius,

three values can be passed:

border-radius
height
width

This allows us to have:

icon and image positioned absolutely or in their natural position
icon and image with border radius
icon with or without background

yceballost: great 🙌✨

yceballost (approved):

atabel (approved):

aweell (approved):

PR #1352, opened 2025-04-24
feat(mistica-css): Table implementation

atabel:
WEB-2198

Not all the cases are implemented, for example the boxed case.


examples/css/index.html
```diff
@@ -83,6 +83,88 @@ <h3 class="mistica-text-title2">Tag</h3>
...
+                            <td data-heading="Price">3,00€</td>
+                            <td data-heading="Average rating">9/10</td>
+                        </tr>
+                    </tbody>
+                </table>
+                <h4 class="mistica-text-title1">Row headings</h4>
+                <table class="mistica-table mistica-table-mobile-collapsed">
+                    <tbody>
+                        <tr>
+                            <th scope="row"><span class="mistica-text-3">Conectividad</span></th>
```
atabel, line 140: In this example I show a table without thead but with rows heading (th with row scope). In AEM we should allow to configure this.

Also, note how I'm overriding the default font size of cells by nesting a span with the mistica-text-3 class. This kind of customization should be possible in AEM too. Ideally we should allow including some other components inside table cells (like Tags)

css/mistica-common.css
yceballost, line 1: Is it possible to have a config for row verical alignment?

In this page use vertical align top for texts in the first column
https://www.telefonicaempresas.es/moviles-empresas/samsung-galaxy-s25?capacity=256GB

yceballost (changes requested):

css/mistica-common.css
atabel: yes, we could do that in AEM side

yceballost (approved):

DavidNietoGonzalez (approved):

PR #1358, opened 2025-05-05
fix: MOVES-10042 screenreader after animations

javierfdezcast:


src/feedback.tsx
```diff
@@ -385,6 +426,7 @@ export const SuccessFeedback = ({
 }: AssetFeedbackProps): JSX.Element => {
     useHapticFeedback('success');
     const {skinName, platformOverrides} = useTheme();
+    const [isShouldRead, setShouldRead] = React.useState(false);
```
anabelengp, line 429: should the initial state depend on areAnimationsSupported?

src/feedback.tsx
```diff
@@ -78,20 +88,33 @@ const renderFeedbackBody = (
...
+                    tabIndex={-1}
+                >
                     <Text6 as="h1">{title}</Text6>
                 </div>
 
                 {normalizedDescription && (
                     <div
                         className={classnames(animateText && styles.feedbackTextAppearMedium)}
                         data-testid="description"
+                        aria-hidden={hideOthers}
```
anabelengp, line 117: why do these divs need to be hiden?they are also animated

javierfdezcast: they are hidden because if not, the screen reader reads then before the title

PR #1359, opened 2025-05-05
feat(PinField): wrap to next line when digits don't fit in viewport

atabel:
https://jira.tid.es/browse/WEB-2146


src/inline.tsx
atabel, line 1: added support for `verticalSpace` prop in `Inline` component when used with `wrap`

src/pin-field.tsx
```diff
@@ -152,7 +152,7 @@ const PinInput = ({
     };
 
     return (
-        <Inline space={8}>
+        <Inline space={8} wrap verticalSpace={16}>
```
atabel, line 155: This is the only change needed in `PinField`, but needed changes in `Inline` to support this

yceballost (approved):

PR #1360, opened 2025-05-06
A11y: ensure contrast compliance for Blau skin

AnaMontes11:
This PR consolidates the accessibility improvements required to meet contrast ratio guidelines by June 28. Key updates include adjusting the `textNavigationBarPrimary` token to `grey6` to ensure adequate contrast over `blauBluePrimary`. As a result, Blau will no longer employ `inverse contexts` on other components (with the exception of inverse cards, which will now use `blauBlueSecondary` as their background). Additionally, we've reviewed and updated all color tokens to guarantee a minimum 3:1 contrast for non-text elements and 4.5:1 for text in both light and dark modes. These changes ensure the Blau skin delivers an inclusive and readable experience across themes.


src/skins/blau.tsx
yceballost, line 1: inputBorder token is missing in this file

yceballost, line 1: el warning no pasa accesibilidad y por tanto el rating no pasa accesibilidad, deberíamos usar uno que minimo pase el 3:1

AnaMontes11: token included

AnaMontes11: Added blauYellow65 to ensure this contrast

aweell:
Inactive tag text in dark mode is not passing

aweell:
Cvv illustrations are failing 3:1 (2.34:1), we can discuss if we consider them decorative. I don't since they describe the user where to look in their credit card for the information we're requesting.

Screenshot 2025-05-07 at 3 55 18 PM

src/skins/blau.tsx
```diff
@@ -107,29 +109,29 @@ export const getBlauSkin: GetKnownSkin = () => {
             textLinkDanger: palette.blauRed,
             textLinkSnackbar: palette.blauPurple30,
             textActivated: palette.blauBlueSecondary,
-            textBrand: palette.blauBluePrimary,
-            inputBorder: palette.grey5,
```
yceballost, line 111: why this change? 🤔

AnaMontes11: You're still seeing the old version with grey5 — I hadn’t committed that yet. But you’ll see it now in this one.

AnaMontes11:
> Cvv illustrations are failing 3:1 (2.34:1) in light mode, we can discuss if we consider them decorative. I don't since they describe the user where to look in their credit card for the information we're requesting.
> 
> Screenshot 2025-05-07 at 3 55 18 PM

I agree — in my opinion, those specific icons should use brandHigh instead of brand, but we should make that change in: Resources/cvvTooltip.

AnaMontes11:
> Inactive tag text in dark mode is not passing
> 
> Screenshot 2025-05-07 at 3 50 08 PM

Yes, true — we have the same issue in Vivo, Tú, and Telefónica (the latter is already fixed in my WIP branch). For this change, I had to adjust the background (darkModeGrey6), since the text was still inaccessible with grey4, and lowering it to grey3 made it too light.

src/skins/blau.tsx
```diff
@@ -108,28 +110,28 @@ export const getBlauSkin: GetKnownSkin = () => {
...
+            inputBorder: palette.grey4,
+            control: palette.grey4,
             controlActivated: palette.blauBlueSecondary,
             controlInverse: palette.blauBluePrimary30,
             controlActivatedInverse: palette.white,
             controlError: palette.blauRed,
-            barTrack: palette.grey2,
-            barTrackInverse: applyAlpha(palette.grey6, 0.2),
+            barTrack: palette.grey3,
+            barTrackInverse: applyAlpha(palette.white, 0.5),
```
yceballost, line 120: this color works worse than grey6, why did you make this change?

AnaMontes11: To maintain parity with the other brands, I thought this worked better in my opinion—but I’ll roll it back.

PR #1370, opened 2025-05-13
feat(mistica-css): TextField / SearchField

atabel:
WEB-2194


examples/css/index.html
```diff
@@ -83,6 +83,30 @@ <h3 class="mistica-text-title2">Tag</h3>
                     <div class="mistica-tag-error">Error</div>
                 </div>
 
+                <h3 class="mistica-text-title2">TextFields</h3>
+                <h4 class="mistica-text-title1">TextField</h4>
+                <div class="vertical-group">
+                    <div class="mistica-text-field">
+                        <input id="text-field-1" type="text" placeholder=" " />
```
atabel, line 90: note that `placeholder` is always needed. Without a placeholder we can't implement the shrinking label effect when the input is filled. With placeholder we can use the `:placeholder-visible` pseudoclass to do the trick. When we don't want the input to have a visible placeholder, we can set it to `" "` as I'm doing in this example

examples/css/index.html
```diff
@@ -83,6 +83,30 @@ <h3 class="mistica-text-title2">Tag</h3>
...
+                        <label for="text-field-1">Label</label>
+                    </div>
+                    <div class="mistica-text-field">
+                        <input id="text-field-1" type="text" placeholder="Some placeholder" />
+                        <label for="text-field-1">Label</label>
```
yceballost, line 95: Long labels is bugged. Can we limit to 1 line? (I know this limitation is not good for accessibility but another solution is increase input field height to fit all the label. This is something that we need to improve in react too, a pending task.

image

examples/css/index.html
```diff
@@ -9,7 +9,7 @@
     </head>
     <body data-mistica-skin="movistar" data-mistica-color-scheme="light">
         <div id="app" class="mistica-responsive-layout">
-            <h1 class="mistica-text-title4">Hello mistica css</h1>
+            <h1 class="mistica-text-title4">Mistica css</h1>
```
yceballost, line 12:
```suggestion
            <h1 class="mistica-text-title4">Mística CSS</h1>
```

Just a stupid thing xD but always use "tick" to "Mística" name

and CSS i think should be in uppercase because is an acronym

yceballost (changes requested):

examples/css/index.html
atabel: should be fixed

yceballost: I don't think so 😬
image
https://mistica-m2fh9sls4-flows-projects-65bb050e.vercel.app/mistica-css

atabel: hmmm... this is how I see it 
image
which browser are you using?

DavidNietoGonzalez (approved):

examples/css/index.html
atabel: I've reproduced it in Firefox. I've made some changes to fix it. Please check again

yceballost (approved):

PR #1372, opened 2025-05-16
feat(mistica-css): counter

atabel:
WEB-2197

Not implemented the delete button case. Do we need this case for AEM?


examples/css/mistica-counter.js
atabel, line 1: For this component to work we need some JS
In AEM side we can copy/paste this snippet

examples/css/index.html
```diff
@@ -82,7 +82,27 @@ <h3 class="mistica-text-title2">Tag</h3>
                     <div class="mistica-tag-warning">Warning</div>
                     <div class="mistica-tag-error">Error</div>
                 </div>
-
+                <h3 class="mistica-text-title2">Counter</h3>
+                <div class="mistica-counter">
+                    <button class="mistica-counter__button" aria-label="decrease" disabled>
```
atabel, line 87: in AEM, we should allow configuring aria-label for the input and increase/decrease buttons

DavidNietoGonzalez (approved):

aweell:
> Not implemented the delete button case. Do we need this case for AEM?

Doesn't seem we need that case, the only usage i spotted was: https://www.movistar.es/fibra-y-movil/autonomos/

examples/css/index.html
```diff
@@ -82,7 +82,27 @@ <h3 class="mistica-text-title2">Tag</h3>
...
+                        <svg width="20" height="20" viewBox="0 0 24 24" aria-hidden>
+                            <path
+                                fill="currentColor"
+                                d="M3.001 12.047a.88.88 0 0 1 .828-.924h16.34a.876.876 0 0 1 .003 1.752H3.829A.876.876 0 0 1 3 12.047"
+                            />
+                        </svg>
+                    </button>
+                    <input type="number" aria-label="counter" value="0" min="0" max="10" step="1" />
```
aweell, line 95: Is possible to write in the input, is something we can block from our side?

aweell (changes requested):
Screenshot 2025-05-16 at 4 42 36 PM

The value is not visible in dark mode

css/mistica-common.css
```diff
@@ -1267,3 +1267,78 @@ button.mistica-display-card:has(.mistica-card__media):active:after {
...
+
+/* Counter */
+
+.mistica-counter {
+    border-radius: var(--mistica-border-radius-button);
+    background: var(--mistica-color-backgroundContainer);
+    display: inline-flex;
+    gap: 8px;
+    overflow: hidden;
+    border: 1px solid var(--mistica-color-inputBorder);
```
aweell, line 1279: In inverse context the border should be white also

Screenshot 2025-05-16 at 4 44 02 PM

examples/css/index.html
atabel: We can change the input to readonly (I've changed it).
But I think it would be better and more accesible if the user can edit it using the keyboard too

examples/css/mistica-counter.js
yceballost, line 1: voice over is not working as same as react version. For example, when you increase the value, the value does not read, or in the react version doesn't receive focus in the number only in +/- controls

yceballost (changes requested):

examples/css/mistica-counter.js
atabel: made some changes, check it now

css/mistica-common.css
```diff
@@ -44,6 +44,8 @@
     --vcolor-checkbox-tick: var(--mistica-color-controlKnobInverse);
     --vcolor-checkbox-background: transparent;
 
+    --vcolor-inputBorder: var(--mistica-color-backgroundContainer);
```
aweell, line 47: Seems that the PR https://github.com/Telefonica/mistica-web/pull/1337 missed the border when the counter is inverse and in dark mode, since the border is visible when is a default context and in dark mode.

Both should look the same, the correct one is the current default context with the visible border.

atabel: should be fixed

aweell (approved):

yceballost (approved):

PR #1374, opened 2025-05-20
feat(mistica-css): slider

atabel:
https://jira.tid.es/browse/WEB-2196


css/mistica-common.css
yceballost, line 1: inverse and dark mode is broken
image
image

yceballost (changes requested):

css/mistica-common.css
atabel: should be fixed

PR #1384, opened 2025-06-05
feat(mistica-css): padding, stack, inline, align

atabel:
WEB-2254

### DIFF
--- a/examples/css/index.html
+++ b/examples/css/index.html
@@ -46,7 +46,7 @@
                 <h2 class="mistica-text-title3">Components</h2>
 
                 <h3 class="mistica-text-title2">Texts / Titles</h3>
-                <div class="horizontal-group">
+                <div class="mistica-inline mistica-gap-16">
                     <div>
                         <p class="mistica-text-1">Text 1</p>
                         <p class="mistica-text-2">Text 2</p>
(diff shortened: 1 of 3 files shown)


examples/css/index.html
```diff
@@ -935,6 +935,103 @@ <h4 class="mistica-text-title1">3 + 9</h4>
                         <div class="placeholder" style="height: 120px; width: auto"></div>
                     </div>
                 </section>
+                <h3 class="mistica-text-title2">Padding</h3>
+                <div class="mistica-padding-8" data-demo-container>
```
yceballost, line 939: why not call it "box"? Suppose because padding is more easy to understand.. but just to know why this misalign with react lib

examples/css/index.html
```diff
@@ -935,6 +935,103 @@ <h4 class="mistica-text-title1">3 + 9</h4>
...
+                </div>
+
+                <h3 class="mistica-text-title2">Stack</h3>
+                <div class="mistica-stack mistica-gap-16" data-demo-container>
+                    <div class="placeholder" style="height: 50px"></div>
+                    <div class="placeholder" style="height: 50px"></div>
+                    <div class="placeholder" style="height: 50px"></div>
+                </div>
+
+                <h3 class="mistica-text-title2">Inline</h3>
```
yceballost, line 1029: include inline wrap too maybe? I think it is usefull

yceballost (approved):

examples/css/index.html
atabel: because we need padding-x, padding-y padding-top, padding-bottom, etc. How do you do that with box? remember we don't have props in css like with react components

PR #1385, opened 2025-06-05
fix(ProgressBar, ProgressBarStepped): fix screen reader anouncement

atabel:
WEB-2236

### DIFF
--- a/src/progress-bar.tsx
+++ b/src/progress-bar.tsx
@@ -36,10 +36,9 @@ export const ProgressBar = ({
     const progressValue = Math.max(0, Math.min(100, progressPercent));
 
     const getFormattedValueText = () => {
-        const name = ariaLabel || texts.loading || t(tokens.loading);
         const completedText = texts.progressBarCompletedLabel || t(tokens.progressBarCompletedLabel);
 
-        return `${name}, ${progressValue.toFixed(1)}% ${completedText}`;
+        return `${progressValue.toFixed(1)}% ${completedText}`;
     };
 
     const a11yProps =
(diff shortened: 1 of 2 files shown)


src/progress-bar.tsx
```diff
@@ -50,7 +50,7 @@ export const ProgressBar = ({
                   'aria-valuenow': progressValue,
```
atabel, line 50: The ticket description suggested removing `aria-valuenow`, but removing it makes the progressbar indeterminate for screen readers, so it must be there.

When `aria-valuenow` is set, the screen reader reads it after the aria-label, resulting in a repeated read because the `aria-label` already includes the value. The best way to avoid this is using `aria-valuetext` instead of `aria-label`

src/progress-bar.tsx
```diff
@@ -35,11 +35,11 @@ export const ProgressBar = ({
...
 
-    const getFormattedLabel = () => {
-        const ariaLabelText = ariaLabel || texts.loading || t(tokens.loading);
-        const completedLabelText = texts.progressBarCompletedLabel || t(tokens.progressBarCompletedLabel);
+    const getFormattedValueText = () => {
+        const name = ariaLabel || texts.loading || t(tokens.loading);
+        const completedText = texts.progressBarCompletedLabel || t(tokens.progressBarCompletedLabel);
 
-        return `${ariaLabelText}, ${progressValue}% ${completedLabelText}`;
+        return `${name}, ${progressValue.toFixed(1)}% ${completedText}`;
```
atabel, line 42: I've rounded the value to 1 decimal to avoid screen reader reading "16.66666666666" for 1/6, for example.

claramunt (approved):

PR #1394, opened 2025-06-12
fix(Stepper): improve a11y

atabel:
WEB-2166

https://www.figma.com/design/YIQIeK0R4bQNNkw2Friv6b/%F0%9F%94%B8-Stepper-Component-Specs?m=auto&node-id=1876-781&t=JtsFc0x2F8qis1w2-1

https://www.w3.org/WAI/tutorials/forms/multi-page/#using-step-by-step-indicator

https://github.com/user-attachments/assets/30be0ace-e1f3-4392-9071-991b6b29819b

### DIFF
--- a/src/text-tokens.tsx
+++ b/src/text-tokens.tsx
@@ -373,10 +373,10 @@ export const progressBarStepLabel: TextToken = {
 };
 
 export const stepperCurrentStep: TextToken = {
-    es: 'Actual',
+    es: 'En curso',
     en: 'Current',
     de: 'Aktuell',
-    pt: 'Atual',
+    pt: 'Em andamento',
 };
 
 export const stepperCompletedStep: TextToken = {
--- a/src/text-tokens.tsx
+++ b/src/text-tokens.tsx
@@ -43,6 +43,8 @@ export type Dictionary = {
     sheetConfirmButton: string;
     progressBarCompletedLabel: string;
     progressBarStepLabel: string;
+    stepperCurrentStep: string;
+    stepperCompletedStep: string;
     pinFieldInputLabel: string;
     counterRemoveLabel: string;
     counterIncreaseLabel: string;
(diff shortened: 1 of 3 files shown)


src/text-tokens.tsx
```diff
@@ -370,6 +372,20 @@ export const progressBarStepLabel: TextToken = {
...
+    pt: 'Atual',
+};
+
+export const stepperCompletedStep: TextToken = {
+    es: 'Completado',
+    en: 'Completed',
+    de: 'Abgeschlossen',
+    pt: 'Concluído',
+};
+
```
atabel, line 388: official translations pending

pladaria (approved):

src/text-tokens.tsx
aweell: Ready: https://jira.tid.es/browse/PRODUCTDSN-3290

atabel: thanks, updated

aweell:
It is possible to adjust the container that shows the focus indicator to the element? I know this only affects a small amout of users that still use screen reader without complete blindness, but we have another ticket related to this topic in lists:  https://jira.tid.es/browse/OBVIVO-2898

(Just to prevent future tickets)

atabel:
> It is possible to adjust the container that shows the focus indicator to the element? I know this only affects a small amout of users that still use screen reader without complete blindness, but we have another ticket related to this topic in lists: https://jira.tid.es/browse/OBVIVO-2898
> 
> (Just to prevent future tickets)

Unfortunately, no :(

This is caused by `ScreenReaderOnly`

> It is possible to adjust the container that shows the focus indicator to the element? I know this only affects a small amout of users that still use screen reader without complete blindness, but we have another ticket related to this topic in lists: https://jira.tid.es/browse/OBVIVO-2898
> 
> (Just to prevent future tickets)

It's a bit tricky because VoiceOver/TalkBack seem to highlight the "invisible" ScreenReaderOnly text. The best I could do is reduce the font-size of the "invisible" text. This is the result:


https://github.com/user-attachments/assets/0ff1160d-a446-4918-b085-49084ee44da2

aweell:
> It's a bit tricky because VoiceOver/TalkBack seem to highlight the "invisible" ScreenReaderOnly text. The best I could do is reduce the font-size of the "invisible" text. 

This is good enough

aweell (approved):

PR #1395, opened 2025-06-12
fix(Spinner): improve aria-live a11y

atabel:
WEB-2245

https://github.com/user-attachments/assets/dfdf30c3-3b2e-44bc-aae4-33a6adab28e2

https://github.com/user-attachments/assets/df9b4740-424b-46c7-8378-0cc5a6c9559c

### DIFF
--- a/src/spinner.tsx
+++ b/src/spinner.tsx
@@ -98,6 +98,8 @@ const Spinner = ({
                 </ScreenReaderOnly>
                 <svg
                     role="progressbar"
+                    // this doesn't work in TalkBack, but it's needed for testing-library tests
+                    aria-label={label}
                     className={styles.spinnerDefault}
                     height={size}
                     style={{...style}}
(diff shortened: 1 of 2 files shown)


src/spinner.tsx
```diff
@@ -30,23 +31,20 @@ const Spinner = ({
...
     const content =
         getPlatform(platformOverrides) === 'ios' ? (
             <svg
-                aria-labelledby={spinnerId}
                 className={styles.spinnerIos}
                 height={size}
                 style={{...style}}
                 role="progressbar"
                 aria-live="polite"
-                aria-busy
```
atabel, line 44: this doesn't make sense here and makes voiceover fail reading the progressbar. `aria-busy` is to be used in the aria that will load the content when the loading finished, not in the spinner/progressbar

pladaria (approved):

PR #1402, opened 2025-06-25
chore(Playroom): set lang attribute in playroom previews

pladaria:
WEB-2267

Requested by VIVO

### Example
image


yceballost (approved):
perfect! 🤩

PR #1403, opened 2025-06-26
feat(mistica-css): support for inverse cards, fixes in switch and padding components.

DavidNietoGonzalez:
Fixes on components.


css/mistica-common.css
DavidNietoGonzalez, line 1: Added card Inverted

css/mistica-common.css
```diff
@@ -117,6 +120,14 @@
     --vcolor-rowChevron: var(--mistica-color-neutralMedium);
 }
 
+.mistica-display-card.mistica-card-inverse,
+.mistica-snap-card.mistica-card-inverse,
+.mistica-card.mistica-card-inverse {
+    --mistica-color-backgroundContainer: var(--mistica-color-backgroundContainerBrand);
```
atabel, line 126: when an inverse card is over an inverse context (an inverse responsive layout for example) the background color should be `backgroundContainerBrandOverInverse`, but I think could ignore this case for the moment. But at least I'd write a TODO comment here

atabel (approved):

css/mistica-common.css
DavidNietoGonzalez: TODO added

PR #1405, opened 2025-06-27
fix(Buttons, Chips, Tags, Sheets, Snackbar): Remove label truncation

yceballost:
> [!NOTE] 
> Already communicated to the OB


src/__screenshot_tests__/__image_snapshots__/table-screenshot-test-tsx-table-mobile-boxed-1-snap.png
aweell, line 1: I get why other screenshots of tables change since the tags now can be higher, but this one?

aweell (approved):

PR #1417, opened 2025-08-06
feat(UnorderedList/OrderedList/ListItem): new components

atabel:
WEB-2143


src/__stories__/ul-story.tsx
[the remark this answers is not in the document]
atabel: I can move them there if you prefer. At first I thought it could be confusing, but no strong opinion

src/index.tsx
```diff
@@ -54,6 +54,7 @@ export type {TagType} from './tag';
 export {Placeholder} from './placeholder';
 export {Title1, Title2, Title3, Title4} from './title';
 export {RowList, Row, BoxedRowList, BoxedRow} from './list';
+export {Ul, Ol, Li} from './ul';
```
pladaria, line 57: Not super fan of this naming. I would prefer something like:
```
<List>
<List ordered>
<ListItem>
```

yceballost: Same here. I like Pedro's proposal or something like <OrderedList> / <UnorderedList>

atabel: I used these names to avoid confusion with the lists we already have

pladaria (approved):

src/index.tsx
atabel: Renamed to `OrderedList` / `UnorderedeList` / `ListItem`

src/__stories__/ul-story.tsx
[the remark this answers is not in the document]
atabel: moved there

PR #1426, opened 2025-08-27
fix(Card): Naked border radius. Default slotAlignment for deprecated cards

pladaria:
* Remove NakedCard border radius when no media is set
* Set default `slotAlignment` for deprecated cards


src/card-media.tsx
```diff
@@ -117,21 +117,36 @@ type DeprecatedHighlightedCardProps = Omit<MediaCardProps, 'size' | 'mediaPositi
...
+>(
+    (
+        {
+            dataAttributes,
+            isInverse,
+            variant,
+            imageUrl,
+            imageSrc,
+            mediaWidth,
+            slotAlignment = 'bottom',
```
pladaria, line 129: this is the relevant change, rest y just automatic code styling

PR #1427, opened 2025-09-03
feat(Carousel): improve control buttons aria-label

atabel:
WEB-2288

### DIFF
--- a/src/carousel.tsx
+++ b/src/carousel.tsx
@@ -296,17 +296,15 @@ export const CarouselPageControls = ({
     const variant = useThemeVariant();
     const prevPageNumberText =
         prevArrowEnabled && pagesCount !== undefined && currentPageIndex !== undefined
-            ? ', ' +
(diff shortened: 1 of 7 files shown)


src/carousel.tsx
```diff
@@ -287,14 +289,30 @@ export const CarouselPageControls = ({
     setShouldAutoplay,
     prevArrowEnabled,
     nextArrowEnabled,
+    pagesCount,
+    currentPageIndex,
 }: CarouselPageControlsProps): JSX.Element => {
     const {texts, t} = useTheme();
     const variant = useThemeVariant();
+    const prevPageNumberText =
```
pladaria, line 297: Perhaps we could add support to the translate function to accept a string: `token: Token | string`.

Inside the translate function:
```js
const text = typeof token === 'string' ? token : token[language] || token.en
```

This way we could write this code as:
```js
t(text.patata || tokens.patata, param1, param2)
```

PR #1428, opened 2025-09-03
feat(VerticalMosaic): improve a11y. Fix reading order and set list role as default

atabel:
WEB-2287


aweell:
When testing with voiceOver in:

https://mistica-73k1x54ip-flows-projects-65bb050e.vercel.app/?path=/story/components-mosaic--vertical-mosaic-story&args=items:5

The screen reader announces the list as two items, first listItem is mosaic elements from 1-4 and second listItem is element 5

pladaria (approved):

PR #1435, opened 2025-09-18
fix(TextFields): focus ring in text fields with prefix and/or end icon

atabel:
https://jira.tid.es/browse/WEB-2316

Before vs after (I've run the test before applying the fix):
input-fields-screenshot-test-tsx-text-field-prefix-and-end-icon-focus-ring-1-diff

This is how it looks in Chrome/Linux:
chrome

Firefox/Linux:
firefox

Safari/Mac:
image


aweell:
Reviewing this i find myself in a dilemma, in case where the endIcon is an interactive element, like for example the CVV field, is a good practice to nest focus indicators when the right part of the input is a complete different interaction? 

Can we know when right icon is interactive to avoid this focus ring nesting?

[Right now both appear focused when end icon is targeted](https://mistica-h6t8hjimr-flows-projects-65bb050e.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8ARAhgFxgMQJYwBsoACAOzQFsYBeAHRCkxnuILQCNC6R0sWB6AHy1SSAMIA3CXkIlyVbmCks2nAtzEA1TQOGkQAXyA) and I'm afraid that may confuse users

I've seen the webapp PR screenshots and this change seems an improvement, if there is a way to avoid the netire focus when two interactive elements share the same input area would be nice ([thinking also in the evolution of the date field](https://github.com/Telefonica/mistica-design/issues/2324))

Screenshot 2025-09-19 at 12 57 02

aweell (approved):
❤️

PR #1436, opened 2025-09-19
chore(icons): add categories in icons

yceballost:


src/icons/__stories__/mistica-icons-story.tsx
```diff
@@ -148,8 +168,9 @@ export const Catalog: StoryComponent<Args> = ({
 };
 
 Catalog.args = {
+    category: 'All',
     filter: '',
-    size: 32,
+    size: 24,
```
yceballost, line 173: I considered this is the correct default because the common size that you will see icons is in 24px in the ui

pladaria (approved):

PR #1475, opened 2026-01-07
chore(Storybook): Enable the Docs tab

KsAkira10:
ref: #1474 

Problem

image

The Docs tab disappeared and it is very useful for seeing the example implementation

Fix

image

I also added some decorators as a suggestion to improve the readability of the example code


Marcosld:
Closing while we internally find a solution

PR #1519, opened 2026-03-13
feat(Icon Proposal): New Icon component for FBR CDN usage

joaoelias1921:
This PR brings a new component proposal for Mistica, which aims to optimize how the library handles SVG icon rendering and bundle size.

Summary of changes:
- New Icon component in src, fetches icons from FBR's running CDN, optimized with cache and fetching only when in user viewport;
- Catalog story in Storybook updated to render using the new Icon component instead of local generated icon files, with optimized rendering;
- If approved, this implementation will allow us to remove 2000+ generated icon files, reducing overall library bundle size (except icons that are used internally by other components, for now);


ieduardogf:
We understand the issue you’re facing with the bundle size of Mística in the context of Module Federation, especially due to the impact of the icons.

That said, we believe that any solution that changes how Mística exposes or consumes icons needs to be carefully evaluated, as it would have an impact on all projects currently using Mística, many of which do not have this problem.

As discussed with the team, one possible direction would be to decouple the library into two parts: Mística Components and Mística Icons. This would allow you to continue consuming the components as usual, while giving you the flexibility to implement a custom strategy for icons on your side (e.g. loading via CDN/API, SSR, or any approach that better fits your architecture).

However, even with this separation, it is not guaranteed that the issue would be fully resolved. Module Federation behavior (especially regarding tree-shaking and bundle size) depends heavily on the specific configuration being used. Have you been able to evaluate whether changes on your side—such as how Mística is packaged or consumed—could mitigate the problem?

Regarding the current PR, in its current form we don’t think it can be accepted. It introduces a fundamental change in how icons are consumed, effectively adding a runtime dependency on an external CDN (in this case, a Brazil-specific system). This has several implications:
	•	It introduces a network dependency for rendering UI elements that are currently static.
	•	It may impact performance and user experience (e.g. delayed icon rendering, layout shifts, or failures if the CDN is unavailable).
	•	It couples the design system to an infrastructure that is not shared or validated across all markets.

Overall, this represents a significant architectural change that would affect all consumers, not just the specific use case you’re trying to solve.

Finally, we would need some time on our side to properly evaluate possible solutions and next steps.

joaoelias1921:
@ieduardogf thank you and the team for reviewing the proposal and replying. We will discuss the matter further internally, and look for different solutions.

PR #1528, opened 2026-04-08
feat(menu): add MenuItem action support for onPress, href, and to

MurilloLeoni:
Adds an `action` prop to MenuItem.

Supports:
- onPress for click actions
- href for external links
- to for internal navigation

Jira: https://jira.tid.es/browse/OBVIVO-3412


yceballost:
Hi @MurilloLeoni, is this PR ready to review? Please, include some reviewers like @Marcosld and me

Please, include in PR body the link of the Jira issue 

Thanks!! 🤟

playroom/snippets.tsx
```diff
@@ -27,22 +27,22 @@ const menuSnippet = {
       renderMenu={({ ref, className }) => (
         <div ref={ref} className={className}>
           <MenuSection>
-            <MenuItem label="option 1" onPress={() => {}} />
+            <MenuItem label="option 1" action={{onPress: () => {}}} />
             <MenuItem
               label="option 2"
-              onPress={() => setState("option 2", !getState("option 2", false))}
+              action={{onPress: () => setState("option 2", !getState("option 2", false))}}
```
yceballost: I'm not a engineer, but.. why did you use a different approach like a buttons? 

In buttons we have a diferent props for each action like: 

```
<ButtonPrimary onPress={() => {}}>Action</ButtonPrimary>

<ButtonPrimary href="www.google.es">Action</ButtonPrimary>

<ButtonPrimary to="/path">Action</ButtonPrimary>
```

src/menu.tsx
```diff
@@ -51,21 +51,24 @@ const getItemIndexInMenu = (menu: HTMLElement | null, item: HTMLElement | null):
...
 };
 
+export type MenuItemAction = {onPress: () => void} | {href: string} | {to: string};
+
 interface MenuItemBaseProps {
     label: string;
     Icon?: (props: IconProps) => JSX.Element;
     destructive?: boolean;
     disabled?: boolean;
-    onPress: (item: number) => void;
```
yceballost, line 60: this change is breaking :( we can't do this

src/menu.tsx
```diff
@@ -51,21 +51,24 @@ const getItemIndexInMenu = (menu: HTMLElement | null, item: HTMLElement | null):
     return itemIndex < 0 ? null : itemIndex;
 };
 
+export type MenuItemAction = {onPress: () => void} | {href: string} | {to: string};
```
yceballost: Sorry, I'm not engineer but, why are you following a different approach than other components? (like buttons)

In buttons we have a diferent props for each action like: 

```
<ButtonPrimary onPress={() => {}}>Action</ButtonPrimary>

<ButtonPrimary href="www.google.es">Action</ButtonPrimary>

<ButtonPrimary to="/path">Action</ButtonPrimary>
```

IMO, this approach is not consistent with the rest of the component. We would like feature parity. 

wdyt @Marcosld

yceballost (changes requested):

src/menu.tsx
Marcosld: Yes, take a look at button component because it would be nice to also implement some additional props for to and href type menu items:
```typescript
interface ToButtonProps extends CommonProps {
    to: string | Location;
    newTab?: boolean;
    fullPageOnWebView?: boolean;
    onNavigate?: () => void | Promise<void>;
}
interface OnPressButtonProps extends CommonProps {
    onPress: (event: React.MouseEvent<HTMLElement>) => void | undefined | Promise<void>;
}
interface HrefButtonProps extends CommonProps {
    href: string;
    newTab?: boolean;
    loadOnTop?: boolean;
    onNavigate?: () => void | Promise<void>;
}
```

MurilloLeoni:
Hi! Thank you very much for the feedback and for taking the time to review this PR. 🙂

I’ve added the Jira ticket link to the PR description as requested. For context, the ticket was later cancelled since the development was handled directly by our team, but this is the related one:
https://jira.tid.es/browse/OBVIVO-3412

Regarding the implementation, I initially introduced an `action` prop to support `href`, `to`, and `onPress` in `MenuItem`. Based on your feedback, I understand that this approach is not aligned with the current API patterns and feature parity across the library.

I’m currently updating the `Menu` component to follow the suggested approach.

src/menu.tsx
```diff
@@ -100,8 +125,8 @@ export const MenuItem = ({
                 name={label}
                 checked={checked}
                 onChange={() => {
-                    if (isMenuOpen && itemIndex !== null) {
-                        onPress(itemIndex);
+                    if (isMenuOpen) {
+                        onPress?.(itemIndex ?? 0);
```
Marcosld: Why did you change this?

src/menu.tsx
```diff
@@ -126,12 +151,66 @@ export const MenuItem = ({
...
+                        {Icon && (
+                            <div className={styles.iconContainer}>
+                                <Icon size={24} color={contentColor} />
+                            </div>
+                        )}
+                        <Text3 regular color={contentColor}>
+                            {label}
+                        </Text3>
+                    </div>
+                </Box>
```
Marcosld, line 187: Perhaps you can extract this duplicated code

src/menu.tsx
MurilloLeoni: Good question! This change was introduced while adapting the component to support href and to.
I realized that using itemIndex ?? 0 could lead to unintended behavior when itemIndex is null, so I’ve updated it to only call onPress when itemIndex is valid, keeping the expected behavior.

yceballost (approved):
nice!

Marcosld (approved):

MurilloLeoni:
Hi, sorry for the delay on my side. We had some high-priority tasks in our team and needed to focus on those deliveries, so this one ended up being paused for a bit.

About this PR: I saw that the changes were already approved, but the pipeline is still failing. I checked the errors and they seem to be related to pipeline configuration (e.g., missing token in checkout, accessibility audit failing in CI, and Azure upload issue), not directly to the code changes.

Could you please clarify how the merge process works in this case? Should I take any action to fix these pipeline issues, or is this something handled on your side?

Thanks!

yceballost:
> Hi, sorry for the delay on my side. We had some high-priority tasks in our team and needed to focus on those deliveries, so this one ended up being paused for a bit.
> 
> About this PR: I saw that the changes were already approved, but the pipeline is still failing. I checked the errors and they seem to be related to pipeline configuration (e.g., missing token in checkout, accessibility audit failing in CI, and Azure upload issue), not directly to the code changes.
> 
> Could you please clarify how the merge process works in this case? Should I take any action to fix these pipeline issues, or is this something handled on your side?
> 
> Thanks!

I think the problems of the actions is not by your fault. I think the PR can be merge when @Marcosld can do that for the next release

PR #1529, opened 2026-04-09
fix(TextField): fix screen reader reading order for optional input fields

Marcosld:
When a field is marked as optional, the `<label>` element contained two separate child spans (label text + optional suffix). Some screen readers navigated these as individual stops, causing the field to be announced in two steps instead of one.

Fix by wrapping both spans in a `role="text"` container, which makes VoiceOver read the whole content as a single text block.

Fixes WEB-2387


ieduardogf (approved):

ieduardogf (approved):

atabel (approved):

src/text-field-components.tsx
```diff
@@ -83,20 +83,25 @@ export const Label = ({
...
-                    &nbsp;(
-                    {texts.formFieldOptionalLabelSuffix || t(tokens.formFieldOptionalLabelSuffix)})
-                </span>
-            ) : null}
+            {/* eslint-disable-next-line jsx-a11y/aria-role -- role="text" makes VoiceOver read the whole div as a single text block. */}
+            <span role="text" className={styles.labelInner}>
+                <span className={styles.labelText}>{children}</span>
+                {showOptional ? (
+                    <span>
+                        &nbsp;(
```
yceballost, line 100: why not using styles for this separation? just curiosity

yceballost (approved):

src/text-field-components.tsx
Marcosld: What you mean?

aweell (approved):
LGTM!

PR #1530, opened 2026-04-09
feat(chore): Minimal github conventions

Marcosld:
ref: WEB-2430

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -26,9 +26,5 @@ React component library for Telefonica's Design System.
 
 ## GitHub conventions
 
-- PR title follows **conventional commits** format; the scope must be the affected **component name** (e.g.
-  `fix(TextField): ...`).
-- PR description: concise summary of the problem and fix, ending with `Ref: <TICKET-ID>` (if there's a jira
-  ticket).
-- Always add reviewers `aweell`, `atabel`, `yceballost` and `Marcosld` to every PR.
+- Read [CONTRIBUTING](./CONTRIBUTING.md) before contributing to Mistica repo.
 - Always add the `AI` label to PRs where the code was written by an AI agent.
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -21,6 +21,13 @@ Don't hesitate to ask any questions and share your ideas
 We would love to accept your Pull Requests but please, before starting your development,
 [create an issue](https://github.com/Telefonica/mistica-web/issues/new/choose).
 
+- PR title must follow **conventional commits** format; the scope must be the affected **component name**
```
yceballost, line 24: I would change the way of this is explained. Something like:

PR title must follow **conventional commits** format:
- **feat(componentName)**: Include or remove features in the library.
	- Example: feat(TextField): Add maxLength prop support
- **fix(componentName)**: Correct a bug or unintended behavior in a component.
	- Example: fix(TextField): Resolve focus loss on re-render
- **chore(componentName)**: Internal changes that don't affect component behavior (config, tooling, dependencies). If no component was affected, use Chore as the scope.
	- Example: chore(Storybook): Update Storybook to v8

yceballost (approved):

AGENTS.md
```diff
@@ -21,3 +21,10 @@ React component library for Telefonica's Design System.
 - If you create a new component always create a snippet for playroom.
 - If you add props to a component always update stories accordingly and add the new props to args and
   argTypes.
+
+---
+
+## GitHub conventions
+
+- Read [CONTRIBUTING](./CONTRIBUTING.md) before contributing to Mistica repo.
+- Always add the `AI` label to PRs where the code was written by an AI agent.
```
yceballost, line 30: Most PRs will be created by an AI agent, right? xDD or do you mean when the PR is fully generated by AI with no human intervention at all?

Marcosld: Yeah, I meant when the code is written by AI

CONTRIBUTING.md
Marcosld: I'll just add a pointer to the allowed types, I think it should be enough

atabel (approved):

PR #1532, opened 2026-04-16
feat(fixedFooterLayout): include footerScrollEffect prop

yceballost:
https://jira.tid.es/browse/WEB-2432

https://www.figma.com/design/ncQ4GRI3ZEjEtNW6pyROX2/%F0%9F%94%B8-Layouts-Specs?node-id=2717-20348&t=hAPavuZ2V8Ucmp6W-4


Marcosld (approved):

src/fixed-footer-layout.tsx
```diff
@@ -43,6 +43,13 @@ type Props = {
     containerBgColor?: string;
     children: React.ReactNode;
     onChangeFooterHeight?: (heightInPx: number) => void;
+    /**
+     * When the scroll effect is enabled, an elevation will be displayed on the footer when the user scrolls up and there is still more content to scroll. This effect is only applied on mobile devices, as in desktop the footer is not fixed.
+     */
+    footerScrollEffect?: boolean;
+    /**
+     * this property determines whether or not the effect is displayed when content passes beneath the fixedFooterLayout
+     */
```
Marcosld: why did you split the comment in two?

Marcosld: I think we can be much more concise:
```
@default true
When enabled an elevation will be displayed on the footer if it is fixed and there is scrollable content below it. Notice this layout is not fixed on Desktop.
```

yceballost: sh*t, I think copilot just autocomplete the above one 😬 didn't see. Updated

PR #1538, opened 2026-04-21
chore: add Evalite for optional LLM and task evals

ieduardogf:
## Summary

Adds [Evalite](https://v1.evalite.dev/guides/quickstart/) (`evalite@beta`) and Vitest as dev dependencies so we can run scored evaluations (tasks, prompts, or LLM outputs) with a local UI and trace storage.

This PR **introduces the plumbing only** (dependencies, scripts, sample eval, README). **Follow-up PRs are expected to add new eval suites** by dropping additional `*.eval.ts` files under `evals/` (or extending existing ones), without further tooling changes.

## How to use

1. **Watch mode (development)** — `yarn eval:dev`  
   Runs Evalite in watch mode, re-runs when eval files change, and serves the results UI (default [http://localhost:3006](http://localhost:3006) per Evalite docs).

2. **Single run (e.g. CI or quick check)** — `yarn eval`  
   Runs all evals once and exits with a non-zero status if scores are below the threshold.

3. **Authoring** — Add or edit files matching `evals/**/*.eval.ts`. See `evals/my-eval.eval.ts` for a minimal example using `evalite()` and deterministic scorers.

4. **Scorers** — For string-only scorers (e.g. `exactMatch`), import from `evalite/scorers/deterministic`. The main `evalite/scorers` entry pulls in LLM-judge scorers that need the optional `ai` peer dependency.

## Docs

- README **Development** section lists the new scripts and links the Evalite quickstart.

Ref: N/A


evals/my-eval.eval.ts
```diff
@@ -0,0 +1,19 @@
+import {evalite} from 'evalite';
+import {exactMatch} from 'evalite/scorers/deterministic';
+
+evalite('My Eval', {
```
Marcosld, line 4: I thought we were going to test this lib against its competitors before pushing a installation+configuration to the repo. I mean, there are other options that could potentially be better, perhaps we should test them before commiting (?)

tsconfig.production.json
```diff
@@ -82,6 +82,7 @@
         "playroom",
         "src/test-utils/**",
         "setup-test-env.tsx",
-        "examples"
+        "examples",
+        "evals"
     ]
```
Marcosld, line 87: You should probably exclude evals folder from the published mistica code too

package.json
```diff
@@ -137,7 +140,8 @@
         "typescript": "^5.9.3",
         "vite": "^7.3.1",
         "vite-plugin-entry-shaking": "^0.5.1",
-        "vite-plugin-no-bundle": "^4.0.0"
+        "vite-plugin-no-bundle": "^4.0.0",
+        "vitest": "^4.1.5"
```
atabel, line 144: we may not need vitest, can't we run evalite with jest_

evals/my-eval.eval.ts
atabel, line 1: maybe we should follow a file convention more consistent with the one we use for tests/stories (-eval.ts instead of .eval.ts)

evals/my-eval.eval.ts
atabel: Yes I think we should test different tools

package.json
```diff
@@ -42,7 +42,9 @@
         "audit-accessibility": "node .github/actions/audit-accessibility/main.js",
         "circular-dependencies": "yarn dpdm src/index.tsx --tree=false --warning=false --exit-code circular:1",
         "prepare": "husky install",
-        "edit-playroom-patch": "playroom/patch-playroom.sh"
+        "edit-playroom-patch": "playroom/patch-playroom.sh",
+        "eval:dev": "evalite watch",
+        "eval": "evalite"
```
atabel, line 47: I think we only need one eval script, If you want to run it with watch you can just run `yarn eval watch`

package.json
```diff
@@ -105,6 +107,7 @@
         "eslint": "^8.57.0",
         "eslint-plugin-mistica-local-rules": "workspace:*",
         "eslint-plugin-storybook": "^10.2.8",
+        "evalite": "beta",
```
atabel, line 110: there isn't an stable release yet?

Marcosld: I got confused by this too. They have a stable version but it is kinda stale and they are working actively in their beta 1.0 version. anyways the beta is quite old too (2 months)

ieduardogf:
I going to close this PR and we will back later on with a more robust solution

PR #1539, opened 2026-04-21
fix(ssr-tests): skip hydration visual mismatch for image test

ieduardogf:
## Summary

- Pass `checkHidrationVisualMismatch: false` to `openSSRPage` in the image SSR acceptance test.
- The Image component renders a skeleton overlay during SSR that is cleared on hydration, producing a race-conditioned visual diff between the pre- and post-hydration screenshots captured by `checkHydrationMismatch`. The existing `expect(await page.screenshot()).toMatchImageSnapshot()` assertion still validates the rendered output.
- Fixes the intermittent CI failure seen e.g. on [run 24743453218](https://github.com/Telefonica/mistica-web/actions/runs/24743453218) and [run 24709816665](https://github.com/Telefonica/mistica-web/actions/runs/24709816665): `Expected image to match or be a close match to snapshot but was 4.17% different (20026 differing pixels)`.

### Why the mismatch happens

- [src/image.tsx](src/image.tsx) renders a skeleton `<div>` overlay while `hideLoadingFallback` is false.
- On SSR, `isRunningAcceptanceTest()` returns false (it explicitly short-circuits when `process.env.SSR_TEST` is set — see [src/utils/platform.tsx:20](src/utils/platform.tsx:20)), so the skeleton is present in the SSR markup.
- After hydration the skeleton is cleared and the image becomes visible, so the pre- and post-hydration frames always differ — `checkHydrationMismatch` catches this as a test failure whenever the timing aligns.

## Test plan

- [ ] CI `build` job (`yarn test-ssr --ci`) passes consistently across reruns
- [ ] Image SSR acceptance test still produces the post-hydration snapshot and catches regressions in the final rendered output

### DIFF
--- a/src/__acceptance_tests__/image-ssr-acceptance-test.tsx
+++ b/src/__acceptance_tests__/image-ssr-acceptance-test.tsx
@@ -1,7 +1,14 @@
 import {openSSRPage} from '../test-utils';
 
 test('ssr image', async () => {
-    const page = await openSSRPage({name: 'image', waitUntil: 'networkidle0'});
+    // The Image component renders a skeleton overlay during SSR that is cleared on hydration,
+    // producing an unavoidable visual diff between the pre- and post-hydration frames. The final
+    // screenshot assertion still validates the rendered output.
+    const page = await openSSRPage({
+        name: 'image',
+        waitUntil: 'networkidle0',
+        checkHidrationVisualMismatch: false,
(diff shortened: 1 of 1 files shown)


ieduardogf:
Already fixes on this [pR](https://github.com/Telefonica/mistica-web/pull/1536)

PR #1540, opened 2026-04-23
ACCOUNT-30901 add support for tracks in video component

andyindahouse:


src/video.tsx
```diff
@@ -103,6 +111,8 @@ export type VideoProps = {
     onPause?: () => void;
     onLoad?: () => void;
     poster?: string;
+    /** track elements for subtitles, captions, etc. */
+    tracks?: ReadonlyArray<VideoTrack>;
```
Marcosld, line 116: Perhaps we should point in this comment to the mdn official doc. Also I don't know if there's already a typing for srcLang that fits this use case, if not we can add the spec: https://developer.mozilla.org/en-US/docs/Glossary/BCP_47_language_tag maybe.

src/video.tsx
```diff
@@ -72,6 +72,14 @@ type VideoSourceWithType = {
     type?: string; // video/webm, video/mp4...
 };
 
+export type VideoTrack = {
+    src: string;
+    kind?: 'subtitles' | 'captions' | 'descriptions' | 'chapters' | 'metadata';
```
Marcosld: I think we only need subtitles for this component. The difference between captions and subtitles won't be noticed by the user, and the other types don't look like they are useful for this component.

src/__stories__/video-story.tsx
```diff
@@ -76,7 +78,22 @@ export const Default: StoryComponent<Args> = ({
         dataAttributes: {testid: 'video'},
     };
 
-    const video = <Video src={!emptySource ? VIDEO_SRC : ''} {...props} ref={videoRef} />;
+    const video = (
+        <Video
+            src={!emptySource ? VIDEO_SRC : ''}
+            {...props}
+            ref={videoRef}
+            tracks={[
```
Marcosld, line 100: Perhaps we could add a storybook control for the track?

andyindahouse: I thought about it but I'm not sure if we have something to configure here xD

Marcosld: I mean so you can enable/disable subtitles?

Marcosld (approved):

src/__stories__/video-story.tsx
andyindahouse: True, I've added a new control to manage subtitles visibility

src/video.tsx
yceballost, line 1: I think this change should be include in the specs files before coding to follow the process :)

So, let me ask whats kind of things we can define here in terms of design, position of the subtitles? I can see the subtitles to near to the bottom border 😬

Can we define the style? or this is depends on the browser / SO? (I prefer if this is something supported natively, but just checking)

src/video.tsx
aweell: I'm not sure about this, if the purpose is to provide a transcription with a translation in a different language subtitles are the good fit, but if the purpose is to provide a transcription of the video dialog and other sounds or users with hearing problems or with the video muted, captions should be used. Maybe we should expose all kind options and component user decides based on context?

Marcosld: At most I would expose subtitles and captions... but I don't know if we really need support for captions. I would skip the other 3 unless we specify their behaviour (the last 2 are not automatically handled):
image
And descriptions is not supported in any browser:
image

src/video.tsx
Marcosld: I think we can change some styles (but not spacing): https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/::cue.

Another option is to hide the html default subtitles and implement our own.

andyindahouse: From the AppBuilder side, our idea was to add support for these tracks directly in the video component using the browser’s default styles. Since we identified some limitations when trying to customize them, we decided to open a discussion with the Design Core team to evaluate the small customization options that could be added later.

As Marcos mentioned, the position can already be modified directly in the code, and this would be the most relevant change for now. Could we move forward with this approach and discuss additional customizations later? We need this ASAP.

src/__stories__/video-story.tsx
```diff
@@ -60,8 +64,20 @@ export const Default: StoryComponent<Args> = ({
     poster,
     emptySource,
 }) => {
+    const [enSubtitlesVisible, setEnSubtitlesVisible] = React.useState(true);
+    const [esSubtitlesVisible, setEsSubtitlesVisible] = React.useState(false);
     const videoRef = React.useRef<VideoElement>(null);
 
+    const toggleSubtitles = (index: number) => {
```
Marcosld, line 71: Will you add controls for switching subtitle languages?

andyindahouse: Hmmm I think this should be done by the component's consumer with "setTracks" control (in the same way that it's done in the story), that logic depends on the order of tracks in the tracks array and the custom UX/UI desired.

aweell (approved):

src/video.tsx
```diff
@@ -116,13 +127,20 @@ export interface VideoElement extends HTMLDivElement {
     /** Stops the video and shows the poster image (if available) */
     stop: () => void;
     setCurrentTime: (time: number) => void;
+    /**
+     * Sets the display mode of a track by its index.
+     * - 'showing': track is visible
+     * - 'disabled': track is inactive
+     */
+    setTrackMode: (index: number, mode: 'showing' | 'disabled') => void;
```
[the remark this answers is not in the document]
andyindahouse: nope

src/video.tsx
```diff
@@ -210,6 +228,7 @@ const Video = React.forwardRef<VideoElement, VideoProps>(
                 disableRemotePlayback
                 muted={muted}
                 loop={loop}
+                crossOrigin={tracks?.length ? 'anonymous' : undefined}
```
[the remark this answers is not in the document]
andyindahouse: no needed

src/__stories__/video-story.tsx
```diff
@@ -1,6 +1,8 @@
 import * as React from 'react';
 import {Stack, ButtonPrimary, Inline, Video, Text3, Title3} from '..';
 import beachVideo from './videos/beach.mp4';
+import subtitlesExampleEn from './videos/test-subtitles-en.vtt';
+import subtitlesExampleEs from './videos/test-subtitles-es.vtt';
 import beachImg from './images/beach.jpg';
```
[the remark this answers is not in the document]
andyindahouse: Done

src/video.tsx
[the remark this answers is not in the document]
andyindahouse: no needed

src/video.tsx
```diff
@@ -116,13 +126,20 @@ export interface VideoElement extends HTMLDivElement {
     /** Stops the video and shows the poster image (if available) */
     stop: () => void;
     setCurrentTime: (time: number) => void;
+    /**
+     * Sets the display mode of a track by its index.
+     * - 'showing': track is visible
+     * - 'disabled': track is inactive
+     */
+    setTrackMode: (index: number, mode: 'showing' | 'disabled') => void;
```
Marcosld, line 134: Is the goal of this method changing the subtitles that are currently enabled? Perhaps we should expose a simpler API, something like:
```tsx
setActiveSubtitleIndex: (i: number) => void
```tsx

PR #1546, opened 2026-05-06
fix(Inline): allow children to expand

AlexandraGallipoliRodrigues:
ticket:  https://jira.tid.es/browse/WEB-2439
figma: https://www.figma.com/design/HA6UnnWC187IL1YujMYlYZ/Main-Navigation-Bar?node-id=2412-5583

- `Inline expand` prop: allows you to specify which Inline child should occupy the remaining space.

[bug](https://mistica-gvej7sa6o-flows-projects-65bb050e.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8BZAhgSwHYDk0DcMBzNAFwwiwCE0AnAHSwAInSIAHAZQBsJSawAayK0IAVyxQAwhF60AvMEhyAzgDoARmiEjxknDBLl8MAL6MW3OkRgXWHHn0V2WSACoALGAFsYANToMNCxSJnxA4NJ5ehAsQzIMExiAPhcWJiQqCAAPJnY0KChsIgBNRQAOU1TmdNqkAEFuYmZs6PAYEJgGEGravoyASSwmuKYVfLAYRQAmABZTJjQmoiwB0h8VNsnO7t7%2BvvcYbNJppl8isW9kqgAZeqkAaSYAMQAlAYARepKkAHo3I4nPb7DK-IYjGDAg6-RrNKGuX5ZbJQv6eHz%2BCIhPbmGoqGBgciUTbAADaaWAaRY5FI3Bg8CYMU4pDopBiABpKUxKAAFWgwFQqekACgAlEx5MkxjBSEyyDAhTFsLBsuymAAGEUcmosUxa2oU7Xpam0%2BkxepgSASVkgPV9Hl8gXCsUSqUy5nrBUgJVHVUARk1aV15M5xrpDJAAFFsuw5LYbZz7fzBUxReLJXi3XLPd6VTamNMA4ag4aDf1Q6aQJwxOx2BBaNbbbVE46U8709LZR7FZIfXmAMyF2rFlgAXRxLDxtIJMCgQ2Vihsma7Xp7ubZ6pF46YtGIHiipfSSHB2BgTAAZmJuNwAOoYKCkDxjCZTYC%2BgBsCyWzTWGy2HXWuyckgnAwHQYAeM8GAwNwUDnpeN53g%2BTBYGgvhtHiYEeDETBWBo0FtCBmHYb88IZHghBGBQ1B0OahLMM2xKpi6wCmJ%2BO5oAAtLh%2BExBGIS0HQKScq49ThMytBjBgABeL4YCoHz8oIbDsAA8rQVDEDYEkAPxMH2pz0nMCwkcJTDAHJCkqEpHBqRpRBaUwABkjkpqZrgAscTAAO51lAVB8mggiKGeSx4gsHgAJ7sF4WCbDE3jBGISxCYaIJ8aQAkMKl0IeaQpEsJuQG-ORxAJJQNC0LRVEomCwwntijAmVgNogA%2B6IqAgJIgH4iQQAA%2BgYXkgCObIgF5iEeB18Akr6ACsvrTGOQA):
image

fix, [local playroom](http://localhost:9000/#?code=N4Igxg9gJgpiBcIA8BZAhgSwHYDk0DcMBzNAFwwiwCE0AnAHSwAInSIAHAZQBsJSawAayK0IAVyxQAwhF60AvMEhyAzgDoARmiEjxknDBLl8MAL6MW3OkRgXWHHn0V2WSACoALGAFsYANToMNCxSJnxA4NJ5ehAsQzIMExiAPhcWJiQqCAAPJnY0KChsIgBNRQAOU1TmdNqkAEFuYmZs6PAYEJgGEGravoyASSwmuKYVfLAYRQAmABZTJjQmoiwB0h8VNsnO7t7%2BvvcYbNJppl8isW9kqgAZeqkAaSYAMQAlAYARepKkAHo3I4nPb7DK-IYjGDAg6-RrNKGuX5ZbJQv6eHz%2BCIhPbmGoqGBgciUTbAADaaWAaRY5FI3Bg8CYMU4pDopBiABpKUxKAAFWgwFQqekACgAlEx5MkxjBSEyyDAhTFsLBsuymAAGEUcmosUxa2oU7Xpam0%2BkxepgSASVkgPV9Hl8gXCsUSqUy5nrBUgJVHVUARk1aV15M5xrpDJAAFFsuw5LYbZz7fzBUxReLJXi3XLPd6VTamNMA4ag4aDf1Q6aQJwxOx2BBaNbbbVE46U8709LZR7FZIfXmAMyF2rFlgAXRxLDxtIJMCgQ2Vihsma7Xp7ubZ6pF46YtGIHiipfSSHB2BgYwmU2AvoAbAsls01hsth11t0mEd8pJFGqqpykJwYHQYAeM8GAwNwUBMAAZmI3DcAA6hgUCkB4TBYGgvhtHigEeDETBWBoYFtP%2B2G4b81S-nghBGBQ1B0OahLMM2xKpi6wCmLeO5oAAtPhhExBGIS0HQKScq49ThMytBjBgABeF4YCoHz8oIbDsAA8rQVDEDYUkAPxMH2pz0nMCxkaJTDAApSkqCpHAaVpRA6UwABkznhgJpBCd0W51L8lHEAklA0LQ9E0SiYLDCe2KMGZWA2iAyHoioCAkiAfiJBAAD6BgAO4gCObIgDliHIcl8Akr6ACsvrTGOQA):
image


src/inline.tsx
yceballost, line 1: I dont feel really easy in terms of devExp to have another component called InlineItem to manage this kind of things

Is not there any other solution? maybe manage this property in the `<Inline>` component?

wdyt @atabel ?

AlexandraGallipoliRodrigues: Good point, thanks! I’ve removed the public InlineItem component and moved the behavior into Inline via a growItems prop

src/navigation-bar.css.ts
```diff
@@ -313,9 +313,16 @@ export const navigationBarContentRightExpanded = style([
     }),
     {
         paddingLeft: 136,
+        minWidth: 0,
+        width: 0,
     },
 ]);
 
+export const navigationBarContentRightExpandedContent = style({
+    width: '100%',
```
Marcosld: What about an `expand` prop on the `Inline` element? It would set its width to 100% whenever it is set. This way we wouldn't have to modify navigation bar I think. This prop could accept what you currently have in flexGrow prop and wouldn't need `fullWidth` to be set.

AlexandraGallipoliRodrigues: like [this](https://github.com/Telefonica/mistica-web/pull/1546/changes/c0b9c88fdc039eba4da201537d6413da7ffb559f)?

src/__stories__/main-navigation-bar-story.tsx
```diff
@@ -60,10 +63,34 @@ export const Default: StoryComponent<Args> = ({
     topSlotBackgroundColor,
     wide,
     paddingX,
+    expandedRightSlot,
 }) => {
     const [selectedIndex, setSelectedIndex] = React.useState(0);
     const {isDesktopOrBigger} = useScreenSize();
 
+    const right = expandedRightSlot ? (
+        <Inline fullWidth space={16} alignItems="center" expand={0}>
```
Marcosld: I think we can skip fullWidth prop here, right?

src/inline.css.ts
```diff
@@ -98,6 +98,16 @@ export const stringSpaceWithWrap = style({
...
 });
 
+export const expand = style({
+    width: '100%',
+    display: 'flex',
+});
+
+export const expandItem = style({
+    flex: 1,
+    minWidth: 0,
```
Marcosld: why the minWidth?

src/inline.tsx
```diff
@@ -93,28 +115,41 @@ const Inline = ({
...
             aria-label={ariaLabel}
             aria-labelledby={ariaLabel ? undefined : ariaLabelledBy}
             {...getPrefixedDataAttributes(dataAttributes, 'Inline')}
         >
-            {React.Children.map(children, (child) =>
-                !!child || child === 0 ? (
+            {childrenArray.map((child, index) => {
+                return (
                     <div
+                        key={getChildKey(child, index)}
```
Marcosld: I would use React.Children.map because key logic is handled automatically:

Image

src/inline.tsx
```diff
@@ -93,28 +115,41 @@ const Inline = ({
     fullWidth,
     wrap,
     dataAttributes,
+    expand,
 }: Props): JSX.Element => {
     const {platformOverrides} = useTheme();
     const isStringSpace = typeof space === 'string';
+    const childrenArray = React.Children.toArray(children).filter((child) => !!child || child === 0);
```
Marcosld: According to the doc react filters empty elements: 

Image

Marcosld (approved):

src/inline.css.ts
AlexandraGallipoliRodrigues: `minWidth: 0` was intentional here. In a flex layout, flex items have `min-width: auto` by default, so an expanded item can refuse to shrink below its content size and push adjacent items out of the available space. Setting `minWidth: 0` allows the expanded item to shrink properly and lets inner components handle truncation/overflow... But i haven't been able to prove this case in a playroom, should I delete the minWidth then?

Marcosld: Hmm I don't know if that behaviour is desirable, perhaps we could live without it for the moment... 
https://developer.mozilla.org/en-US/play?id=pEK8gpiRaKCWYKNxPpXPkgOoAEQy8NkIAdjrAPjKL0gP%2BAzWTL%2FLoG9rA2WXLAksvUN8leTXOzoc46Dg

src/inline.css.ts
```diff
@@ -98,6 +98,15 @@ export const stringSpaceWithWrap = style({
     justifyContent: space,
 });
 
+export const expand = style({
+    width: '100%',
+    display: 'flex',
+});
+
+export const expandItem = style({
+    flex: 1,
```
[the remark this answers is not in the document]
AlexandraGallipoliRodrigues: we won't change this as we talked [here](https://github.com/Telefonica/mistica-web/pull/1546#discussion_r3232544354)

yceballost (approved):

src/__stories__/main-navigation-bar-story.tsx
```diff
@@ -171,4 +187,7 @@ Default.argTypes = {
         control: {type: 'select'},
         if: {arg: 'wide'},
     },
+    expandedRightSlot: {
+        control: {type: 'boolean'},
+    },
```
atabel: I think this isn't needed, it should be the default

src/inline.tsx
```diff
@@ -93,28 +110,40 @@ const Inline = ({
     fullWidth,
     wrap,
     dataAttributes,
+    expand,
 }: Props): JSX.Element => {
     const {platformOverrides} = useTheme();
     const isStringSpace = typeof space === 'string';
+    const childrenArray = React.Children.toArray(children).filter((child) => child !== '');
```
atabel: more idiomatic
```suggestion
    const childrenArray = React.Children.toArray(children).filter(Boolean);
```

src/inline.tsx
```diff
@@ -93,28 +110,40 @@ const Inline = ({
...
             {...getPrefixedDataAttributes(dataAttributes, 'Inline')}
         >
-            {React.Children.map(children, (child) =>
-                !!child || child === 0 ? (
+            {React.Children.map(childrenArray, (child, index) => {
+                return (
                     <div
                         role={role === 'list' ? 'listitem' : undefined}
+                        className={classnames(
+                            shouldExpand && shouldExpandItem(expand, index) && styles.expandItem
```
atabel: isn't the `shouldExpand` redundant?
```suggestion
                            shouldExpandItem(expand, index) && styles.expandItem
```

atabel (approved):

PR #1549, opened 2026-05-20
fix(Sheet): make close button clickable on iOS

Marcosld:
## Summary
- Move the Sheet dismiss (close) button outside the scrollable `<section>` so it is not clipped by the parent's `overflowY: auto`.
- On iOS, the close button was rendered outside the visible scroll area and therefore not clickable. Hoisting it to the sibling container (alongside `sheetTopDraggingArea`) restores tap interaction without affecting visual placement.


yceballost (approved):
nice

PR #1555, opened 2026-06-01
feat(Pagination): add Pagination component aligned with Cyber skin

AlexandraGallipoliRodrigues:
## Summary

Adds a new `Pagination` component to mistica-web, designed for Cyber and built against the standard skin contract.

Ticket: WEB-2441

### Component

- `Pagination` supports controlled (`currentPage`) and uncontrolled (`defaultPage`) usage, with `onChange`.
- Previous and Next use `ButtonLink` controls. They remain visible and disabled at the first/last page.
- `hideNavigationControls` and `hidePageList` allow rendering only the page list or only navigation controls.
- `maxPages` controls the dynamic visible page range. The default is five pages on desktop and three on mobile.
- On mobile, boundary pages are hidden to optimise available space; compact navigation-only mode is enabled explicitly with `hidePageList`.
- `showEllipsis` controls truncation, including the single-hidden-page ellipsis avoidance rule.
- `mode="iconOnly"` renders chevron-only navigation controls.
- `disabled` prevents page changes.
- Invalid page values below `1` throw in development and test environments.

### Styling and contexts

- Page items use a 32px circular visual state inside responsive interaction areas.
- Regular page items use the pagination hover/pressed states; the current page uses `brandLow` and the activated text token.
- Previous and Next use `ButtonLink` without the `small` variant and keep a transparent hover/active background.
- Brand and negative contexts use their respective text and state tokens. `alternative` inherits the default token behaviour.
- `hidePageList` keeps Previous and Next horizontally aligned with the required spacing.

### Accessibility

- Pagination is exposed as a localised `<nav>` landmark that includes the current and total page count.
- The page list is an ordered list and ellipses are excluded from the accessibility tree.
- Previous, Next, page items, and the current page have localised labels.
- The current page remains in the keyboard navigation sequence and exposes `aria-current="page"`.
- Previous and Next remain available to assistive technologies at page boundaries and are announced as disabled.
- Consumers can override navigation labels and the navigation landmark label through props or theme text overrides.

### Tooling and coverage

- Two Playroom snippets: `Pagination` and `Pagination iconOnly`.
- Cyber light and iOS themes exposed in Playroom.
- Storybook story with controls for page state, visibility options, mode, context variant, and navigation labels.
- 15 unit tests covering interaction, controlled/uncontrolled usage, disabled states, invalid values, and page-range resolution.
- 20 screenshot cases covering desktop, mobile, compact navigation-only mode, custom labels, icon-only controls, and theme contexts.

## Screenshots / verification

Run `yarn playroom`, switch to the **Cyber** skin in the selector, and load the `Pagination` snippet. Verify:

- [x] Hover on a non-current page shows a subtle neutral grey background.
- [x] The current page has a light cyan background (`brandLow`) with dark-blue text.
- [x] Previous / Next labels.
image

- [x] Each tap area is 48×48px.
Captura de pantalla 2026-06-29 a las 13 00 14
image
image

## A11y tests

- [x] Android:

https://github.com/user-attachments/assets/180b27fd-12a7-4265-9aab-f66fce53de6b

- [x] IOS:

https://github.com/user-attachments/assets/eb1527cd-9bf1-426e-8d98-99debdd893fd

🤖 Generated with Claude Code


src/pagination.tsx
```diff
@@ -0,0 +1,364 @@
...
+
+    const showPrevious = activePage > 1;
+    const showNext = activePage < totalPages;
+
+    return (
+        <nav
+            aria-label={resolvedAriaLabel}
+            className={classnames(isCompact ? styles.containerCompact : styles.container, {
+                [styles.containerNavOnly]: hidePageList && !isCompact,
+            })}
```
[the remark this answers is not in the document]
AlexandraGallipoliRodrigues: Intentional. The compact view uses containerCompact (vertical layout) which has its own internal spacing. Applying containerNavOnly (16px horizontal gap) on top of compact would override the vertical spacing and break the stacked < / > layout the Figma defines for viewport < 375px. The 16px nav-only gap is specifically the desktop "Previous and Next only" treatment.

src/__screenshot_tests__/pagination-screenshot-test.tsx
```diff
@@ -0,0 +1,46 @@
...
+    const cases: Array<[string, string, Device]> = [];
+    for (const story of STORIES) {
+        for (const device of DEVICES) {
+            cases.push([story.name, story.id, device]);
+        }
+    }
+    return cases;
+};
+
+test.each(getCases())('Pagination %s - %s', async (_name, id, device) => {
```
Marcosld: Perhaps a table?

src/__screenshot_tests__/pagination-screenshot-test.tsx
```diff
@@ -0,0 +1,46 @@
...
+    await openStoryPage({id, device});
+
+    const pagination = await screen.findByTestId('Pagination');
+    const image = await pagination.screenshot();
+    expect(image).toMatchImageSnapshot();
+});
+
+// Compact view only renders below 375px viewport. We use MOBILE_IOS_SMALL (320px)
+// to drive the vertical layout and the current ± 1 page reduction.
```
Marcosld: I see a lot of comments that we probably don't need, right? We should probably update AGENTS.md so it doesn't generate that much comments

src/__stories__/pagination-story.tsx
Marcosld, line 1: I think we should have added controls instead of creating a lot of examples like the rest of the stories. Perhaps this is something we should add to AGENTS.md so it is the default behaviour

src/pagination.tsx
Marcosld, line 1: There are very annoying layout shifts when ellipsis appear and when buttons disappear, I think we should address that @aweell @yceballost

src/pagination.tsx
```diff
@@ -0,0 +1,364 @@
...
+    end = Math.min(totalPages - 1, end);
+
+    for (let page = start; page <= end; page++) {
+        pages.add(page);
+    }
+
+    const orderedPages = Array.from(pages).sort((a, b) => a - b);
+    const items: Array<PaginationItem> = [];
+
+    orderedPages.forEach((page, index) => {
```
Marcosld: Probably .map ?

src/pagination.tsx
```diff
@@ -0,0 +1,364 @@
...
+        }
+
+        if (!isControlled) {
+            setInternalPage(nextPage);
+        }
+
+        onChange?.(nextPage);
+    };
+
+    const items = isCompact
```
Marcosld: We should render compact view depending on css only to avoid layout shifts. window size won't be available until js is executed in the browser

src/pagination.tsx
brtbrt: +1!

src/__stories__/pagination-story.tsx
brtbrt: 100% agreed. we can improve this by exploiting controls. that way it will be improving the DX!

src/__screenshot_tests__/pagination-screenshot-test.tsx
brtbrt: we could take advantage of a feedback loop by using a skill that "auto documents" (better said: it helps updating/creating) things that the AI does not the way we want. each time we say "no", the skill `/create-doc` (or something similar) can update the thing to fine-tune the thing for the next round.
I've tried it and it works quite well! it saved me time and it improved my agentic code experience.  take look :)

https://github.com/Telefonica/webapp/blob/a8839f60de1ce350613289f459abb69222edb495/.agents/skills/create-doc/SKILL.md

p.d.
I've found it very effective the fact of giving a good and bad example. most of the time it does help the AI doing the thing you expect

src/__screenshot_tests__/pagination-screenshot-test.tsx
```diff
@@ -0,0 +1,46 @@
...
+test('Pagination CompactView - MOBILE_IOS_SMALL', async () => {
```
brtbrt: are we sure we want this? by fiddling with the component, I see that even smaller viewports could take advantage of a single line, maybe.

http://localhost:9000/#?code=N4Igxg9gJgpiBcIA8AlGBnADhAduglgG4wAyAhgJ4QCuALgHwA6OABC0gApkDm%2BOZtfLha0ItMgBsu3DAF5gAJgDMAXxZhqAJ00wctaTHkBGNS1gAzMtQn6ehxWqgV%2BAW3xgAwjT0sA9ExwkXzQsXAJicio6ehAVIA

should we talk to design maybe?

src/pagination.tsx
```diff
@@ -0,0 +1,364 @@
...
+    defaultPage?: number;
+    onChange?: (page: number) => void;
+
+    hideNavigationControls?: boolean;
+    hidePageList?: boolean;
+    showEllipsis?: boolean;
+
+    dynamicCount?: number;
+
+    navLeftLabel?: string;
```
aweell, line 33: https://mistica-pl9771ukj-flows-projects-65bb050e.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8AFAhgcwJYDs0BcsIcAdHAAnLwDcAZGAM31rQCMYAbAXhJHxgDO%2BXmUo0ASlgwALZm048%2Bg4SFHl8EfGg7oMgrsACMABgC%2BagLbQYi2AzQBXDirVgHAJ3cwc%2BXTeB6%2BADKWvwAFLwADpi4BEQ4AJKQOH68ADTkAMwAlOYU5MQAwtJoOHoGYdF62eRcAHzkAjDBoTARIFWxhMRJxKkgGVUwuWQA9HVkZKgxeN2k%2BVawiljJAPI4HACeIvkaWjqY%2BkZmrh5ePn4GgSEEbVEzcX2H6Vkj%2BUUlZf6VhzX1jc0buF7thZvF%2BoNfnlxmQBnxpDALIIEABtECFCAWCwOHBYfCbAD6hU27HcIAAumkQAB3LBQfDSASogDsADYABzk0xAA

What i should do differently to display previous/next custom button labels?

src/pagination.tsx
AnaMontes11: I’ve updated the specs to fix it:
**Navigation controls:**

- Previous control is disabled when user is in first page, next control is disabled when user reaches last page.
- By default previous and nexts will be always visible unless are configured to be hidden.

 You can check it [here](https://www.figma.com/design/EGAPoaNdLMiuAUKGfhCtfR/%F0%9F%94%B8-Pagination-Specs?node-id=22-1623&t=wRDsPZoJEtO6Ap7V-4).

playroom/frame-component.tsx
```diff
@@ -62,6 +63,7 @@ const skinToLang: Record<string, string> = {
     [O2_SKIN]: 'en-GB',
     [O2_NEW_SKIN]: 'en-GB',
     [ESIMFLAG_SKIN]: 'es-ES',
+    [CYBER_SKIN]: 'es-ES',
```
aweell, line 66: It looks that previewTools is not working with Cyber skin, is this expected?

https://mistica-pl9771ukj-flows-projects-65bb050e.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8AFATjAbgSxgdwBUIIAbAZwD4kB6dLXQ48igHQDtUBDAc2zc4Au2CGwAE7UaP6YAMjABmAmZwBGMEgF4WIATDIDtEqZ0wAlbNwAWS1eq069BkEYEQBnEih56NwAIwADAC%2BRgC20DD2sPKcAK4kTuJiomCxaBhsAl7ckcA5AgDK7roAFNoADjx8gsJsAJKQbNkw2gA0ogDMAJQhySIAwpacbDm%2BJZU5XaIaFKJkMIXFMGUgE9VCIg0izW2iEzA97NQUIK06ljCheggA2iD9EKGhsWzYAgCeAPr972poIABdM54bBQASWMi3DoANgCAKCQA

AlexandraGallipoliRodrigues: hmmm idk it does work for me...

https://github.com/user-attachments/assets/d0c5affc-65c7-4724-b74b-ca550a114b4d

Marcosld: I think it is the PreviewTools component that is broken

brtbrt: @Marcosld  correct, we figured it out yesterday in pair programming. the skin is missing/not supported yet. I'd open a ticket to address it in a different branch. it's a quick and easy fix :)

src/__screenshot_tests__/pagination-screenshot-test.tsx
AlexandraGallipoliRodrigues: @aweell @AnaMontes11 what do you think?
https://mistica-ck9tg1gt6-flows-projects-65bb050e.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8AFAhgcwJYDs0BcsIcACfCfNAG3QxgGcBeYAJgGYBfE2AMzQFcq%2BWjGYsuAegB8AHRwgANCHwALGAFsGCANogAwhHXr%2BOLPgCeAfT3mARjABOi-YeOmL1u48tYA8gGUQAF0lAHcsKFV6HTYANgAGBQBGeJYAFiCOIA

src/pagination.tsx
AlexandraGallipoliRodrigues: The navLeftLabel/navRightLabel props are rendered on desktop viewports (≥1024px), or at least that's what I understand from the Figma, should i change it?

Image

https://mistica-pl9771ukj-flows-projects-65bb050e.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8AFAhgcwJYDs0BcsIcAdHAAnLwDcAZGAM31rQCMYAbAXhJHxgDO%2BXmUo0ASlgwALZm048%2Bg4SFHl8EfGg7oMgrsACMABgC%2BagLbQYi2AzQBXDirVgHAJ3cwc%2BXTeB6%2BADKWvwAFLwADpi4BEQ4AJKQOH68ADTkAMwAlOYU5MQAwtJoOHoGYdF62eRcAHzkAjDBoTARIFWxhMRJxKkgGVUwuWQA9HVkZKgxeN2k%2BVawiljJAPI4HACeIvkaWjqY%2BkZmrh5ePn4GgSEEbVEzcX2H6Vkj%2BUUlZf6VhzX1jc0buF7thZvF%2BoNfnlxmQBnxpDALIIEABtECFCAWCwOHBYfCbAD6hU27HccIxWJxeMJxNJBKwqyCIAAumkQAB3LBQfDSASokwAJgALGlDABWQwC0UADkyAuZpiAA

playroom/frame-component.tsx
Marcosld: https://github.com/Telefonica/mistica-web/pull/1569

src/pagination.tsx
```diff
@@ -0,0 +1,322 @@
...
+    totalPages: number;
+    currentPage?: number;
+    defaultPage?: number;
+    onChange?: (page: number) => void;
+
+    hideNavigationControls?: boolean;
+    hidePageList?: boolean;
+    showEllipsis?: boolean;
+
+    dynamicCount?: number;
```
yceballost: maybe `maxPages` ? trying to parity with StackingGroup component that one of it props is `maxItems`

src/pagination.tsx
```diff
@@ -0,0 +1,322 @@
...
+                                <PaginationLabel weight="medium">...</PaginationLabel>
+                            </span>
+                        </li>
+                    );
+                }
+
+                if (item.current) {
+                    return (
+                        <li key={item.page} className={liClassName}>
+                            <button
```
yceballost: why not use `Touchable` component? just asking!

src/pagination.tsx
```diff
@@ -0,0 +1,322 @@
...
+    return (
+        <nav
+            aria-label={resolvedAriaLabel}
+            className={classnames(styles.container, {
+                [styles.containerNavOnly]: hidePageList,
+            })}
+            {...getPrefixedDataAttributes(dataAttributes, 'Pagination')}
+        >
+            {!hideNavigationControls && (
+                <button
```
yceballost: why not use `<ButtonLink>` for previous and next pages?

now, these buttons is not having hover effects and weirds horizontal spacings in active state

yceballost (changes requested):

src/pagination.tsx
yceballost, line 1: looks broken in playroom 🤔
https://mistica-qux1oxj9x-flows-projects-65bb050e.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8AFAhgcwJYDs0BcsIcAdHAAnIFtoYBeEkWAMzQFcAbfR8vANwBKWDAAt8AGTQAjGBwYg0AZyhLGZSvgj40HdBhiK6wAIwAGAL7ryYNgCdbMHPj31g%2B-AGVt%2BGAApGAA6YuAREOC6MADTkAEwAlJYU5MQAwiJoOPpGvkH6ceR0AHzkijCe3n6BwXiExBEg0bkwCWQA9IVkIOZAA

AnaMontes11:
The ellipsis element uses `textSecondary` as its token in the default context. 
Which token should it use in the brand, negative, and alternative contexts?

playroom/frame-component.tsx
brtbrt: @AlexandraGallipoliRodrigues we can resolve this I guess!

src/__stories__/pagination-story.tsx
```diff
@@ -0,0 +1,74 @@
...
+import {Pagination} from '..';
+
+export default {
+    title: 'Components/Pagination',
+    argTypes: {
+        mode: {
+            options: ['default', 'iconOnly'],
+            control: {type: 'select'},
+        },
+        totalPages: {control: {type: 'number'}},
```
brtbrt: some parameters could be `undefined`. can we adapt storybook to showcase that? plus covering with tests!
plus: the counter allows negative numbers. do we want it?

src/pagination.tsx
brtbrt: @yceballost can you screenshot it? I see it working right

brtbrt: @AlexandraGallipoliRodrigues have you covered these cases with a unit (or acceptance) test? just checking!

src/__screenshot_tests__/pagination-screenshot-test.tsx
```diff
@@ -0,0 +1,58 @@
...
+            hidePageList: true,
+            navLeftLabel: 'Anterior capítulo',
+            navRightLabel: 'Siguiente capítulo',
+        },
+    },
+];
+
+const DEVICES: ReadonlyArray<Device> = ['DESKTOP', 'MOBILE_IOS'];
+
+const TABLE = CASES.flatMap((c) => DEVICES.map((device) => ({...c, device})));
```
brtbrt: I find it "difficult" to read. I mean, I'd prefer readability over non-duplication here, if it applies. I usually prefer a declarative, yet longer, `test.each` that give me the whole picture in a glance of what I'm testing

src/__stories__/pagination-story.tsx
brtbrt, line 1: I see that some props are not exposed on storybook right? I'd cover `onChange` because it's important, even though it's test covered!

src/__tests__/pagination-test.tsx
```diff
@@ -0,0 +1,129 @@
...
+    const onChange = jest.fn();
+    render(
+        <ThemeContextProvider theme={makeTheme()}>
+            <Pagination totalPages={5} defaultPage={1} onChange={onChange} />
+        </ThemeContextProvider>
+    );
+
+    await userEvent.click(screen.getByRole('button', {name: 'Ir a la página 3'}));
+
+    expect(onChange).toHaveBeenCalledWith(3);
```
brtbrt: I see we always use the same number for testing things? 5 pages, call next expected 3. 

1. I'd change the values of the test below at least, just to broaden the test cases and make them more robust
2. a more general opinion (and to be discussed outside this PR): what about using something like https://www.npmjs.com/package/@faker-js/faker to create stable tests, but with data that could cover even more cases?

src/pagination.css.ts
```diff
@@ -0,0 +1,262 @@
+import {style} from '@vanilla-extract/css';
+import {sprinkles} from './sprinkles.css';
+import {vars as skinVars} from './skins/skin-contract.css';
+import * as mq from './media-queries.css';
+
+const COMPACT_MQ = '(max-width: 374px)';
+
+export const container = style([
+    sprinkles({
```
brtbrt, line 7: here we go. I don't remember anything about sprinkles. my next week's due chore: go over the sprinkles documentation. 🥵

src/pagination.css.ts
```diff
@@ -0,0 +1,262 @@
+import {style} from '@vanilla-extract/css';
+import {sprinkles} from './sprinkles.css';
+import {vars as skinVars} from './skins/skin-contract.css';
+import * as mq from './media-queries.css';
+
+const COMPACT_MQ = '(max-width: 374px)';
```
brtbrt: do we really want the canvas to determine the pagination compact mode? or the real available space where it's used in? just checking, pals! you tell me

src/pagination.tsx
brtbrt: +1, I was feeling "lost" with  that semantic param description.

src/pagination.tsx
```diff
@@ -0,0 +1,309 @@
...
+        t(texts.paginationGoToPage || tokens.paginationGoToPage, String(page));
+
+    return (
+        <ul className={classnames(styles.pageList, className)}>
+            {items.map((item, index) => {
+                if (item.type === 'ellipsis') {
+                    return (
+                        <li key={`ellipsis-${index}`} className={styles.pageListItem}>
+                            <span className={styles.ellipsis} aria-hidden="true">
+                                <PaginationLabel weight="medium">...</PaginationLabel>
```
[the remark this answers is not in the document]
brtbrt: @AlexandraGallipoliRodrigues actually it could be right

src/pagination.tsx
brtbrt: @AlexandraGallipoliRodrigues  maybe we could give claude a round telling to try to exploit the mística components that we have already. let's see if it improves it. after the round, you could ask "why a button in place of a touchable?" to force the reasoning

src/pagination.css.ts
brtbrt, line 1: I see minor issues in the styling, that maybe they could be addressed later on.

example:
Image

see the selection? the left padding is not the same of the right one! the same applies for the previous button. could you fiddle with it a little bit to see if we can improve it?

src/pagination.tsx
yceballost: image

src/pagination.css.ts
yceballost: I reported the same here!
https://github.com/Telefonica/mistica-web/pull/1555#discussion_r3396221509

AnaMontes11:
hi @AlexandraGallipoliRodrigues I've updated the compact view spec. Reviewing it, the horizontal row fits fine at 375px [see example](https://mistica-qux1oxj9x-flows-projects-65bb050e.vercel.app/?path=/story/components-pagination--default&args=totalPages:13;currentPage:4;dynamicCount:4;mode:iconOnly&globals=viewport.value:375-568), so forcing the cross layout there was unnecessary. There's no automatic layout at 375px anymore: for space-constrained contexts we use `hidePageList`, which hides the Page List and leaves only the Previous and Next arrows aligned horizontally, avoiding scroll. I've added an example of the < 375px view with controls only to make it clear how it looks.
[Link to specs](https://www.figma.com/design/EGAPoaNdLMiuAUKGfhCtfR/%F0%9F%94%B8-Pagination-Specs?node-id=2072-29490&t=HqWWwjgNxxq69eE9-4)

src/pagination.tsx
AlexandraGallipoliRodrigues: image

Because it isn’t focusable when disabled, but Figma doesn’t specify this scenario for accessibility. I think that if the button isn’t clickable, there’s no need to indicate this, and the component’s default behaviour is fine, so I’ll change it

src/pagination.tsx
AlexandraGallipoliRodrigues: it was specified in figma, was the case of the compact view. But the figma has been updated, so i'll change it

AnaMontes11: > it was specified in figma, was the case of the compact view. But the figma has been updated, so i'll change it

I have a comment below explaining the reasoning. thanks @AlexandraGallipoliRodrigues

AlexandraGallipoliRodrigues: done it, is it okey now?

brtbrt:
hey hey @AlexandraGallipoliRodrigues  it's failing the build. once you've fixed it I'll be reviewing it properly. feel free to re-request the review so that it notifies me :)

src/pagination.tsx
```diff
@@ -0,0 +1,376 @@
...
+                                onPress={() => {}}
+                                aria-current="page"
+                                aria-disabled
+                                aria-label={currentPageLabel(item.page)}
+                            >
+                                <span className={styles.pageContent}>
+                                    <PaginationLabel weight="medium">{item.page}</PaginationLabel>
+                                </span>
+                            </Touchable>
+                        </li>
```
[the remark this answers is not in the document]
AlexandraGallipoliRodrigues: we intentionally keep the current page in the keyboard navigation order because the Figma accessibility specification includes it as an interactive target. No change needed here.

src/__tests__/pagination-test.tsx
AlexandraGallipoliRodrigues: Done. I varied the interaction test data to cover different page ranges and transitions. I agree deterministic test data is sufficient here; using faker would be better discussed separately if we want to adopt it consistently across the repository

Marcosld (changes requested):
There's still layout shift when navigating almost all the pages in: https://mistica-cova6o2a5-flows-projects-65bb050e.vercel.app/?path=/story/components-pagination--default. Also hover works when elements are disabled.. is that ok?

AnaMontes11:
Hi @AlexandraGallipoliRodrigues 

- For the Navigation Controls, could you use the Icon Button and Button Link components? I think they are not being applied at the moment, so the component is not working as expected.
- The ellipsis element in the default and alternative contexts uses the text secondary color token so that it is visually differentiated from the page numbers.

AnaMontes11:
For viewports below 375px, I noticed that the component is not behaving as expected. The Page List should be hidden using `hidePageList`, leaving only the Previous and Next controls displayed and aligned horizontally.

AlexandraGallipoliRodrigues:
> For viewports below 375px, I noticed that the component is not behaving as expected. The Page List should be hidden using `hidePageList`, leaving only the Previous and Next controls displayed and aligned horizontally.

Updated. The component now hides the page list by default below 375px.

For mobile widths with the page list visible, the recommended configuration is `mode="iconOnly"`; otherwise the Previous/Next labels plus the page items do not fit in the available width. The component still supports `mode="default"` with labels, and this can be selected from Storybook controls. Is that expected @AnaMontes11 @yceballost ??

The screenshots use `mode="iconOnly"` for mobile page-list examples to match the layout that fits the Figma spec and also because with `mode="default"` + more tha 5 pages does not fit thw available with

src/pagination.tsx
AlexandraGallipoliRodrigues: Unit tests cover:
- Previous/Next are always rendered at boundaries and become disabled instead of disappearing.
- The number of pagination items stays stable while changing pages, both for desktop and mobile ranges.

Screenshot tests cover the visual states for first page, last page, ellipsis, mobile, desktop, and compact view.

src/__screenshot_tests__/__image_snapshots__/pagination-screenshot-test-tsx-pagination-alternative-context-desktop-1-snap.png
brtbrt, line 1: I have troubles loading the pictures. is it only me or a github glitch?

src/__screenshot_tests__/pagination-screenshot-test.tsx
brtbrt: @aweell ping 🙃

src/__screenshot_tests__/pagination-screenshot-test.tsx
```diff
@@ -0,0 +1,198 @@
...
+
+test.each(CASES)('Pagination $name - $device', async ({args, device}) => {
+    await openStoryPage({id: STORY_ID, device, args});
+
+    const pagination = await screen.findByTestId('Pagination');
+    const image = await pagination.screenshot();
+    expect(image).toMatchImageSnapshot();
+});
+
+test('Pagination interactive items have 48px targets on mobile', async () => {
```
brtbrt: since it's not screenshotting things, maybe we can move to an acceptance test

src/__screenshot_tests__/pagination-screenshot-test.tsx
```diff
@@ -0,0 +1,198 @@
...
+
+test('Pagination ellipses are non-interactive and keep the mobile layout size', async () => {
+    await openStoryPage({
+        id: STORY_ID,
+        device: 'MOBILE_IOS',
+        args: {totalPages: 20, currentPage: 10, mode: 'iconOnly'},
+    });
+
+    const pagination = await screen.findByTestId('Pagination');
+    const ellipses = await pagination.$$('li[aria-hidden="true"]');
```
brtbrt: maybe weak? WDYT guys?

src/__screenshot_tests__/pagination-screenshot-test.tsx
```diff
@@ -0,0 +1,198 @@
...
+        expect(await ellipsis.evaluate((element) => (element as HTMLElement).tabIndex)).toBe(-1);
+    }
+});
+
+test.each`
+    variantOutside   | color
+    ${'default'}     | ${'rgb(96, 94, 92)'}
+    ${'brand'}       | ${'rgb(255, 255, 255)'}
+    ${'negative'}    | ${'rgb(255, 255, 255)'}
+    ${'alternative'} | ${'rgb(96, 94, 92)'}
```
brtbrt: just checking: are we sure we want them hardcoded here? it could be a good idea (stable tests) as well as a bad one in my mind? I don't see any standard to follow among other existing tests

src/__stories__/pagination-story.tsx
brtbrt: @AlexandraGallipoliRodrigues if you're done with this feel free to resolve opened conversations :) this way will be way more easier to follow the pending changes (and review it)

src/pagination.tsx
```diff
@@ -0,0 +1,364 @@
...
+    dataAttributes,
+    'aria-label': ariaLabel,
+}: PaginationProps): JSX.Element | null => {
+    const isControlled = currentPage !== undefined;
+    const [internalPage, setInternalPage] = React.useState(defaultPage);
+    const {texts, t} = useTheme();
+    const variant = useThemeVariant();
+    const {isTabletOrSmaller} = useScreenSize();
+    const windowWidth = useWindowWidth();
+    const hidePageList = hidePageListProp ?? windowWidth < 375;
```
brtbrt: magic numbers are quite dangerous in my mind! I'd separate the const literal into a semantic const with a proper name. I'd involve design for that

src/pagination.tsx
```diff
@@ -0,0 +1,364 @@
...
+
+    dataAttributes?: DataAttributes;
+    'aria-label'?: string;
+};
+
+type PaginationItem = {type: 'page'; page: number; current: boolean} | {type: 'ellipsis'};
+
+const clamp = (value: number, min: number, max: number): number => Math.min(Math.max(value, min), max);
+
+const validatePositivePageNumber = (name: string, value: number | undefined) => {
```
brtbrt: do we want this? I think I've never seen this behavior. I've always seen a graceful degradation in these cases. in case we want it, please double check we have it test-covered! Talk to the stakeholders to see what to expect in those "wrong" props cases

src/pagination.tsx
```diff
@@ -0,0 +1,364 @@
...
+    mode?: 'default' | 'iconOnly';
+    disabled?: boolean;
+
+    dataAttributes?: DataAttributes;
+    'aria-label'?: string;
+};
+
+type PaginationItem = {type: 'page'; page: number; current: boolean} | {type: 'ellipsis'};
+
+const clamp = (value: number, min: number, max: number): number => Math.min(Math.max(value, min), max);
```
brtbrt: 🥵 I find it difficult associating a clamping to this. can we find a better semantic name?

src/pagination.tsx
```diff
@@ -0,0 +1,364 @@
...
+    navRightLabel?: string;
+
+    mode?: 'default' | 'iconOnly';
+    disabled?: boolean;
+
+    dataAttributes?: DataAttributes;
+    'aria-label'?: string;
+};
+
+type PaginationItem = {type: 'page'; page: number; current: boolean} | {type: 'ellipsis'};
```
brtbrt: ExclusifyUnion maybe better defines this

src/pagination.tsx
```diff
@@ -0,0 +1,364 @@
...
+};
+
+type PageListProps = {
+    items: Array<PaginationItem>;
+    disabled?: boolean;
+    className?: string;
+    onPageClick: (page: number) => void;
+};
+
+const TILE_STYLE: React.CSSProperties = {
```
brtbrt: Not a big fan of uppercase consts, but that's my personal taste :)

src/pagination.tsx
```diff
@@ -0,0 +1,364 @@
...
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
+    width: '100%',
+    height: '100%',
+};
```
brtbrt: as const?

src/__screenshot_tests__/pagination-screenshot-test.tsx
```diff
@@ -0,0 +1,198 @@
...
+        args: {totalPages: 20, currentPage: 10, variantOutside},
+    });
+
+    const pagination = await screen.findByTestId('Pagination');
+    const ellipsis = await pagination.$('li[aria-hidden="true"] span');
+
+    expect(await ellipsis?.evaluate((element) => getComputedStyle(element).color)).toBe(color);
+});
+
+test('Pagination compact view fits at 360px', async () => {
```
Marcosld: I think compact view is meant to be enabled manually, like a prop: `withCompactView: boolean` that indicates if it should be enabled below 375px @AnaMontes11

src/__screenshot_tests__/pagination-screenshot-test.tsx
```diff
@@ -0,0 +1,198 @@
+import {openStoryPage, screen} from '../test-utils';
+
+import type {Device, StoryArgs} from '../test-utils';
+
+const STORY_ID = 'components-pagination--default';
+
+const CASES: ReadonlyArray<{name: string; device: Device; args: StoryArgs}> = [
+    {name: 'Default', device: 'DESKTOP', args: {totalPages: 9, currentPage: 3}},
+    {name: 'Default', device: 'MOBILE_IOS', args: {totalPages: 9, currentPage: 3, mode: 'iconOnly'}},
```
Marcosld: I think iconOnly should be the default behavior for mobile viewports, even tho it should also be configurable @AnaMontes11

src/pagination.tsx
Marcosld: I wouldn't include it

src/pagination.tsx
Marcosld: I wouldn't include this either

src/pagination.tsx
Marcosld: Please use CSS queries, not JS

ieduardogf (approved):

AnaMontes11:
Captura de pantalla 2026-06-30 a las 14 34 14
hi @AlexandraGallipoliRodrigues I think this is related to the visible range point I raised earlier. Here the first and last pages, which are fixed, are missing. The structure should be: fixed first + ellipsis + dynamic + ellipsis + fixed last.
Could you confirm what range value this example is using? The default range should be 3.

Thanks!

src/pagination.tsx
```diff
@@ -0,0 +1,387 @@
...
+    'aria-label': ariaLabel,
+}: PaginationProps): JSX.Element | null => {
+    const isControlled = currentPage !== undefined;
+    const [internalPage, setInternalPage] = React.useState(defaultPage);
+    const {texts, t} = useTheme();
+    const variant = useThemeVariant();
+    const {isMobile, isTabletOrSmaller} = useScreenSize();
+    const hidePageList = hidePageListProp === true;
+    const compactView = withCompactView && !hideNavigationControls;
+    const resolvedMode = mode ?? (isTabletOrSmaller ? 'iconOnly' : 'default');
```
yceballost: I dont really like this logic, its not very transparent. I would removed it.

If a product want to add this logic for his product is something to do in the product, not in the component imo

src/__screenshot_tests__/__image_snapshots__/pagination-screenshot-test-tsx-pagination-default-mobile-ios-1-snap.png
yceballost, line 1: The last and first pages are always should be displayed, right?

@AnaMontes11 is this correct?

src/pagination.tsx
```diff
@@ -0,0 +1,387 @@
...
+            {!hidePageList && (
+                <PageList
+                    items={items}
+                    disabled={disabled}
+                    className={classnames({[styles.pageListCompact]: compactView})}
+                    onPageClick={goToPage}
+                />
+            )}
+            {!hideNavigationControls &&
+                (resolvedMode === 'iconOnly' ? (
```
yceballost: navigation buttons and icons is not showing hover states

Why are you using an InternalIconButton instead of `<IconButton>`? just curious

playroom/snippets.tsx
```diff
@@ -4583,6 +4583,27 @@ export default [
...
+        code: `
+            <Pagination
+                totalPages={10}
+                currentPage={getState("paginationPage", 3)}
+                onChange={(page) => setState("paginationPage", page)}
+            />`,
+    },
+    {
+        group: 'Pagination',
+        name: 'Pagination iconOnly',
```
yceballost: I would remove this snippet, just one snippet with all props should be enough

```
<Pagination
  mode="iconOnly"
  showEllipsis={true} 
  totalPages={10}
  currentPage={getState("paginationPage", 3)}
  onChange={(page) => setState("paginationPage", page)}
/>

```

src/__stories__/pagination-story.tsx
yceballost, line 1: include pagination component also in skin story please! 

https://mistica-kg4h93k5y-flows-projects-65bb050e.vercel.app/?path=/story/private-components-in-different-skins--default

this is the way to check if it is working in the rest of skins

src/pagination.tsx
```diff
@@ -0,0 +1,387 @@
...
+type EllipsisItem = {type: 'ellipsis'};
+type PaginationItem = ExclusifyUnion<PageItem | EllipsisItem>;
+
+const DEFAULT_DYNAMIC_PAGE_COUNT = 3;
+
+export const getPaginationItems = ({
+    totalPages,
+    currentPage,
+    maxPages = DEFAULT_DYNAMIC_PAGE_COUNT,
+    showEllipsis = true,
```
yceballost: IMO, showEllipsis should be false by default, just use showEllipsis if you want this feature

@aweell @AnaMontes11 are u agree? wdyt?

yceballost (changes requested):

src/__screenshot_tests__/__image_snapshots__/pagination-screenshot-test-tsx-pagination-default-mobile-ios-1-snap.png
AlexandraGallipoliRodrigues: on mobile I think not
image

src/pagination.tsx
Marcosld: I think this is the intended behavior according to the figma: https://www.figma.com/design/EGAPoaNdLMiuAUKGfhCtfR/%F0%9F%94%B8-Pagination-Specs?node-id=13-308&m=dev

yceballost: Checking internally

yceballost: Solved. We are going to remove this logic from specs. 

This logic have to be managed in the product

Marcosld: Sorry por el mareo @AlexandraGallipoliRodrigues :'(

src/pagination.tsx
AlexandraGallipoliRodrigues: because IconButton has a hover and I thought we didn't want it, but I just speaked with Ana and I'll change it to IconButton again

src/pagination.tsx
aweell: How this works with `maxPages`? If `showEllipsis` is false even if `maxPages` is by default defined as 3 all pages will be shown?

Why the ellipsis is not internally dependant of `maxPages` value? By default is 3 and will show ellipsis when defined and value is < totalPages

AlexandraGallipoliRodrigues: I’ve updated the logic so `maxPages` now enables ellipsis automatically when it’s explicitly provided and its value is lower than `totalPages`.

So the behavior is now:

- If neither `showEllipsis` nor `maxPages` is provided, all pages are shown.
- If `maxPages` is provided, ellipsis are shown when needed.
- If `showEllipsis={false}` is explicitly provided, it forces all pages to be shown even if `maxPages` is set, do you agree with this @aweell @yceballost @AnaMontes11 ?
- If `showEllipsis={true}` is provided without `maxPages`, it uses the default dynamic count of 3.

I also added unit tests covering these cases.

src/__screenshot_tests__/__image_snapshots__/pagination-screenshot-test-tsx-pagination-alternative-context-desktop-1-snap.png
yceballost: only you, I can see the pictures

brtbrt: (yep, it was an adblock)

aweell (commented):
Is expected that mobile viewport don't show iconButtons when mode is `iconOnly`?

https://mistica-jj6b6j3am-flows-projects-65bb050e.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8AFATjAbgSxgdwBUIIAbAZwD4kB6dLXQ48igHQDtUBDAc2zc4Au2CG3YACMQFtoMALwsQ2SGwDybEgE8FY8WIEQBnEih4wys4AEYArAF9dZAK5o0ER2yh9uJ7jADCbmwCFgBM9mwSYM4YQT5ywL4CAMqGAjAAFAoADjx8gsJscQoANGIAzACU4RIifgAWnGy%2BFuk5vhVishRiZDDJqRnZufxCIkUgpW0wVezUrGwTIAJ1MJJmCADaIACyEDhkhmgA%2BgBy%2BCAAusUgeNhQy2SbZQBsAAzFAOzPABzFNpYhC62IA

src/pagination.tsx
```diff
@@ -0,0 +1,457 @@
...
+};
+
+export const Pagination = ({
+    totalPages,
+    currentPage,
+    defaultPage = 1,
+    onChange,
+    hideNavigationControls = false,
+    hidePageList: hidePageListProp,
+    surroundingPageCount,
```
aweell, line 236: The spec was mean to default this prop as 1, does this make sense or may be a reason to be aware of and we shouldn't provide a dafault here?

https://www.figma.com/design/EGAPoaNdLMiuAUKGfhCtfR/%F0%9F%94%B8-Pagination-Specs?node-id=171-13646&t=drXmc3PPshP6l93b-4

src/pagination.tsx
```diff
@@ -0,0 +1,457 @@
...
+            ref={setContainerElement}
+            aria-label={resolvedAriaLabel}
+            className={classnames(styles.container, {
+                [styles.containerNavOnly]: hidePageList,
+            })}
+            {...getPrefixedDataAttributes(dataAttributes, 'Pagination')}
+        >
+            {!hideNavigationControls &&
+                (mode === 'iconOnly' ? (
+                    <IconButton
```
yceballost, line 307: missing bleed prop for iconButton too

https://mistica-jj6b6j3am-flows-projects-65bb050e.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8AlGBnADhAduglgG4wAyAhgJ4QCuALgHwA6OABC0gAoA2ZYMAFhC6wATvSQB6brwFDRTHMzacyAc3w4ytfLiVsWAW2gwAvIxD5IOAPI4uFc3ra0ItMlw5qMJ4AEYADAC%2BTizo1CIiNDhQGqqeqjAAwlG0Pr7BrPpg4SIwOLTxpsAJtADKbrQwABTmmGoaWjo4heYANCwAzACUGfosuIn8ZDgJPlV1CV0sJvShMGUV1bX1mtq4LSDtEzA9ehIKyvENa4qZzq7uhehpQSFhEVExI4XJ1PlpvVk5eQVePiXlLRLEATY5NDbtbqfNgDIYjIrjLxTGZzBZAmoglaNdZeNosba7TL7Zh6JAAQS4%2BFUrAAHmYQHkoOYDvoVOpVk0QmwjLB6ZZcLZ7I4zvoXG4PF5rn5biK2PdIm8nnEvK93n5oV8Ij9Cv95oDKhjQRycQk8VCuf0cINhqNgIjJtNZug9YtDViThD8UiNcSRWywbpZSwxZdJTcNXKco9Yi8Uh8LdktfkdcUXejluzsc1cZtOoS%2BjCrXDbfado7UfrgUas56CT6WZIKVScA3Eu4uDRaCEyOhnalgEgAJJWABCEBpaFU1B4IhY%2Bw1Aw7zrGyNmwECGu0tC4pnMpQgBhgwfw25gwr6sHQYBE%2BEwJ3pABV%2BPh0CwXywyCxL9fbycWAAzCBZ1ofgjzAdtO3PfQACM6BcHAfAtJARzg9YbwMMgRAoUIMK4LhLQ4XJexXct10CFkCxYMkwBOJCJBQ2h4MI-AMKwiiWA1WDGNwEgNAAa0QoNkNQnBeJwPicPbAiiKlKpVxYMj2L6MS%2BLohj4JU9jPl9SQ0CwXACGIcgqDoBRNhAECYEPdAEAAbRAABZCBCBfNwRAAfQAORgAB3EAAF1WhAHz8CgECbPgWyAHYADYAA5-MCIA

src/pagination.css.ts
```diff
@@ -0,0 +1,302 @@
...
+        color: skinVars.colors.textSecondary,
+    },
+    media: {},
+});
+
+export const navigationButtonLink = style({
+    flexShrink: 0,
+    selectors: {
+        '&&:hover:not([disabled]), &&:active:not([disabled])': {
+            backgroundColor: 'transparent',
```
yceballost: why? 🤔

src/pagination.tsx
```diff
@@ -0,0 +1,457 @@
...
+                        ref={prevControlRef}
+                        Icon={IconChevronLeftRegular}
+                        type="brand"
+                        backgroundType="transparent"
+                        disabled={disabled || isPrevDisabled}
+                        aria-label={resolvedPrevAriaLabel}
+                        onPress={() => goToPage(activePage - 1)}
+                    />
+                ) : (
+                    <ButtonLink
```
yceballost, line 318: missing "small" prop in all of buttonLinks and in IconButtons

@AnaMontes11 can you confirm this? because I see all of the examples in specs with buttons in small variants but I cant find nothing about the button have to be small in text.

yceballost (changes requested):

src/pagination.tsx
AlexandraGallipoliRodrigues: i thought after the meeting that we wanted to fit as many pages as possible in the screen but i'll change it to 1

src/pagination.css.ts
AlexandraGallipoliRodrigues: ButtonLink doesn't expose a prop to disable its hover/pressed background. The Pagination spec shows Previous/Next as navigation text controls without the ButtonLink background treatment, so this override keeps the ButtonLink behavior/accessibility while removing only the background state for this component.
https://mistica-web.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8AhArgFwxAdgGQEscBrAAlwAUAnGAZ1oF5gAKASlIYD5TgBfXzgEEwGAriQB6dFlyESnADo4QAGhAYAFjAC2dBAG0QANQIA3CAH0AcjADuIALprbBKJtoGAzADYADA94gA

src/pagination.tsx
AlexandraGallipoliRodrigues: image
I avoided adding bleed to IconButton because it clips the hover/pressed background at the container edge. ButtonLink only needs the text padding removed, but IconButton's background is the visible interaction area.

src/pagination.css.ts
Marcosld: I think we wanted to keep the default behavior. I can't find where in the spec do we ask to disable hover/pressed

src/pagination.tsx
aweell: Since the pagination should be at least always inside a responsive layout, shouldn't clip right?

Making the example I've seen a weird behaviour in the icon button with responsive:

https://mistica-jj6b6j3am-flows-projects-65bb050e.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8AlGBnADhAduglgG4wAyAhgJ4QCuALgHwA6OSACmQOb45m367MABIIC20GAF5GIfJBwB5HABsK0oYNoRaZJew4YJwAIwAGAL7r01AE7WaOKNw56YAYXu1DRizmFgb1jA4tC6G%2BrQAytq0MAAU0pic3Lz8OC7SADSCAMwAlD7CuK4AFmQ4%2Boaxifq5ghL0gugwkdFxCUk8fLjpIFnVMPnMAPRMLENoWLgExORUdPS9ILTFMCIYCADaIACyEIT46NrWAPoAcjAA7iAAuhkgF-hQy%2Bib2QBsJhkA7G8AHBlGACsRgATNczEA

Screenshot 2026-07-03 at 09 42 11

src/pagination.css.ts
```diff
@@ -0,0 +1,302 @@
...
+            },
+        },
+    },
+]);
+
+export const pageListItem = style({
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
+    width: 48,
```
aweell: https://www.figma.com/design/EGAPoaNdLMiuAUKGfhCtfR/%F0%9F%94%B8-Pagination-Specs?node-id=2734-1419&t=drXmc3PPshP6l93b-4

Current implementation shows overflowing page numbers for more than 2 figures

https://mistica-jj6b6j3am-flows-projects-65bb050e.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA6A7APABQIYHMCWKWALnhCqgAQUC20MAvEiHpCgPIoA2Ank5RUQhEsnbDhgBnesACMAJgAMCgL78JAVwBOmiOpRQCOMTADCulEWkzVKKmC2aYF49PFEAysKIwAFEwAOuATEpCjGTAA0FADMAJQ2VGQmABZYKOLSPoHisRT0AHwUEjAeXr4BQYQkZOEgUdkw8agA9PmodSBEyTDUkggA2iAAshAAbngSwpoA%2BgByMADuIAC6ESALeFBdEgPRAGwKEQDsewAcETIArPLLykA

Image

src/pagination.tsx
Marcosld: Nice, by defaulting to 1 we can remove a lot of code :P

src/pagination.tsx
```diff
@@ -0,0 +1,457 @@
...
+    totalPages: number;
+    currentPage: number;
+    surroundingPageCount?: number;
+    includeBoundaryPages?: boolean;
+}): Array<PaginationItem> => {
+    if (totalPages <= 1) {
+        return [];
+    }
+
+    const activePage = Math.min(Math.max(currentPage, 1), totalPages);
```
Marcosld: min and max are overly defensive and repeated all over the place

src/pagination.tsx
```diff
@@ -0,0 +1,457 @@
...
+    const surroundingCount = Math.max(0, Math.floor(surroundingPageCount ?? totalPages));
```
Marcosld: Same. Also surrounding count should be a mandatory prop

src/pagination.tsx
```diff
@@ -0,0 +1,457 @@
...
+    if (surroundingPageCount === undefined || totalPages <= visibleCount) {
+        return Array.from({length: totalPages}, (_, index) => {
+            const page = index + 1;
+            return {type: 'page', page, current: page === activePage};
+        });
+    }
+
+    const boundaryCount = includeBoundaryPages ? 1 : 0;
+    const leftCount = Math.floor((visibleCount - 1) / 2);
+    const rightCount = visibleCount - 1 - leftCount;
```
Marcosld: This is just the same than surroundingCount because it is symmetric

src/pagination.tsx
```diff
@@ -0,0 +1,457 @@
...
+    const visibleCount = 2 * surroundingCount + 1;
+
+    if (surroundingPageCount === undefined || totalPages <= visibleCount) {
+        return Array.from({length: totalPages}, (_, index) => {
+            const page = index + 1;
+            return {type: 'page', page, current: page === activePage};
+        });
+    }
+
+    const boundaryCount = includeBoundaryPages ? 1 : 0;
```
Marcosld: This will provoke layout shift as it depends on isMobile media query. I think instead we should produce the full element list for mobile and desktop and then hide the desktop/mobile specific elements depending on the viewport using CSS. WDYT? (LMK if you have any doubts). Perhaps we can do it in another task to avoid extending this one much longer.

src/pagination.css.ts
AlexandraGallipoliRodrigues: https://github.com/Telefonica/mistica-web/pull/1555#discussion_r3398145912

src/pagination.tsx
```diff
@@ -0,0 +1,457 @@
...
+        type: 'page',
+        page,
+        current: page === activePage,
+    });
+    const createPageRange = (start: number, end: number): Array<PaginationItem> =>
+        Array.from({length: Math.max(0, end - start + 1)}, (_, index) => createPageItem(start + index));
+
+    if (!includeBoundaryPages) {
+        const rangeStart = Math.max(Math.min(activePage - leftCount, totalPages - visibleCount), 2);
+        const rangeEnd = Math.min(rangeStart + visibleCount - 1, totalPages);
```
Marcosld: I think this min is not needed

src/pagination.tsx
```diff
@@ -0,0 +1,457 @@
...
+        const items: Array<PaginationItem> = [];
+
+        if (rangeStart > 2) {
+            items.push({type: 'ellipsis'});
+        } else {
+            items.push(...createPageRange(1, rangeStart - 1));
```
Marcosld:
```suggestion
            items.push(createPageItem(1));
```

src/pagination.tsx
```diff
@@ -0,0 +1,457 @@
...
+        }
+
+        items.push(...createPageRange(rangeStart, rangeEnd));
+
+        if (rangeEnd < totalPages - 1) {
+            items.push({type: 'ellipsis'});
+        } else {
+            items.push(...createPageRange(rangeEnd + 1, totalPages));
```
Marcosld:
```suggestion
            items.push(createPageItem(totalPages));
```

src/pagination.tsx
```diff
@@ -0,0 +1,457 @@
...
+        items.push(createPageItem(boundaryCount + 1));
+    }
+
+    items.push(...createPageRange(siblingsStart, siblingsEnd));
+
+    if (siblingsEnd < totalPages - boundaryCount - 1) {
+        items.push({type: 'ellipsis'});
+    } else if (totalPages - boundaryCount > boundaryCount) {
+        items.push(createPageItem(totalPages - boundaryCount));
+    }
```
Marcosld: Perhaps for symmetry with line 169:
```suggestion
    } else if (boundaryCount < totalPages - boundaryCount) {
        items.push(createPageItem(totalPages - boundaryCount));
    }
```

src/pagination.tsx
```diff
@@ -0,0 +1,457 @@
...
+
+        if (rangeEnd < totalPages - 1) {
+            items.push({type: 'ellipsis'});
+        } else {
+            items.push(...createPageRange(rangeEnd + 1, totalPages));
+        }
+
+        return items;
+    }
+
```
Marcosld, line 111: boundaryCount is always going to equal 1 from here on... Perhaps all this logic would be simpler to follow with hardcoded values? IDK

src/pagination.tsx
```diff
@@ -0,0 +1,457 @@
...
+    children,
+    weight,
+}: {
+    children: React.ReactNode;
+    weight?: PaginationLabelWeight;
+}): JSX.Element => {
+    const {textPresets} = useTheme();
+
+    return (
+        <Text3 as="span" weight={weight ?? textPresets.button.weight} color="currentColor" wordBreak={false}>
```
Marcosld: is this needed?

src/pagination.tsx
```diff
@@ -0,0 +1,457 @@
...
+};
+
+type PaginationLabelWeight = 'regular' | 'medium';
+
+const PaginationLabel = ({
+    children,
+    weight,
+}: {
+    children: React.ReactNode;
+    weight?: PaginationLabelWeight;
```
Marcosld: Shouldn't weight be always the button weight according to specs?

src/pagination.tsx
```diff
@@ -0,0 +1,457 @@
...
+    onPageClick: (page: number) => void;
+};
+
+const TILE_STYLE = {
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
+    width: '100%',
+    height: '100%',
+} as const;
```
Marcosld: why styles here instead of css file?

src/pagination.tsx
```diff
@@ -0,0 +1,457 @@
...
+                        <li key={item.page} className={styles.pageListItem}>
+                            <Touchable
+                                className={classnames(
+                                    styles.currentPage,
+                                    styles.currentPageVariants[variant]
+                                )}
+                                style={TILE_STYLE}
+                                onPress={() => {}}
+                                aria-current="page"
+                                aria-disabled
```
Marcosld, line 199: what about using `disabled` prop?

src/pagination.tsx
```diff
@@ -0,0 +1,457 @@
...
+
+const getTruncatedItemsCount = ({
+    surroundingPageCount,
+    includeBoundaryPages,
+}: {
+    surroundingPageCount: number;
+    includeBoundaryPages: boolean;
+}): number => 2 * surroundingPageCount + (includeBoundaryPages ? 5 : 3);
+
+const getAutoSurroundingPageCount = ({
```
Marcosld: I guess we are removing this and all related logic

Marcosld (changes requested):

src/pagination.tsx
AlexandraGallipoliRodrigues: yep, i'm on it

src/pagination.tsx
AlexandraGallipoliRodrigues: I've simplified the helper to remove repeated min/max where possible. I'm still keeping `currentPage` clamped because it is a public controlled prop and we already support graceful handling for out-of-range page values.

src/pagination.tsx
AlexandraGallipoliRodrigues: As decided the default `surroundingPageCount` should be `1`, so I kept it optional with that default. It remains configurable for products that need a wider dynamic range.

src/pagination.tsx
AlexandraGallipoliRodrigues: Done, removed `leftCount`/`rightCount` and now use `surroundingCount` directly because the range is symmetric.

src/pagination.tsx
AlexandraGallipoliRodrigues: Good point. I haven't moved this to CSS in this PR because it would require rendering both desktop/mobile structures and hiding one with CSS, which is a larger structural change. I'd rather split that into a follow-up if we want to fully avoid SSR/layout-shift concerns here.

src/pagination.tsx
AlexandraGallipoliRodrigues: Done, simplified that range calculation.

src/pagination.tsx
AlexandraGallipoliRodrigues: Done, when the range reaches the beginning we now add the first page directly.

src/pagination.tsx
AlexandraGallipoliRodrigues: Done, same for the last page.

src/pagination.tsx
AlexandraGallipoliRodrigues: Done, removed the generic `boundaryCount` condition and made the desktop first/last page handling explicit

src/pagination.tsx
AlexandraGallipoliRodrigues: Done, `boundaryCount` has been removed. The helper now uses explicit first/last page handling instead of a generic boundary count.

src/pagination.tsx
AlexandraGallipoliRodrigues: Done. I moved those layout rules to the pagination CSS. We still need to override Touchable’s default `display: block / width: 100%` so the page content and current-page background stay centered, but it’s no longer done with inline styles.

src/pagination.tsx
AlexandraGallipoliRodrigues: Done, removed the configurable weight. Pagination labels now always use `textPresets.button.weight`

src/pagination.tsx
AlexandraGallipoliRodrigues: I checked this, but I think we should keep `aria-disabled` instead of `disabled` for the current page. `Touchable disabled` renders a real disabled button, which removes the current page from the interactive/focusable flow. Figma expects the current page to be announced as current, so `aria-current="page"` + `aria-disabled` keeps that behavior without allowing an action.

yceballost (approved):

src/pagination.tsx
aweell: All should appear aligned, I would not worry about the clip as commented earlier and add bleed to iconButton also:

https://mistica-l47vl9auj-flows-projects-65bb050e.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8AlGBnADhAduglgG4wAyAhgJ4QCuALgHwA6OABC0gMq1lgDWLWHjAC8wAIwA2AL5NWbdgElIOAEJ1auZvPm4ACgCcM6UQAoAlC2H0WAd3w4oEGwDoyAGxj7aJxiEyH0dBgoAEJfMyktbRYlXFFYnABhAAsYQn1cEhgAM1o0AHNqNzJ9SLltWgpMEV8AI30yB18o7VqeXnyM6gcAFSqakFoGvEwSmBxaZvL5dABbdzcW%2BVqPYKzcpYB6WW0kNVoNHBJ7XiW2PQDjYHNLa2ApMuiWFZg1nMnpti4S2gT45RSaQyR3eBSKJUe0TmCyWO2iAEEwLR8JppkhNvtDsccLw4WwkLoyPl7GRkainrNoAMACIAUQAYvCAKokHpTJ4abhuQn5DCiMQABkh2nQ1H0XQc9nyPJgiRoE35wvkYDFhgmMtEvNo31oMB8fiJJLJOBlvgANCwAMwRM4sXApRq80yjXkWKwCGDa7i6-Uuo0ok1EmDmlgumA26bbJYEw04UkB22U2DCXz4ZQAeRwbgo7OinPcMqugqVbFF4vlUClMrl3VoittKvF41oGuAWp1et8frjxtNIAt1pLdqSyUdImuYbd1iCXtJnYNxJ7Ab7FsnSqjcnR3z4snRaCwuAIxHIVDoshAUiAA

src/pagination.css.ts
aweell: +1, buttonLink hover and pressed behaviour should not be altered.

src/pagination.tsx
```diff
@@ -0,0 +1,362 @@
...
+    disabled = false,
+    dataAttributes,
+    'aria-label': ariaLabel,
+}: PaginationProps): JSX.Element | null => {
+    const isControlled = currentPage !== undefined;
+    const [internalPage, setInternalPage] = React.useState(defaultPage);
+    const {texts, t} = useTheme();
+    const variant = useThemeVariant();
+    const {isMobile} = useScreenSize();
+    const hidePageList = hidePageListProp === true;
```
Marcosld: This var is not needed

src/pagination.tsx
```diff
@@ -0,0 +1,362 @@
...
+    const resolvedAriaLabel = ariaLabel ? `${sectionLabel}, ${ariaLabel}` : sectionLabel;
+    const resolvedPrevLabel = navLeftLabel || texts.paginationPrevPage || t(tokens.paginationPrevPage);
+    const resolvedNextLabel = navRightLabel || texts.paginationNextPage || t(tokens.paginationNextPage);
+    const resolvedPrevAriaLabel =
+        navLeftLabel || texts.paginationPrevPageAriaLabel || t(tokens.paginationPrevPageAriaLabel);
+    const resolvedNextAriaLabel =
+        navRightLabel || texts.paginationNextPageAriaLabel || t(tokens.paginationNextPageAriaLabel);
+
+    const goToPage = (page: number) => {
+        const nextPage = Math.min(Math.max(page, 1), totalPages);
```
Marcosld: If we are going to keep these Math.min(Math.max(...)) clampings I would keep it as a `clamp` function. I see we already use it for meter component. Perhaps we can extract it to utils.

src/pagination.css.ts
```diff
@@ -0,0 +1,331 @@
...
+        boxSizing: 'border-box',
+
+        '@media': {
+            [mq.desktopOrBigger]: {
+                gap: 8,
+                padding: '8px 16px',
+            },
+        },
+    },
+]);
```
Marcosld, line 25: Something like this should suffice:

```suggestion
export const container = style([
    sprinkles({
        display: 'inline-flex',
        alignItems: 'center',
        rowGap: {
            mobile: 4,
            desktop: 8,
        },
        paddingY: 8,
        paddingX: {
            mobile: 0,
            desktop: 16,
        },
        maxWidth: '100%',
    }),
    {
        width: 'fit-content',
        boxSizing: 'border-box',
    },
]);
```

Same applies for all styles below, we should try to use sprinkles whenever available because it reduces generated css clases.

src/pagination.css.ts
```diff
@@ -0,0 +1,331 @@
...
+    media: {},
+});
+
+export const navigationButtonLink = style({
+    flexShrink: 0,
+});
+
+export const navigationIconButton = style({
+    flexShrink: 0,
+});
```
Marcosld: I'd say we don't need this... I don't think there's a problem with buttonlink showing ellipsis in small widths instead of overflowing

src/pagination.css.ts
```diff
@@ -0,0 +1,347 @@
...
+globalStyle(
+    `${navigationButtonLink} ${buttonStyles.smallTouchableArea}[disabled]:active ${buttonStyles.smallTouchableVisual}`,
+    {
+        backgroundColor: 'transparent',
+    }
+);
+
+export const navigationButtonLinkVariants = styleVariants({
+    default: {},
+    brand: {
```
Marcosld: This can be removed I think. buttonLinks should behave by default

src/pagination.tsx
```diff
@@ -0,0 +1,362 @@
...
+        <Text3 as="span" weight={textPresets.button.weight} color="currentColor" wordBreak={false}>
+            {children}
+        </Text3>
+    );
+};
+
+type PageListProps = {
+    items: Array<PaginationItem>;
+    disabled?: boolean;
+    className?: string;
```
Marcosld, line 164: Unused prop

src/pagination.tsx
```diff
@@ -0,0 +1,362 @@
...
+
+    return (
+        <ol className={classnames(styles.pageList, className)}>
+            {items.map((item, index) => {
+                if (item.type === 'ellipsis') {
+                    return (
+                        <li
+                            key={`ellipsis-${index}`}
+                            className={classnames(
+                                styles.pageListItemEllipsis,
```
Marcosld: This class is the same than `pageListItem`, perhaps we can simplify

src/pagination.tsx
```diff
@@ -0,0 +1,362 @@
...
+                                styles.ellipsisVariants[variant]
+                            )}
+                            aria-hidden="true"
+                        >
+                            <span className={classnames(styles.ellipsis, styles.ellipsisVariants[variant])}>
```
Marcosld: Is this span and its classes needed at all?

src/pagination.css.ts
```diff
@@ -0,0 +1,347 @@
...
+        width: 'auto',
+        minWidth: 48,
+        padding: '0 12px',
+
+        selectors: {
+            '&&': {
+                width: 'auto',
+                minWidth: 48,
+                padding: '0 12px',
+            },
```
Marcosld, line 116: why is this repeated? Is the width and minwidth needed?

src/pagination.css.ts
```diff
@@ -0,0 +1,347 @@
...
+        },
```
Marcosld: Why is this repeated?

src/__stories__/pagination-story.tsx
```diff
@@ -0,0 +1,106 @@
...
+Default.args = {
+    totalPages: 9,
+    defaultPage: 1,
+    surroundingPageCount: 1,
+    hideNavigationControls: false,
+    disabled: false,
+    navigationControls: 'buttonLink',
+    variantOutside: 'default',
+    onChange: 'none',
+};
```
Marcosld, line 110: I'd set a default value for the rest of the params too so it appears in storybook by default

src/__tests__/pagination-test.tsx
```diff
@@ -0,0 +1,280 @@
...
+            {type: 'ellipsis'},
+            {type: 'page', page: 30, current: false},
+            {type: 'page', page: 31, current: true},
+            {type: 'page', page: 32, current: false},
+            {type: 'ellipsis'},
+            {type: 'page', page: 40, current: false},
+        ]);
+    });
+
+    test('allows configuring more surrounding pages on desktop', () => {
```
Marcosld: this one can be removed

src/__acceptance_tests__/pagination-acceptance-test.tsx
```diff
@@ -0,0 +1,126 @@
...
+        expect(box.width).toBeGreaterThanOrEqual(expectedBox.minWidth);
+    }
+    if (expectedBox.height !== undefined) {
+        expect(box.height).toBe(expectedBox.height);
+    }
+
+    return box;
+};
+
+test('Pagination interactive items have 48px targets on mobile', async () => {
```
Marcosld: Can we test this by clicking with an offset and checking it works, perhaps something like: `
    page.click({
        offset: {x: 47, y: 47}
    })`
could work? That way we don't check implementation detail

src/__acceptance_tests__/pagination-acceptance-test.tsx
```diff
@@ -0,0 +1,126 @@
...
+    for (const target of [page, currentPage]) {
+        await expectBoundingBox(target, {width: MOBILE_TARGET_SIZE, height: MOBILE_TARGET_SIZE});
+    }
+
+    for (const target of [previous, next]) {
+        await expectBoundingBox(target, {minWidth: MOBILE_TARGET_SIZE, height: MOBILE_TARGET_SIZE});
+    }
+});
+
+test('Pagination ellipses are non-interactive and keep the mobile layout size', async () => {
```
Marcosld: Same, perhaps try clicking or tabbing. Size should be covered by screenshots

src/__acceptance_tests__/pagination-acceptance-test.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+const cyberColors = getCyberSkin().colors;
+const ELLIPSIS_COLOR_CASES: ReadonlyArray<{variantOutside: StoryArgs['variantOutside']; color: string}> = [
+    {variantOutside: 'default', color: getRgbColor(cyberColors.textSecondary)},
+    {variantOutside: 'brand', color: getRgbColor(cyberColors.textButtonSecondaryBrand)},
+    {variantOutside: 'negative', color: getRgbColor(cyberColors.textButtonSecondaryNegative)},
+    {variantOutside: 'alternative', color: getRgbColor(cyberColors.textSecondary)},
+];
+
+test.each(ELLIPSIS_COLOR_CASES)('Pagination ellipsis color in $variantOutside context', async (testCase) => {
```
Marcosld: Probably should be covered by screenshots

src/__acceptance_tests__/pagination-acceptance-test.tsx
```diff
@@ -0,0 +1,126 @@
...
+    const ellipsis = await pagination.$('li[aria-hidden="true"] span');
+
+    if (!ellipsis) {
+        throw Error('Ellipsis should be available');
+    }
+
+    expect(await ellipsis.evaluate((element) => getComputedStyle(element).color)).toBe(testCase.color);
+});
+
+test('Pagination hides the page list when configured', async () => {
```
Marcosld: Same, screenshots

Marcosld (commented):
Remember to add doc about this component in doc files. At least we should have some examples using it so llms learn. Check components.md and patterns.md files.

src/pagination.css.ts
AlexandraGallipoliRodrigues: The duplicated declarations were removed. The remaining && block is still needed to override the shared interactiveArea styles (width: 100%; height: 100%). minWidth keeps the required target size, while width: auto and the horizontal padding allow long page numbers, like 1200, to grow without clipping.

src/pagination.tsx
AlexandraGallipoliRodrigues: This is used now to pass the desktop/mobile visibility classes to each PageList, so we can render both lists and switch them with CSS instead of viewport JS. But this will be changed in another issue, right?

src/__stories__/pagination-story.tsx
AlexandraGallipoliRodrigues: Done, I’ve added default values for the Storybook args so the controls show the component state by default. I left currentPage as undefined intentionally, so the default story remains uncontrolled and interactive through defaultPage; setting currentPage by default would make it controlled without local state.

src/__acceptance_tests__/pagination-acceptance-test.tsx
```diff
@@ -47,12 +45,14 @@ test('Pagination ellipses are non-interactive', async () => {
...
+    const ellipsisPoint = await getCenterBetweenElements(
+        await screen.findByRole('button', {name: 'Página anterior'}),
+        previousPage
+    );
 
-    await ellipses[0].click();
+    await page.mouse.click(ellipsisPoint.x, ellipsisPoint.y);
 
-    await screen.findByRole('button', {name: 'Página 10, página actual'});
+    expect(await currentPage.evaluate((element) => element.getAttribute('aria-current'))).toBe('page');
```
Marcosld: Perhaps we can use a query like this: https://testing-library.com/docs/queries/byrole/#current and just check the element is present with the text we want.

src/__acceptance_tests__/pagination-acceptance-test.tsx
```diff
@@ -47,12 +45,14 @@ test('Pagination ellipses are non-interactive', async () => {
...
-    const pagination = await screen.findByRole('navigation');
-    const ellipses = await getVisibleElements(await pagination.$$('li[aria-hidden="true"]'));
-
-    expect(ellipses).toHaveLength(2);
+    const previousPage = await screen.findByRole('button', {name: 'Ir a la página 9'});
+    const currentPage = await screen.findByRole('button', {name: 'Página 10, página actual'});
+    const ellipsisPoint = await getCenterBetweenElements(
+        await screen.findByRole('button', {name: 'Página anterior'}),
+        previousPage
+    );
```
Marcosld: Maybe a simple findByText (with `...` as text) + click ?

Marcosld (approved):

PR #1557, opened 2026-06-02
feat(Chore): package mistica-web as a Claude Code plugin

ieduardogf:
## Summary

- Add `.claude-plugin/plugin.json` manifest so the repo can be installed as a Claude Code plugin from the marketplace
- Clean up `skills/mistica-react/SKILL.md` frontmatter — removed `license`, `metadata.author`, and `metadata.version` fields (those now live in the plugin manifest where they belong) and kept only `name` and `description` per the standard plugin skill format

## How to install

Once merged, the plugin will be installable in any Claude Code session:

```
/plugin marketplace add Telefonica/mistica-web
/plugin install mistica-web
```

This makes the `mistica-react` skill available as `/mistica-web:mistica-react` for anyone building UIs with `@telefonica/mistica`.

## Test plan

- [ ] Verify `.claude-plugin/plugin.json` is valid JSON with required fields
- [ ] Verify `skills/mistica-react/SKILL.md` frontmatter only has `name` and `description`
- [ ] Test plugin install via `/plugin marketplace add` + `/plugin install` in a Claude Code session

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1564, opened 2026-06-04
chore(Chore): move Claude Code plugin into plugin/ subdirectory

Marcosld:
## Summary

The mistica-web git repository is ~3GB, so installing the Claude Code plugin added in #1557 by cloning the full repo fails (or is impractically slow). Claude Code marketplaces support a [`git-subdir` plugin source](https://code.claude.com/docs/en/plugin-marketplaces#git-subdirectories) that does a sparse, partial clone (`--filter=tree:0` + cone sparse-checkout) of just one subdirectory — but it requires the plugin to live in a subdirectory rather than at the repo root.

- Move `.claude-plugin/` and `skills/` into `plugin/` so marketplaces can reference the plugin as `{ "source": "git-subdir", "path": "plugin" }`, downloading **~3MB instead of ~3GB**
- Align `plugin.json` with the ai-sdlc-marketplace entry: `name` must match the marketplace plugin name (`mistica-web`), `displayName` carries the human-readable title, `repository` uses the string form, and the `figma` plugin is declared as a dependency

A companion PR in `Telefonica/ai-sdlc-marketplace` switches the marketplace entry to the `git-subdir` source. **Merge order**: this PR first, then update the marketplace PR's pinned `sha` to this PR's squash-merge commit before merging it (the pre-merge branch sha becomes unreachable after squash + branch deletion).

## Test plan

- [x] Sparse partial clone from GitHub with cone `plugin` downloads ~3MB (verified against this branch)
- [x] `claude plugin install mistica-web` from a test marketplace using `git-subdir` + `path: "plugin"` installs in ~5s; installed cache contains exactly `plugin.json` + `skills/mistica-react/SKILL.md`
- [x] `plugin.json` is valid JSON; `name` matches the marketplace entry
- [x] Test npx skills add

🤖 Generated with [Claude Code](https://claude.com/claude-code)


skills
```diff
@@ -0,0 +1 @@
+plugin/skills
```
Marcosld, line 1: We want npx skills add to keep working as usual

raycoarana (approved):

PR #1571, opened 2026-06-11
chore(Storybook): Vivo community group

yceballost:


yceballost:
can we remove example story? o only show in local deploys @Marcosld 

https://mistica-2vbr1npvk-flows-projects-65bb050e.vercel.app/?path=/story/community-examplecomponent--default

Marcosld:
> can we remove example story? o only show in local deploys @Marcosld
> 
> https://mistica-2vbr1npvk-flows-projects-65bb050e.vercel.app/?path=/story/community-examplecomponent--default

Yes, remove the top level one. I think the one in cyber folder should be removed by them when they push a component

Marcosld (approved):

PR #1572, opened 2026-06-12
fix(Menu): Change menu interaction with checkbox items

yceballost:


Marcosld (approved):

src/menu.tsx
```diff
@@ -153,7 +153,6 @@ export const MenuItem = ({
                 onChange={() => {
                     if (isMenuOpen && itemIndex !== null) {
                         onPress?.(itemIndex);
-                        closeMenu();
```
brtbrt, line 156: can you add a test about this? with AI should suffice. this way, the test can be used as a contract of the behavior (plus we take advantage of the regression tests!)

PR #1581, opened 2026-06-23
fix(Inline): Problem with alignItems center

yceballost:
This PR fix a problem with center alignments in Inline component. You can see the screenshots diff to see the improvement


src/switch-component.css.ts
```diff
@@ -118,6 +118,7 @@ export const container = style([
     containerBase,
     {
         cursor: 'pointer',
+        display: 'flex',
```
yceballost, line 121: image

This change resolve this problem, also improve the ios switch screenshots

src/inline.tsx
```diff
@@ -131,7 +131,18 @@ const Inline = ({
...
-            style={{...applyCssVars(calcInlineVars(space, verticalSpace)), alignItems}}
+            style={{
+                ...applyCssVars({
+                    ...calcInlineVars(space, verticalSpace),
+                    ...(alignItems !== 'stretch'
+                        ? {
+                              [styles.vars.childDisplay]: 'flex',
+                              [styles.vars.childAlignItems]: alignItems,
+                          }
+                        : {}),
```
Marcosld: This logic should live in the `calcInlineVars` function

src/inline.css.ts
```diff
@@ -122,6 +127,11 @@ globalStyle(`${inline} > div`, {
             padding: 0, // restore
             // negative gap is not supported in flexbox, so we use negative margins instead
             margin: `0 0 0 calc(min(${space}, 0px))`,
+            // Propagate alignItems to wrapper divs so icon SVGs (inline elements) are
+            // correctly centered within each wrapper, not just the wrappers themselves.
+            // Only activates when childDisplay is set (i.e. alignItems !== 'stretch').
+            display: fallbackVar(childDisplay, 'block'),
+            alignItems: fallbackVar(childAlignItems, 'normal'),
```
Marcosld: I believe we fall back to block/normal to preserve current behavior when alignItems is `stretch`, but current behavior is probably wrong. I think we should fallback to `flex`/`stretch`: We could remove the conditional in the inline.tsx component and the fallback function here.

yceballost: I tried to use flex / stretch values for fallback, but some screenshot breaks

inline cases, the vertical align change in this case
image

the progressBar stepped disappear
image

but we have an enhance with flex/stretch in rating!
image

But i think the first case is quite breaking if we go by this way. wdyt?

src/__screenshot_tests__/__image_snapshots__/checkbox-screenshot-test-tsx-checkbox-custom-render-1-snap.png
yceballost, line 1: music for my eyes

src/inline.css.ts
Marcosld: Yeah... Perhaps we can fix it in the future

Marcosld (approved):

src/community/__screenshot_tests__/__image_snapshots__/advanced-data-card-screenshot-test-tsx-advanced-data-card-slots-1-1-snap.png
yceballost, line 1: this is expected, we are showing more space than 8px

Image

brtbrt (approved):

PR #1612, opened 2026-07-01
fix(Select): fix Select validation on desktop devices Ref: WEB-2450

### LINKED ISSUE #1613: fix(Select): fix Select validation on desktop devices
## Problem

A mandatory `Select` (without the `optional` prop) does not display the "this field is required" error after submitting a `Form` on the desktop (non-native) variant.

## Root cause

Introduced in **16.45.0** (commit `d2cf20071`, #1464). That change removed the native `required` DOM attribute from desktop inputs and moved required-detection into the field registration (`required ?? input.required` in `form.tsx`). `useFieldProps` was updated to register `required: !optional`, but `Select` uses its own `register` call and was never updated. On desktop the underlying input no longer reports `input.required`, so `validateFields` never flagged the field as mandatory.

The native/mobile variant was unaffected because it renders a real `<select required>` element, so `input.required` stayed `true`.

## Fix

Pass `required: !optional` in `Select`'s `register` call (aligning it with `useFieldProps`), and add `optional` to the effect's dependency array.

Fixes #1612

andyindahouse:
Closes #1613

## Summary
Fixes a validation regression where a mandatory `Select` (no `optional` prop) did
not display the "this field is required" error after submitting a `Form` on the
desktop (non-native) variant.
## Root cause
Introduced in **16.45.0** (commit `d2cf20071`, #1464). That change removed the
native `required` DOM attribute from desktop inputs and moved required-detection
into the field registration (`required ?? input.required` in `form.tsx`).
`useFieldProps` was updated to register `required: !optional`, but `Select` uses
its own `register` call and was never updated. On desktop the underlying input no
longer reports `input.required`, so `validateFields` never flagged the field as
mandatory.
The native/mobile variant was unaffected because it renders a real
`<select required>` element, so `input.required` stayed `true`.
## Fix
Pass `required: !optional` in `Select`'s `register` call (aligning it with
`useFieldProps`), and add `optional` to the effect's dependency array.

Before:
Screenshot 2026-07-01 at 12 31 35

Now:
Screenshot 2026-07-01 at 12 31 51


src/__screenshot_tests__/form-screenshot-test.tsx
```diff
@@ -0,0 +1,29 @@
+import {openStoryPage, screen} from '../test-utils';
+
+test('Form fields after submit validation', async () => {
```
brtbrt: can we be more expressive on the description? Like the purpose of the test itself would be testing the "required" error message right? I'd describe that

src/__screenshot_tests__/form-screenshot-test.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+    await (await screen.findByText('Send')).click();
+
+    const fieldWrapper = await screen.findByTestId('form-wrapper');
+    const image = await fieldWrapper.screenshot();
+
+    expect(image).toMatchImageSnapshot();
+});
+
+test('Form fields after submit validation on desktop', async () => {
```
brtbrt: can we merge the two tests into a single one, by exploiting `.each`?

brtbrt (approved):

PR #1614, opened 2026-07-01
chore(ci): auto-label new issues with to-groom

brtbrt:
## Summary

Adds a GitHub Actions workflow that automatically applies the `to-groom` label to every newly opened issue.

## Details

- Triggers on `issues: opened`, so the label is applied immediately when an issue is created.
- Uses `actions/github-script` with `issues: write` permission to call the GitHub API.

## Prerequisite

The `to-groom` label must exist in the repository (Settings → Labels) for the workflow to apply it successfully.


.github/workflows/label-new-issues.yml
```diff
@@ -0,0 +1,21 @@
...
+            issues: write
+        steps:
+            - uses: actions/github-script@v7
+              with:
+                  script: |
+                      await github.rest.issues.addLabels({
+                        owner: context.repo.owner,
+                        repo: context.repo.repo,
+                        issue_number: context.issue.number,
+                        labels: ['to-groom'],
```
Marcosld: why to-groom? what does this mean?

brtbrt: that's the standard term in agile management environments https://www.atlassian.com/agile/project-management/backlog-grooming

brtbrt: it means "you have this new guy in the backlog, but you haven't checked if it's in a good shape to be taken, it's not prioritized among the other issues we have already (maybe it's super important, maybe the latest concern ever)

Marcosld: I see, I didn't know the term. I'm more used to `triage` in the context of github issues. It is the one suggested in the example action to add labels to new issues: https://docs.github.com/en/actions/tutorials/manage-your-work/add-labels-to-issues#introduction.
If you prefer to use agile terms I think `to-refine` would be a more familiar one. In the article you shared it says `grooming` is and old term currently replaced by `refining`.

brtbrt: no prob, I can use the one that better fits our domain!

Marcosld (approved):

brtbrt:
we're not gonna proceed with this project organization

PR #1658, opened 2026-07-16
feat(release): enable ! notation for breaking changes in semantic-release

### LINKED ISSUE #1657: fix: semantic-release not recognizing ! notation for breaking changes
## Summary

Fix the `!` notation for breaking changes in commit subjects (`feat(scope)!: description`), which used to be validated on PRs, but ignored as a major when releasing.

## The whys

1. `ci` workflow is using `commit-analyzer` 13.0.1 (the action is using it). `uses: amannn/action-semantic-pull-request@v5`
2. `release` was using another one: (10.0.0?) which was buggy about this

The two workflows used to work with different semantics, risky.

### Core Fix
- Upgrade `@semantic-release/commit-analyzer` from 10.0.0 to 13.0.1 (includes conventional-commits-parser 6.4.0)
- Add explicit `parserOpts.breakingHeaderPattern` configuration to enable `!` notation parsing
- Update release workflow to pass NPM token to validate job for proper dry-run

### Testing & Verification
- Verified `!` notation correctly triggers major version bumps (e.g., 17.1.0 → 18.0.0 for `feat(ci)!:`)
- Tested with semantic-release dry-run to confirm breaking change detection works

Dry-run output showed: `The next release version is 18.0.0`

## Related Issues

- Fixes #1611 (major semantic fix needed)
- Tracks #1660 (upstream issue - TODO to remove breakingHeaderPattern when fixed)

## Migration

Breaking changes can now use the shorter syntax:
```
feat(api)!: remove legacy endpoint
```

Instead of:
```
feat(api): remove legacy endpoint

BREAKING CHANGE: legacy endpoint removed
```

Both syntaxes still work. The `!` notation is now the preferred approach.

## Testing

- ✅ Verified `feat(ci)!: ...` correctly detected as major version bump
- ✅ Verified dry-run shows correct next version
- ✅ Verified CI validates PR titles with semantic-release action

brtbrt:
## Summary

Fix the `!` notation for breaking changes in commit subjects (`feat(scope)!: description`), which used to be validated on PRs, but ignored as a major when releasing.

## The whys

1. `ci` workflow is using `commit-analyzer` 13.0.1 (the action is using it) . `uses: amannn/action-semantic-pull-request@v5`
2. `release` was using another one: (10.0.0?) which was buggy about this

the two workflows used to work with different seantics, risky.

### Core Fix
- Upgrade `@semantic-release/commit-analyzer` from 10.0.0 to 13.0.1 (includes conventional-commits-parser 6.4.0)
- Add explicit `parserOpts.breakingHeaderPattern` configuration to enable `!` notation parsing
- Update release workflow to pass NPM token to validate job for proper dry-run

### Testing & Verification!
- Verified `!` notation correctly triggers major version bumps (e.g., 17.1.0 → 18.0.0 for `feat(ci)!:`)
- Tested with semantic-release dry-run to confirm breaking change detection works

https://github.com/Telefonica/mistica-web/actions/runs/29528124907/job/87721546071
https://github.com/Telefonica/mistica-web/actions/runs/29513716048/job/87673343062

```
[4:03:02 PM] [semantic-release] › ℹ  The next release version is 18.0.0
```

## Related Issues

Fixes #1657 
Tracks #1660 (upstream issue - TODO to remove breakingHeaderPattern when fixed)

## Migration

Breaking changes can now use the shorter syntax:
```
feat(api)!: remove legacy endpoint
```

Instead of:
```
feat(api): remove legacy endpoint

BREAKING CHANGE: legacy endpoint removed
```

Both syntaxes still work. The `!` notation is now the preferred approach.

## Testing

- ✅ Verified `feat(ci)!: ...` correctly detected as major version bump
- ✅ Verified dry-run shows correct next version
- ✅ Verified CI validates PR titles with semantic-release action

closes #1657


Marcosld:
I would not add a forced dry run validation. I'd just add a checkbox into the release job to perform a dry run.

brtbrt:
## Update: Custom releaseRules Configuration

Added explicit  to the  plugin to ensure breaking changes are properly recognized.

### Why this is needed

The angular preset alone does not reliably recognize the `feat(scope)!` notation for breaking changes (see #3757, #231). By adding explicit releaseRules with `{ breaking: true, release: 'major' }`, we ensure that commits marked as breaking (by the parser recognizing the `!` notation) will trigger major releases.

### What was added

```json
"releaseRules": [
  { "breaking": true, "release": "major" },
  { "revert": true, "release": "patch" },
  { "type": "feat", "release": "minor" },
  { "type": "fix", "release": "patch" },
  { "type": "perf", "release": "patch" },
  { "scope": "no-release", "release": false }
]
```

The key rule is the first one: when a commit's `breaking` flag is true (set by conventional-commits-parser@6+ when it sees `!`), it triggers a major release.

.github/workflows/release.yml
```diff
@@ -13,13 +13,12 @@ permissions:
     pull-requests: write
 
 jobs:
-    release:
-        name: Release
+    validate:
```
Marcosld: I think you must add the manual approval here too as it uses both npm and github tokens @Yermanaco

.github/workflows/release.yml
```diff
@@ -13,13 +13,12 @@ permissions:
     pull-requests: write
 
 jobs:
```
Marcosld, line 20: IMO this is overkill and not needed. I would add a checkbox on the github action to perform a dry run if you want to

package.json
```diff
@@ -79,6 +79,7 @@
         "@axe-core/puppeteer": "^4.4.2",
         "@azure/storage-blob": "^12.9.0",
         "@semantic-release/changelog": "^6.0.1",
+        "@semantic-release/commit-analyzer": "^13.0.1",
```
Marcosld: normally you would use resolutions for this... On the other hand.. why is this needed?

package.json
```diff
@@ -190,7 +191,37 @@
...
+                        },
+                        {
+                            "type": "perf",
+                            "release": "patch"
+                        },
+                        {
+                            "scope": "no-release",
+                            "release": false
+                        }
+                    ]
```
Marcosld: I think it is a good idea to allow ! as a pattern for major versions, but I don't think we need all this custom logic. Just use https://conventional-changelog.js.org/presets/conventional-commits/ like suggested here: https://github.com/semantic-release/semantic-release/issues/2339, which already supports !.

Marcosld (changes requested):

.github/workflows/release.yml
brtbrt: the trade off is critical here: double-checking a release process, which is run at most once-twice a week, is way cheaper than unpublish-deprecate a package and checking why it failed.

- in case you expect a minor and a major is about to release: maybe something you didn't want has been integrated?
- in case you expect a major and a minor is released: people could use it automatically breaking a lot of stuff

if it were me yesterday, I greatly would have appreciated this stop point

.github/workflows/release.yml
brtbrt: yep definitively. done

package.json
brtbrt: we're using ci and release  workflow with different versions. This could lead to unexpected unmatching results, like we've just experienced with `!`

I explained it better on the summary.
anyway, yes, we could move into resolution: my bad here

package.json
brtbrt: yep, tested and we can simplify this defensive conf.

brtbrt:
> I would not add a forced dry run validation. I'd just add a checkbox into the release job to perform a dry run.
I like it, but I'd run it twice each time? it'd be cumbersome and even more error-prone

.github/workflows/release.yml
Marcosld: Why don't you add a checkbox in the github action run dialog to trigger dry runs? Or you can even make the current cliArgs text input work: 
image
and add `--dry-run` when you want to check.

package.json
Marcosld: why don't you just update semantic-release package?

brtbrt: we'd end up with the same problem in the end: checking the transitional dependencies would be a nightmare

.github/workflows/release.yml
brtbrt: that's exactly the opposite of I wanna do: securing the thing. We can do the opposite by passing a "skip check release version" option

brtbrt: Image

.github/workflows/ci.yml
```diff
@@ -118,6 +118,7 @@ jobs:
                       feat
                       chore
                       revert
+                  requireScope: false
```
Marcosld, line 121: why this change?

.github/workflows/release.yml
```diff
@@ -6,20 +6,26 @@ on:
                 description: 'CLI args'
                 required: false
                 default: ''
+            skipValidation:
```
Marcosld, line 9: I would change this the other way around (checkbox to enable dryrun). You can even simplify and have just 1 parametrizable job (the user can run it with dry run enabled or not, it doesn't have 2 steps)

package.json
```diff
@@ -190,7 +190,11 @@
             [
                 "@semantic-release/commit-analyzer",
                 {
-                    "preset": "angular"
+                    "preset": "angular",
+                    "_comment": "TODO #1660: Remove breakingHeaderPattern once semantic-release angular preset supports ! notation natively",
+                    "parserOpts": {
+                        "breakingHeaderPattern": "^(\\w*)(?:\\((.*)\\))?!: (.*)$"
+                    }
```
Marcosld, line 197: You resolved my previous comment but didn't change the preset to conventional-commits

Marcosld (changes requested):

.github/workflows/ci.yml
brtbrt: Makes the scope optional (`feat: …` and `feat(button): …` both pass). It is the action's default.
I set it explicitly to document the intent and stay consistent with semantic-release, which does not require a scope either.

.github/workflows/release.yml
brtbrt: Happy to collapse it into a single parametrizable job, but not to invert the flag. Validation stays default-on. An opt-in safety check is one that gets forgotten exactly when it matters. `skipValidation` is a deliberate opt-out, not the other way around.
better safe than sorry.

package.json
brtbrt: Good catch — the analyzer alone was not enough. release-notes-generator parses commits independently, so without the same override a `feat!: …` bumps the major version but still renders as a plain feature in the CHANGELOG. I have applied `breakingHeaderPattern` to both plugins, so `!` is now handled consistently.

We are staying on the angular preset to keep our existing changelog format. `breakingHeaderPattern` is the only gap for `!` support, and it is now covered in both places. TODO #1660 tracks removing the overrides once angular supports ! natively. Resolving this.

.github/workflows/release.yml
Marcosld: The only case when this really matters is when you publish a breaking change as part of a minor/patch version. Don't make everyone go through 2 steps for something that happened once in the last 2 years.

package.json
Marcosld: Why do you resolve the conversation?

That's probably because in the config we are setting preset to angular in the release notes generator config and you forgot to change it too:
image


You can set it top level according to the doc:
https://semantic-release.gitbook.io/semantic-release/usage/plugins#plugin-options-configuration
image

Marcosld (changes requested):

package.json
brtbrt: if you want to transition from `angular` to `conventional-commits`: we can do it. 
I avoided it because it means further (little) changes I didn't want to introduce. I just wanted the `!` to be working, without altering anything else.

Aspect | angular | conventionalcommits
-- | -- | --
Release heading | # for minor and major, ## for patch | always ##
Breaking section | at the bottom, ### BREAKING CHANGES | at the top, ### ⚠ BREAKING CHANGES
Section order | Bug Fixes, then Features | Features, then Bug Fixes
feature: type | hidden | shown under Features, but it still does not bump
Release-As: footer | ignored | forces the commit into the changelog, but does not change the bump

if we agree on these little changes, we can perfectly switch to that preset that is supposed to be working as we expect

.github/workflows/release.yml
brtbrt: I did want to make everyone go through 2 steps:
- the action is not triggered daily, hourly etc. you can assume 5 more seconds
- it delivers added value by checking unexpected: 
  - minor being major
  - major being minor
  a hidden `BREAKING CHANGEs` delivered on a commit could make you have unexpected releases way too late.  (the squash commit carries the note through the merge dialog)

but those are simply my two cents on it, and shared to clarify why I foster this.
If the team agrees on your point and it's valued heavier than my points, let's proceed with the optional way

package.json
Marcosld: I would just discard this PR. The documentation seems clear enough: https://conventional-changelog.js.org/presets/angular/#breaking-changes.

It seems like we already have: 
Image
to add `--dry-run` arg, which should suffice.

brtbrt:
@Marcosld decided to not proceed with this: 
- major must be declared on the body of the merge
- `!` is ignored
  - `!` on the title does not invalidate the title

PR #1662, opened 2026-07-17
feat(skin): support responsiveLayoutMargin design tokens 

### LINKED ISSUE #1611: Enable responsiveLayoutMargin token
Blocks https://github.com/Telefonica/mistica-design/pull/2662 

Check if PR can be merged and update what is needed to allow using this token in ResponsiveLayout component

brtbrt:
- `resopnsiveLayoutMargin` has been introduced
- skins have been imported by https://github.com/Telefonica/mistica-design/pull/2662
- package has been "improved" to let it run more easily by local command line

# Todo
- [x] `generate-design-tokens` from workflows, not locally, to check things are ok
- [x] update screenshots it diffs are ok with our changes


packages/generate-design-tokens/index.js
```diff
@@ -150,7 +150,22 @@ export const get${toPascalCase(skinName)}Skin: GetKnownSkin = () => {
                 .join(',')},
         },
         spacing: ${JSON.stringify(
-            Object.fromEntries(Object.entries(designTokens.spacing).map(([name, {value}]) => [name, value]))
+            Object.fromEntries(
+                Object.entries(designTokens.spacing).map(([name, spacingValue]) => {
+                    if (name === 'responsiveLayoutMargin') {
```
Marcosld: Perhaps we can perform a more generic check here

src/skins/types/index.tsx
```diff
@@ -117,6 +119,7 @@ export type SpacingConfig = {
     heroPadding: PaddingYValues;
     headerPadding: PaddingYValues;
     drawerPadding: PaddingValues;
+    responsiveLayoutMargin?: MultiBreakpointValue<string>;
```
Marcosld: Are these strings? Perhaps we should align with brasil to see what breakpoints will they be setting in the end

src/skins/skin-contract.css.ts
```diff
@@ -363,6 +363,7 @@ const spacing: ToThemeTokens<SpacingConfig> = {
     heroPadding: {top: '', bottom: ''},
     headerPadding: {top: '', bottom: ''},
     drawerPadding: {top: '', bottom: '', left: '', right: ''},
+    responsiveLayoutMargin: {mobile: '', tablet: '', desktop: '', largeDesktop: ''},
```
Marcosld: I think this breaks the convention a bit... on the other hand... are you assigning these css vars any values at all? check out `theme-context-provider.tsx`

Marcosld (changes requested):

src/skins/types/index.tsx
brtbrt: @Telefonica/mistica-design what kind of values you wanna admit? Just numbers?

https://github.com/Telefonica/mistica-design/pull/2662/changes

src/skins/skin-contract.css.ts
brtbrt: if you have a better idea of how to shape this I'm happy to change it, because it complicates the typing a lot

about the default values: yeah, but where would we place it? because this is the contract 🤔. I've currently placed them here:

https://github.com/Telefonica/mistica-web/blob/2282d4d6e78da95eeb874ea712b40386ac47ae0a/src/skins/defaults.tsx#L103

packages/generate-design-tokens/css-generator.js
```diff
@@ -306,6 +306,12 @@ export const generateCommonCssSrc = (DESIGN_TOKENS_FOLDER) => {
         )
         .join('\n');
 
+    const responsiveLayoutMargin = designTokens.spacing?.responsiveLayoutMargin || {};
+    const mobileMargin = responsiveLayoutMargin.mobile?.value || '16px';
+    const tabletMargin = responsiveLayoutMargin.tablet?.value || '24px';
+    const desktopMargin = responsiveLayoutMargin.desktop?.value || '40px';
+    const largeDesktopMargin = responsiveLayoutMargin.largeDesktop?.value || 'calc((100vw - 1416px) / 2)';
```
brtbrt: are we ok with this?

packages/generate-design-tokens/index.js
```diff
@@ -150,7 +150,22 @@ export const get${toPascalCase(skinName)}Skin: GetKnownSkin = () => {
...
+                        return [
+                            name,
+                            {
+                                mobile: spacingValue.mobile.value,
+                                tablet: spacingValue.tablet.value,
+                                desktop: spacingValue.desktop.value,
+                                largeDesktop: spacingValue.largeDesktop.value,
+                            },
+                        ];
+                    }
```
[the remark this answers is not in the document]
brtbrt: > The generator assumes designTokens.spacing.responsiveLayoutMargin always exists with mobile/tablet/desktop/largeDesktop entries. 

actually we're all ok with this: it must always be present right?

packages/generate-design-tokens/package.json
```diff
@@ -1,7 +1,10 @@
 {
     "name": "generate-design-tokens",
     "type": "module",
-    "packageManager": "yarn@3.2.4",
+    "packageManager": "yarn@3.6.1",
```
brtbrt, line 4: to sync with root packageManager. this way we can use workspaces

src/community/skins/cyber-skin.tsx
```diff
@@ -1112,6 +1112,7 @@ export const getCyberSkin = (): Skin => {
                     desktop: 40,
                 },
             },
+            responsiveLayoutMargin: {mobile: '16px', tablet: '32px', desktop: '48px', largeDesktop: '64px'},
```
brtbrt: I've added myself manually for cyber, while this is being solved

https://github.com/Telefonica/mistica-design/pull/2662#pullrequestreview-4723962151

.yarn/cache/prettier-npm-3.9.5-568e1c1fd9-a475080228.zip
brtbrt, line 1: used by used by the packages/generate-design-tokens package, which specifies prettier@^3.3.3 as a dependency

yarn.lock
```diff
@@ -16775,6 +16784,15 @@ __metadata:
   languageName: node
   linkType: hard
 
+"prettier@npm:^3.3.3":
```
brtbrt, line 16787: used by packages/generate-design-tokens

src/skins/types/colors.tsx
brtbrt, line 1: this is super confusing. this is auto generated, whether index.tsx is not.

src/skins/skin-contract.css.ts
```diff
@@ -304,11 +325,13 @@ const borderRadii: BorderRadiiConfig = {
...
-    ? string
-    : T extends object
-      ? {[K in keyof T]: ToThemeTokens<T[K]>}
-      : string;
+type ToThemeTokens<T> = T extends {mobile: number} & {tablet: any}
+    ? {[K in keyof T]: ToThemeTokens<T[K]>}
+    : T extends {mobile: number}
+      ? string
+      : T extends object
+        ? {[K in keyof T]: ToThemeTokens<T[K]>}
```
brtbrt: I don't like it too much, but that's what we need to make responsivemarging work properly (if we want it like that?)

src/skins/skin-contract.css.ts
brtbrt: relates: https://github.com/Telefonica/mistica-web/pull/1662/changes#r3604712094

brtbrt:
@Telefonica/mistica-web-reviewers a lot of tests outputs have changed because I performed a skin import to test the thing. and I see margins as well as other things on that branch. 

To do this in a clean manner:
1. I'd import master changes before in a separated branch, to be up to date, except for the new margin token(s)
2. I'd re-perform the import from the dedicated mistica-design branch to check things here, affecting margins only

WDYT?

src/community/skins/cyber-skin.tsx
```diff
@@ -1112,6 +1133,13 @@ export const getCyberSkin = (): Skin => {
                     desktop: 40,
                 },
             },
+            responsiveLayoutMargin: {
```
brtbrt: by hand

brtbrt:
Hey, just wanted to flag the vars change for review. We're now using `skinVars.spacing.responsiveLayoutMargin` directly for the responsive margins instead of hardcoded constants. The values are sourced from the skin tokens and converted to CSS in `responsive-layout.css.ts` using `calc()`. Not 100% sure this is the cleanest approach - could also keep the constants and just have them reference the skin tokens. Thoughts?

src/sheet-common.tsx
```diff
@@ -280,11 +279,11 @@ const Sheet = React.forwardRef<HTMLDivElement, SheetProps>(({onClose, children,
...
 
-const paddingX = {
-    mobile: MOBILE_SIDE_MARGIN,
-    tablet: TABLET_SIDE_MARGIN,
+const getPaddingX = (marginVars: typeof skinVars.spacing.responsiveLayoutMargin) => ({
+    mobile: marginVars.mobile,
+    tablet: marginVars.tablet,
     desktop: 40, // to keep consistency with the rest of the dialogs components
-} as const;
+});
```
brtbrt: are we ok with this? now the would be changing according to the tokens!

src/sheet-common.tsx
```diff
@@ -307,6 +300,13 @@ export const SheetBody = ({
     link,
     children,
 }: SheetBodyProps): JSX.Element => {
+    const {spacing} = useTheme();
```
brtbrt, line 304: http://localhost:6006/?path=/story/components-modals-sheet--default&globals=viewport.value:826-896

now sheets would use the new tokens for margins! should we still use consts as a separated thing? use the new tokens/intine vars? Not sure!

@Telefonica/mistica-design 
@Telefonica/mistica-web-reviewers

annefranca: @yceballost

packages/generate-design-tokens/index.js
brtbrt: how?

css/blau.css
brtbrt, line 1: generated via `yarn workspace generate-design-tokens run generate ../../../mistica-design/tokens` 

`2648-responsive-layout-update-app-50-vivo-new` branch

src/sheet-common.tsx
brtbrt: ping

src/__screenshot_tests__/__image_snapshots__/card-screenshot-test-tsx-card-cover-footer-with-image-background-vivo-evolution-1-snap.png
brtbrt, line 1: slight changes in color by importing tokens

src/theme-context-provider.tsx
```diff
@@ -242,33 +242,49 @@ const ThemeContextProvider = ({theme, children, as, withoutStyles = false}: Prop
...
-                },
-            };
-        });
-        return Object.assign({}, ...tokenValues) as typeof vars.spacing;
-    }, [contextTheme]);
+    const spacingDesktopVars = React.useMemo(
+        () =>
+            Object.fromEntries(
+                Object.entries(contextTheme.spacing).map(([token, values]) => {
+                    if (token === 'responsiveLayoutMargin') {
```
Marcosld: I'd check if `desktop` property exists instead of this literal check

src/theme-context-provider.tsx
```diff
@@ -242,33 +242,49 @@ const ThemeContextProvider = ({theme, children, as, withoutStyles = false}: Prop
...
+                })
+            ) as typeof vars.spacing,
+        [contextTheme]
+    );
+
+    const spacingMobileVars = React.useMemo(
+        () =>
+            Object.fromEntries(
+                Object.entries(contextTheme.spacing).map(([token, values]) => {
+                    if (token === 'responsiveLayoutMargin') {
```
Marcosld: Same here

Marcosld (approved):

src/theme-context-provider.tsx
brtbrt: I could duck-type it, but why would we? The name check is intentional.

Per the token schema, `spacingProperties.value` is an `anyOf`: a scalar `{mobile, desktop}` **or** a directional `{top, right, bottom, left}` where each side is itself `{mobile, desktop}`. So `desktop` is not unique to `responsiveLayoutMargin` — it lives in every spacing token, just nested for the directional ones.

That means `'desktop' in values` really asks "was this authored in scalar form?", not "is this `responsiveLayoutMargin`?". They coincide today, but the schema lets any spacing token be scalar. Example — the day someone adds:

```json
"someGap": { "value": { "mobile": 8, "desktop": 16 }, "type": "spacing" }
```

it silently falls into this branch and is emitted unitless (`${values.desktop}`, no `px`), even though it needs `px`. Only `responsiveLayoutMargin` wants that, because it alone is consumed as `calc(... * 1px)` and declared as a leaf var in the contract.

So the literal check matches on meaning, not shape — more robust, not less. Happy to switch if you still prefer it, but I would rather not hand the next scalar token a footgun.

src/sheet-common.tsx
yceballost: Yes, use the tokens here too.

Are the specs updated with this change @annefranca ?

src/theme-context-provider.tsx
Marcosld: hmmmm just add the `px` suffix here and stop using calc... why are you using calc in the first place?

brtbrt: because I wanted to be consistent with 

https://github.com/Telefonica/mistica-web/blob/cbdf9372a472ff9619a3361909bc7248dfe4b3ea/src/responsive-layout.css.ts#L10

and the unitless

```
export const TABLET_SIDE_MARGIN = 32;
export const LARGE_DESKTOP_SIDE_MARGIN = 64;
export const EXTRA_LARGE_DESKTOP_MAX_WIDTH = 1704;
```
anyway, switching to the branching you suggest. it's perfectly viable

src/sheet-common.tsx
brtbrt: the point is that we're using some refs that are tokens, others that comes like hardcoded consts. Plus `mobile` could come like any int that is not potentially a `PadSize`.

https://github.com/Telefonica/mistica-design/blob/923c8b13f0aad8c5751557e128f08886244921e3/tokens/schema/skin-schema.json#L1216

How do you usually tackle this?

Marcosld: I think the tokens value should be reviewed by design team so they match our conventions. I wouldn't cast to PadSize anywhere in this PR but rely on design reviewing these tokens before they are merged. Perhaps there's some case where they won't match a PadSize and that's okay.

Marcosld (approved):

src/sheet-common.tsx
brtbrt: piece of suggestion: maybe they could update the JSON schema with something like this?

```
{
  "$defs": {
    "padSize": {
      "type": "integer",
      "enum": [0, 2, 4, 8, 12, 16, 20, 24, 32, 40, 48, 56, 64, 72, 80]
    }
  },
  "properties": {
    "mobile":  { "$ref": "#/$defs/padSize" },
    "desktop": { "$ref": "#/$defs/padSize" }
  }
}
```

aweell (approved):

yceballost (approved):

src/sheet-common.tsx
```diff
@@ -280,12 +280,6 @@ const Sheet = React.forwardRef<HTMLDivElement, SheetProps>(({onClose, children,
     );
 });
 
-const paddingX = {
-    mobile: MOBILE_SIDE_MARGIN,
-    tablet: TABLET_SIDE_MARGIN,
-    desktop: 40, // to keep consistency with the rest of the dialogs components
```
aweell, line 286: This change creates an unexpected increase of padding in sheet view in desktop, the value should not be replaced by LARGE_DESKTOP_SIDE_MARGIN.

Discussed offline to create a token for dialogPadding left and right definitions so can be used in both dialogs and sheet desktop view

aweell: https://github.com/Telefonica/mistica-design/issues/2703

PR #1663, opened 2026-07-18
feat: support community skins generation

### LINKED ISSUE #1588: Update generate-design-tokens package to import from community skins
Update:

https://github.com/Telefonica/mistica-web/tree/master/packages/generate-design-tokens

Allowing to import from community/cyber skin introduced in:

https://github.com/Telefonica/mistica-design/pull/2658

brtbrt:
## How the script is organized
The generator is built around one concept: a skin family. Each family is a descriptor object declaring where its tokens live, where its output goes, its list of skins, its file naming, and how its module is shaped (skinModule). 

There are two: 
`OFFICIAL_SKINS_FAMILY` 
and 
`COMMUNITY_SKINS_FAMILY`.

A single generic pipeline, `generateSkinFamily(family)`, does all the work: list the family's skins, generate each `.tsx` and `.css`, then run any family-specific shared files. `main()` just runs it for each family.

**The two families differ only in data**, so official and community skins no longer have separate code paths.

## Related Issues
Fixes #1588


packages/generate-design-tokens/index.js
```diff
@@ -24,8 +24,18 @@ console.log('Using design tokens from:', DESIGN_TOKENS_FOLDER);
 
 const SKINS_FOLDER = path.join(import.meta.dirname, '..', '..', 'src', 'skins');
 const CSS_FOLDER = path.join(import.meta.dirname, '..', '..', 'css');
+const COMMUNITY_SKINS_FOLDER = path.join(import.meta.dirname, '..', '..', 'src', 'community', 'skins');
+const COMMUNITY_CSS_FOLDER = path.join(import.meta.dirname, '..', '..', 'css', 'community');
 
-const KNOWN_SKINS = ['blau', 'movistar', 'o2', 'telefonica', 'vivo', 'vivo-evolution', 'esimflag'];
+const OFFICIAL_SKINS = ['blau', 'movistar', 'o2', 'telefonica', 'vivo', 'vivo-evolution', 'esimflag'];
+const COMMUNITY_SKINS = ['cyber'];
```
Marcosld: Perhaps we can inline these into the skin families constants ?

packages/generate-design-tokens/index.js
```diff
@@ -171,45 +193,134 @@ export type Colors = {
...
+ */
 
-    for (const skinName of KNOWN_SKINS) {
-        console.log('Generating tokens for skin', skinName);
+/**
+ * A skin family describes where its tokens live, where the generated files go, how each skin
+ * module is shaped, and which skins it contains. Known and community skins share the same
+ * generation pipeline and differ only in this descriptor.
+ *
+ * @typedef {{
```
Marcosld: mistica-web already requires node >= 24: 
Image

Could we take advantage of https://nodejs.org/learn/typescript/run-natively#running-typescript-natively and migrate this file to native typescript already? :P

Marcosld (approved):
GJ!

packages/generate-design-tokens/index.ts
```diff
@@ -24,8 +24,18 @@ console.log('Using design tokens from:', DESIGN_TOKENS_FOLDER);
...
-const KNOWN_SKINS = ['blau', 'movistar', 'o2', 'telefonica', 'vivo', 'vivo-evolution', 'esimflag'];
+const OFFICIAL_SKINS = ['blau', 'movistar', 'o2', 'telefonica', 'vivo', 'vivo-evolution', 'esimflag'];
+const COMMUNITY_SKINS = ['cyber'];
+
+const GENERATED_BANNER = `/**
+ * @generated
+ * This file was automatically generated by the generate-design-tokens script.
+ * Do not edit it by hand: your changes will be overwritten the next time the script runs.
+ */
+`;
```
brtbrt, line 40: this is not the best way to do that, but we have a issue to better address how to organize generated files. but way much better than nothing right now!

packages/generate-design-tokens/index.js
brtbrt: they're here for convenience, like a frontmatter style, but actually they don't convince me: is there a good reason I don't know to have the supported skins hardcoded here? 

1. do we have special cases/contexts where we want to exclude upcoming-and-already-published skins?
2. if yes, shouldn't we use an env variable for that? or like a param etc.

packages/generate-design-tokens/index.js
brtbrt: yes, but `engines` is never inherited across manifest boundaries AFAIK. but it's doable if you want to

Marcosld: You should have node >=24 to work in the repo so I'd say it's fine. Taking in account you added all the types in this PR I'd do it with typescript already.

packages/generate-design-tokens/index.js
Marcosld: I don't know, feel free to investigate it. You can move them into the family constants and move the constants up if you want to have them on top.

packages/generate-design-tokens/index.js
brtbrt: if you're sure that a single PR changing the script into TS + changing the behavior is not way too much,  yes, let's proceed

css/blau.css
brtbrt, line 1: testing the generation by using mistica-design `master` locally

tsconfig.json
```diff
@@ -90,5 +90,5 @@
     },
 
     // this also excludes paths from VSCode auto import feature
-    "exclude": ["dist", "dist-ts", "dist-es", "node_modules", "examples"]
+    "exclude": ["dist", "dist-ts", "dist-es", "node_modules", "examples", "packages"]
```
Marcosld, line 93: Hmmm.. shouldn't it be the other way around? Include in dev / exclude in pro tsconfig?

Marcosld (approved):
awsom

PR #1672, opened 2026-07-28
feat(ButtonLink): added ButtonLinkNeutral

### LINKED ISSUE #1582: [Mística] new neutral button link
The Button Link component currently lacks a neutral typology, which limits its use in different contexts. This issue covers the implementation of a new neutral variant for the Button Link component, ensuring it aligns with the existing design system tokens. Design specifications are already available.

Additionally, it is proposed to unify the component under a single type, which would include: ButtonLinkBrand (current default variant), ButtonLinkNeutral (new variant) and ButtonLinkDanger (currently independent variant).
[Buttons Specs](https://www.figma.com/design/koROdh3HpEPG2O8jG52Emh/%F0%9F%94%B8-Buttons-Specs?node-id=0-1085&t=lT2KwRPeUZGohopi-4)

[Design issue](https://github.com/Telefonica/mistica-design/issues/2621?issue=Telefonica%7Cmistica-design%7C2626)

AlexandraGallipoliRodrigues:
[issue](https://github.com/Telefonica/mistica-web/issues/1582)


brtbrt (approved):

Marcosld:
I suppose we didn't unify because it would be a breaking change? Perhaps we can do it and mark with @deprecated the ButtonLinkDanger?

brtbrt:
> I suppose we didn't unify because it would be a breaking change? Perhaps we can do it and mark with @deprecated the ButtonLinkDanger?

that should have been specified in the issue description, actually. if you wanna deprecate things, please add the removal into the next major release https://github.com/Telefonica/mistica-web/issues/1645

AnaMontes11:
> I suppose we didn't unify because it would be a breaking change? Perhaps we can do it and mark with @deprecated the ButtonLinkDanger?

For some context, @Marcosld @brtbrt : from the Design side, we think the button ecosystem needs a proper review. However, this isn't the right time to tackle it, and we don't want to block the PR because of it.

What we're proposing is to take advantage of the introduction of the **neutral** style and add it as a new `type` (ideally, **danger** should follow the same approach). The goal is to converge towards an API like `type: default | neutral | danger`, aligned with `IconButton`.

We believe this is the right time to introduce it this way, rather than creating a separate component that we'd end up deprecating later.

From our perspective, we're happy to mark `ButtonLinkDanger` as deprecated and introduce both `danger` and `neutral` as new `type` values.

Even the `ButtonLink` naming itself is something that could be revisited in the future.

If you want we can have a quick call to discuss it.

Marcosld:
> > I suppose we didn't unify because it would be a breaking change? Perhaps we can do it and mark with @deprecated the ButtonLinkDanger?
> 
> that should have been specified in the issue description, actually. if you wanna deprecate things, please add the removal into the next major release #1645

Well, it is specified in the issue:

> Additionally, it is proposed to unify the component under a single type, which would include: ButtonLinkBrand (current default variant), ButtonLinkNeutral (new variant) and ButtonLinkDanger (currently independent variant).

I'm up to making the change you propose @AnaMontes11 ! That's why I flagged it just in case we forgot

AnaMontes11:
> > > Supongo que no nos unificamos porque sería un cambio radical. Tal vez podamos hacerlo y marcarlo con@deprecated¿El enlace del botón es peligroso?
> > 
> > 
> > Eso debería haberse especificado en la descripción del problema. Si quieres descontinuar funciones, por favor, agrega la eliminación en la próxima versión principal [#1645.](https://github.com/Telefonica/mistica-web/issues/1645)
> 
> Bueno, está especificado en el problema:
> 
> > Además, se propone unificar el componente bajo un único tipo, que incluiría: ButtonLinkBrand (variante predeterminada actual), ButtonLinkNeutral (nueva variante) y ButtonLinkDanger (variante actualmente independiente).
> 
> Estoy dispuesto a hacer el cambio que propones.@AnaMontes11¡Por eso lo marqué, por si acaso lo olvidábamos!

Great @Marcosld thanks for adding it!

aweell:
What's the reason behind this commit? https://github.com/Telefonica/mistica-web/pull/1672/changes/d50494925cb99903dd698ce27f4afff0aeb6a906 

This is manually changing the definition in mistica-design?

https://github.com/Telefonica/mistica-design/blob/production/tokens/blau.json#L2374

AlexandraGallipoliRodrigues:
> What's the reason behind this commit? [d504949](https://github.com/Telefonica/mistica-web/commit/d50494925cb99903dd698ce27f4afff0aeb6a906)
> 
> This is manually changing the definition in mistica-design?
> 
> https://github.com/Telefonica/mistica-design/blob/production/tokens/blau.json#L2374

The reason for that commit was that, with the current generated token values, `ButtonLink type="neutral"` over media/dark contexts didn’t match the Figma spec: it rendered like a regular/default link instead of the neutral button link shown in design, with light background and dark text:
image


That said, you’re right: changing those values directly in `mistica-web` is not ideal if the source of truth is `mistica-design`. I’ll revert this token override here and, should I follow up by updating the token definition from `mistica-design` instead? Or should someone from desing do it instead?

src/button.tsx
```diff
@@ -324,14 +324,20 @@ type ButtonProps = ExclusifyUnion<
...
 
-type ButtonLinkProps = ExclusifyUnion<
+export type ButtonLinkType = 'default' | 'danger' | 'neutral';
+
+type BaseButtonLinkProps = ExclusifyUnion<
     FakeButtonProps | ToButtonProps | OnPressButtonProps | HrefButtonProps
 > & {
     bleedLeft?: boolean;
     bleedRight?: boolean;
     bleedY?: boolean;
```
Marcosld, line 334: Why not just add new "type" prop here?

src/__stories__/button-story.tsx
```diff
@@ -233,17 +267,40 @@ DangerButton.args = defaultArgs;
 LinkButton.args = {
     ...defaultArgs,
     chevron: 'default',
+    type: 'default',
 };
 LinkButtonDanger.args = defaultArgs;
+LinkButtonDanger.parameters = {
+    docs: {
+        description: {
+            story: 'Deprecated. Use ButtonLink with type="danger" instead.',
```
Marcosld: I think we have docs disabled... or at least I cannot find this comment. I think we should remove danger and neutral stories from storybook, it should be enough with the ButtonLink story + `type` control

src/__stories__/button-story.tsx
```diff
@@ -225,6 +258,7 @@ SecondaryButton.storyName = 'ButtonSecondary';
 DangerButton.storyName = 'ButtonDanger';
 LinkButton.storyName = 'ButtonLink';
 LinkButtonDanger.storyName = 'ButtonLinkDanger';
+LinkButtonNeutral.storyName = 'ButtonLink type neutral';
```
Marcosld: I'd remove this story or follow the convention for storyName (`ButtonLinkNeutral`)

Marcosld (approved):

yceballost:
check this documentation for agents to include new props here
https://github.com/Telefonica/mistica-web/blob/master/doc/components.md#button-variants

AnaMontes11 (approved):
good job!

aweell (approved):

PR #1677, opened 2026-07-31
feat(List): add small prop, boxedDefaultPadding token and update divider alignment

MurilloLeoni:
This PR updates `RowList` and `BoxedRowList`.

Changes
- Adds a `small` prop for compact row rendering, reducing text scale across title, subtitle, description and detail.
   https://github.com/Telefonica/mistica-design/issues/2676
- Introduces the `boxedDefaultPadding` token to control horizontal padding, replacing previously hardcoded values.
   https://github.com/Telefonica/mistica-design/issues/2679
- Updates the title `font-weight` via the updated design token from mistica-design.
   https://github.com/Telefonica/mistica-design/issues/2684
- Updates the divider alignment so it starts at the text content area rather than the full row width.
   https://github.com/Telefonica/mistica-design/issues/2678

Related
- Token PR in mistica-design: https://github.com/Telefonica/mistica-design/pull/2698

Spec

- Updated Lists specification:
https://www.figma.com/design/Be8QB9onmHunKCCAkIBAVr/%F0%9F%94%B8-Lists-Specs?node-id=0-1


brtbrt:
I would have split into several PRs to ease focusing on the thing you're changing, and having a 1-1 relation with the subtasks

https://github.com/Telefonica/mistica-design/issues/2675

aweell (commented):
Should screenshot examples for small list variant be included?

src/list.tsx
```diff
@@ -34,6 +35,11 @@ import type {TouchableElement, TouchableProps} from './touchable';
 import type {DataAttributes, TrackingEvent, IconProps} from './utils/types';
 import type {ExclusifyUnion} from './utils/utility-types';
 
+const RowListDividerContext = React.createContext(false);
+
+type ListContextType = {small: boolean};
```
brtbrt, line 38: not sure if we wanna introduce such a generic provider, or narrow it down to a specific `small` provider. have you considered it?

```
type ListCompactContextType = {small: boolean};
const ListCompactContext = React.createContext<ListCompactContextType>({small: false});
```

src/list.tsx
```diff
@@ -780,30 +831,23 @@ export const RowList = ({
...
-                    {index < lastIndex && (
-                        <Box paddingX={16}>
-                            <Divider />
-                        </Box>
-                    )}
-                </React.Fragment>
-            ))}
-        </div>
+        <ListContext.Provider value={{small}}>
+            <RowListDividerContext.Provider value>
```
brtbrt: this is weird right? `value` should be at least `true` or `false` I think 🤔.
does @Telefonica/mistica-design agree with this behavior? showing/hiding the dividier

MurilloLeoni: hi! setting it explicitly as `value={true}` actually breaks the `react/jsx-boolean-value` lint rule (screenshot attached), so the shorthand was the way to go here.

image

aweell:
@MurilloLeoni can you link to the spec of the list component with the updates?

src/list.tsx
```diff
@@ -34,6 +35,11 @@ import type {TouchableElement, TouchableProps} from './touchable';
 import type {DataAttributes, TrackingEvent, IconProps} from './utils/types';
 import type {ExclusifyUnion} from './utils/utility-types';
 
+const RowListDividerContext = React.createContext(false);
```
brtbrt: why have you introduced this provider? is the divider specification changing too? and actually, I see it always `true` (useless then?) + a weird branching 

```
            {hasDivider && !hideDivider && (
```
on line 289. I don't know: can you explain it better?

brtbrt (changes requested):

MurilloLeoni:
> I would have split into several PRs to ease focusing on the thing you're changing, and having a 1-1 relation with the subtasks
> 
> [Telefonica/mistica-design#2675](https://github.com/Telefonica/mistica-design/issues/2675)

hii! Just to clarify, would you prefer that we split this current PR into several smaller ones, or is this more of a guideline to follow for the remaining subtasks and future PRs?

src/list.tsx
MurilloLeoni: the `RowListDividerContext` defaults to `false`, and only `RowList` sets it to `true`.

I introduced it so `Content` can know when it is being used inside a `RowList`. This replaces the previous approach where `Divider` elements were added manually between rows in `RowList`.

Components such as `BoxedRowList` use `Content` without this provider, so `hasDivider` stays `false` and no divider is rendered there.

The `hasDivider && !hideDivider` condition is needed for the `dualAction` layout (for example, a row with both `onPress` and a `switch`).

In that case, `Content` is rendered inside the `BaseTouchable` on the left side only. If the divider were rendered there, it would cover only the touchable area, not the full row including the control on the right. For that reason, `hideDivider` prevents the internal divider, and a separate divider is rendered below the whole row instead.

MurilloLeoni:
> @MurilloLeoni can you link to the spec of the list component with the updates?

Sure! I’ll add the updated Lists specification link to the PR description and send here.
https://www.figma.com/design/Be8QB9onmHunKCCAkIBAVr/%F0%9F%94%B8-Lists-Specs?node-id=0-1

brtbrt:
> hii! Just to clarify, would you prefer that we split this current PR into several smaller ones, or is this more of a guideline to follow for the remaining subtasks and future PRs?

no worries, just for the future!

src/list.tsx
brtbrt: yes but why do you wanna it in the first place? isn't the `hideDivider` enough? why the context, which is always `true`?

src/list.tsx
brtbrt, line 1: why `boxedDefaultPadding` is imported but never used?

MurilloLeoni: i checked the current `src/list.tsx` and I can’t find any `boxedDefaultPadding` import there. I think the comment may be referring to an older diff or another file.

src/list.tsx
MurilloLeoni: Make sense. I’m thinking of refactoring `Content` so it no longer reads `RowListDividerContext` directly and instead only receives a `showDivider?: boolean` prop.

`RowContent` would become the only place that consumes the context and computes the final value before passing it down. The divider condition in `Content` would then become much simpler:

```tsx
{showDivider && (
    <div className={styles.rowDivider}>
        ...
    </div>
)}
```

This would keep the context usage contained in a single place and make the divider logic more explicit.

Wdyt?

brtbrt: https://gist.github.com/brtbrt/5dd556b9e78837cf297b1d27f1e2b69e suggested (not 100% sure if something is changing. have a try)
this way, we could get rid off the `showDivider` which is a quite imperative prop name, and we can also spare it! lemme know if it fits your needs

src/list.tsx
```diff
@@ -780,30 +830,23 @@ export const RowList = ({
...
+                    aria-labelledby={ariaLabelledBy}
+                    aria-label={ariaLabel}
+                    aria-live={ariaLive}
+                    aria-atomic={ariaAtomic}
+                    {...getPrefixedDataAttributes({testid: 'RowList', ...dataAttributes})}
+                >
+                    {children}
+                </div>
+            </RowListDividerContext.Provider>
+        </ListCompactContext.Provider>
```
brtbrt: maybe `Compact` is not a semantic name we use on mistica? WDYT @Marcosld

src/list.tsx
brtbrt: > Wdyt?

1. oh I see, the context was used implicitly as `false` . maybe the context is the easiest way to handle that information, you ar right, but I wouldn't use it implicitly (i.e. on accordions). if it was me, I'd explicitly put it because it's more clear when you read it
2. I'd avoid imperative names like `showDivider` and I'd chose a more functional-alike naming

src/list.tsx
Marcosld: If you ask me for my opinion I would create ONE generic `ListContext` and add all the props child rows need. Simple & extensible

src/list.tsx
MurilloLeoni: Thanks for the suggestion! I tried it, and unfortunately it breaks the main goal of this task.

The reason the divider currently starts at the text (instead of the asset) is that it lives inside the same CSS Grid as the asset and the text content:
```css
.content {
  display: grid;
  grid-template-columns: auto 1fr;
}

.rowDivider {
  grid-column: 2 / -1;
}
```
`grid-column: 2 / -1` only works because the divider is a grid item of that same grid container.

In your version, the divider is rendered as a sibling of `RowContent`'s output (outside `Content`), so it's no longer part of that grid. As a result, it loses the column alignment and goes back to spanning the full row width, starting from the asset, which reverts the divider alignment update introduced in this task.

There's also a second issue: rows that only have a `switch`, `checkbox`, `radioValue`, or `iconButton` (without `href`, `to`, or `onPress`) go through `renderRowWithSingleControl`, and in your version that path never renders the divider. because of that, those rows lose the divider completely.

I'm attaching a screenshot showing both issues: the divider is back to full width, and it's missing for non-`dualAction` rows.

I agree that `showDivider` isn't the nicest name, but I think it's safer to keep the divider inside `Content`'s grid and simply rename the prop, rather than changing where the divider is rendered.

Would that work for you?

image
image

aweell (approved):

src/list.tsx
```diff
@@ -466,6 +479,29 @@ const RowContent = React.forwardRef<TouchableElement, RowContentProps>((props, r
     const [slotText, setSlotText] = React.useState<string>('');
     const [rightText, setRightText] = React.useState<string>('');
     const assetText = getAssetText(asset);
+    const {hasDivider} = React.useContext(ListContext);
+    const [dividerOffset, setDividerOffset] = React.useState(0);
+    const measureDividerOffset = React.useCallback(
```
Marcosld: I am not a fan of this calculations because they won't work in SSR, but I can't think of a better solution with current spec :(

Marcosld (approved):

src/list.tsx
Marcosld: Can we remove `hasDivider` prop from the context and pass it down as a prop to `RowContent` component? `RowComponent` is only rendered in the `Row` / `BoxedRow` components where we already know if `hasDivider` should be true or false. Would that be ok @MurilloLeoni ?

MurilloLeoni: Hi @Marcosld , that makes sense to me, since RowContent is only used inside Row and BoxedRow, hasDivider can be passed explicitly at those call sites instead of going through context. I've already pushed the changes.

Marcosld: Nice! I'll merge this ASAP

Marcosld: For some reason the diff is very weird... :o
image

PR #1690, opened 2026-08-11
fix(SearchField): prevent layout shift when clear button appears

### LINKED ISSUE #1635: Layout shift when using search field in right navigation bar
When you start typing the field expands to show the `x` on the right:
https://mistica-web.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8BZAhgSwHYDk0DcMBzNAFwwiwCE0AnAHSwAInbiALUgXmEZaQCSWADbYYTAM4AHNGBg8AjADYAvkzSiiWAaRgBbCV3rgYWXQxBMYADxlYoPAAwqAfHxYfP-AMow6YdgAxDBhhKCYAMwBXYWEAdQwoUnYmLDQ9eWMJP1oA4yZhNAAjUKMQX392fIB6Nyx3Ly8kPEISckoaWgBBMHbmSgAFWhgJQ2AACgBKJi4XJmAVNToMNABaQpLhMoBRM1o6YzrG-i78MgPmY5YJDAAveWAMCQAREYBrUggpAHlaKmIiDBaEwAPxMADMACYmPAmJCACwqBrHbAYcgaQzGFAAGWMyMaElyZU4pCkEng1WqUgwYAkUT0ADopOwIJ8JNUFJDHI4OVzHHjLo1avirh5Hi93p8fn8AUCmAAyeVMADku1I%2B1oyqRgtFSGqLWIZAo1DoPT6R2OeqEoiwMBc2uFWBAABoQMl9CMEABtEAoCCECSkOgAfRwMAA7iAALqu8OJZISb0KAAcUKjKiAA

AlexandraGallipoliRodrigues:
Resolves #1635

## Changes

- Reserve space for the SearchField clear button before it is displayed.
- Keep the input and label width stable while the clear button appears or disappears.
- Add an acceptance test covering the field width before and after typing.

## Why

The clear button is displayed only after entering text. Previously, its appearance
increased the SearchField width and shifted adjacent navigation items. The field
now reserves that space from its initial render.

## Demo

https://github.com/user-attachments/assets/d996171c-0a8b-4caf-8d83-233d7bc2a340

## Before

https://github.com/user-attachments/assets/ec6587a5-bf36-47eb-8828-6098c6e06e7b

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/text-field-base.css.ts
```diff
@@ -331,6 +331,12 @@ export const endIconContainer = style([
     },
 ]);
 
+export const endIconPlaceholder = style({
+    flexShrink: 0,
+    width: `calc(${iconContainerSize.small} - (${iconContainerSize.small} - ${iconSize.small}) / 2)`,
```
AlexandraGallipoliRodrigues: FieldEndIcon is small and bleeds 6px to the right, so reserve its effective layout width

Marcosld: Can we export `bleedArea` from icon-button.css and use it here so it is more clear?

src/text-field-base.tsx
```diff
@@ -160,6 +160,7 @@ interface TextFieldBaseProps {
     startIcon?: React.ReactNode;
     endIcon?: React.ReactNode;
     endIconOverlay?: React.ReactNode;
+    reserveEndIconSpace?: boolean;
```
Marcosld: Perhaps add a comment here explaining what this should be used for

Marcosld (approved):
GJ!

src/text-field-base.tsx
brtbrt: pitimimi: 

1. actually I'd change the name of this, because it looks imperative programming. try ask claude if we could use a better naming. maybe `endIconSpaceReserved`?
2. +1 on marcos' comment. actually, we can add jsdoc on the whole block, that does help when using the components! :)

src/text-field-base.tsx
```diff
@@ -307,6 +309,8 @@ export const TextFieldBase = React.forwardRef<any, TextFieldBaseProps>(
 
         const startIconWidth = `calc(${iconSize.small} + ${styles.fieldElementsGap}px)`;
         const endIconWidth = `calc(${styles.iconButtonSize} + ${styles.fieldEndIconGap}px)`;
+        const hasEndIconSlot = !!endIcon || reserveEndIconSpace;
```
brtbrt: I'd avoid using these namings: 

`hasEndIconSlot` vs `hasEndIconSpace`: this is the "risky pair" IMHO. They sit on adjacent lines, differ by one word, and the semantic difference (slot excludes overlay, space includes it) is not recoverable from the names alone. A reader has to trace both definitions to know which is which. That is the kind of near-homonym pairing that causes the wrong one to get used later.
- `slot` is established vocabulary in this repo for a prop that accepts external `React.ReactNode` content to render in a designated area: slot and footerSlot in card-media.tsx/card-cover.tsx, and topSlot/burgerMenuSlot in navigation-bar.tsx for instance! Every existing `*Slot` is a content-injection prop, not a layout-space flag.


suggestion:
- `hasEndIcon`
- `reservesRightSpace`


These avoid the slot convention collision and no longer read as synonyms of each other :)

src/text-field-base.tsx
```diff
@@ -160,7 +160,8 @@ interface TextFieldBaseProps {
     startIcon?: React.ReactNode;
     endIcon?: React.ReactNode;
     endIconOverlay?: React.ReactNode;
-    reserveEndIconSpace?: boolean;
+    /** Reserves space for an end icon that is rendered conditionally. */
```
brtbrt: I guess the name is now self explanatory without any further comment :)

brtbrt (approved):

PR #1695, opened 2026-08-12
feat(TextField): adds new input padding values and prop small to input field variants 

AngeloOBonfante:
#### This PR adjusts the Input field variants.

- Implements the new left/right padding values that will be added in https://github.com/Telefonica/mistica-design/pull/2725
- Adds new prop small on input and its field variants.
- Story files have been updated to use prop small.

Token modification pull request is still open so this might require the skins to be re-generated once the token is merged.

Ref: https://github.com/Telefonica/mistica-design/issues/2672


yceballost (approved):

yceballost:
please include test for new props

yceballost (approved):

PR #1719, opened 2026-09-02
feat(callout): updates container tokens for callout

AngeloOBonfante:
This PR implements the two new tokens that will be added in mistica-design on the callout component:

- Callout now uses backgroundContainerAlternativeOverAlternative and backgroundContainerAlternativeOverBrand.
- Screenshot tests updated.
- Skin tokens have been re-generated from the branch used in https://github.com/Telefonica/mistica-design/pull/2824

Callout changes will only be visible in vivo-evolution.

related mistica-design issue: https://github.com/Telefonica/mistica-design/issues/2692
related mistica-design pr: https://github.com/Telefonica/mistica-design/pull/2824


src/callout.css.ts
```diff
@@ -38,11 +38,13 @@ export const closeButtonContainerSize = style({
...
     'default-over-negative': [sprinkles({background: skin.vars.colors.backgroundContainer})],
-    'default-over-alternative': [sprinkles({background: skin.vars.colors.backgroundContainer})],
+    'default-over-alternative': [
+        sprinkles({background: skin.vars.colors.backgroundContainerAlternativeOverAlternative}),
+    ],
     'default-over-default': [sprinkles({background: skin.vars.colors.backgroundContainerAlternative})],
     'default-over-media': [sprinkles({background: skin.vars.colors.backgroundContainer})],
 
-    'brand-over-brand': [sprinkles({background: skin.vars.colors.backgroundContainerBrandOverInverse})],
+    'brand-over-brand': [sprinkles({background: skin.vars.colors.backgroundContainerAlternativeOverBrand})],
```
yceballost, line 47: why we use a token call **alternative**OverBrand in a case that define "brand-over-brand"? looks wrong🤔

I think you need to create a relation of contexts between all of the alternative variants:

alternative-over-default
alternative-over-alternative
alternative-over-brand
alternative-over-negative
alternative-over-media

@aweell or @Marcosld can you confirm this?

yceballost (changes requested):
Callout specs with this definition is missing. We dont use random files with definition to make changes in the library. The process requires update official specs documentation to avoid to have definition out of date

The callout specs file
https://www.figma.com/design/LzhzNPzE8nY9d7Yvo10d5d/%F0%9F%94%B8-Callout-Specs?node-id=0-1

src/callout.css.ts
Marcosld: I'm not sure I understand your question

AngeloOBonfante: Hello, just for context, assigning AlternativeOverBrand to brand-over-brand is intentional in order to differentiate the Callout in vivo-evolution while not altering the look of other brands. 
Also since the default color is already Alternative, adding an Alternative context might not work.

I believe @aweell can confirm this.

brtbrt: related:
https://github.com/Telefonica/mistica-design/pull/2824#issuecomment-5541496753

yceballost: talked with @aweell, we are ok with this solution

yceballost (approved):

Marcosld (approved):

brtbrt (approved):

AlexandraGallipoliRodrigues (approved):

Marcosld:
@yceballost  let us know when design pr is ready so we merge both

ieduardogf:
@yceballost  Can we merge this PR? or should we wait for the related mistica-design PR?

PR #1720, opened 2026-09-02
feat(Header): add new headerPretitle token and add an option to change header title preset

BarbosaNath:
Changes
- Update `renderPretitle` auxiliary function to use `headerPretitle` token
- Add `titleTextPreset` property to change header `title`
- Add `renderTitle` auxiliary function to create header `title` with `titleTextPreset`

Related
- Figma Spec: [Header Specs](https://www.figma.com/design/Xm3r8nRNlbAiZNlkaH5YX1/%F0%9F%94%B8-Header-Specs?node-id=1649-9779&t=xpkZLhx669MB0qNX-4)
- Issue: https://github.com/Telefonica/mistica-design/issues/2816
- PR in mistica-design: https://github.com/Telefonica/mistica-design/pull/2825


src/header.tsx
```diff
@@ -47,45 +51,77 @@ export const Header = ({
...
         if (typeof pretitle === 'string') {
-            return <Text3 {...baseTextProps}>{pretitle}</Text3>;
+            return (
+                <Text
+                    {...baseTextProps}
+                    desktopSize={textPresets.headerPretitle.size.desktop}
+                    mobileSize={textPresets.headerPretitle.size.mobile}
+                    desktopLineHeight={textPresets.headerPretitle.lineHeight.desktop}
+                    mobileLineHeight={textPresets.headerPretitle.lineHeight.mobile}
+                    weight={textPresets.headerPretitle.weight}
```
Marcosld: Perhaps we can move these into `basePretitleProps` and reuse, just like in the `renderTitle` function below

src/__screenshot_tests__/header-screenshot-test.tsx
```diff
@@ -131,3 +131,29 @@ test.each(DEVICES)('Header with bleed', async (device) => {
...
+test.each`
+    device          | titleTextPreset
+    ${'MOBILE_IOS'} | ${'title1'}
+    ${'MOBILE_IOS'} | ${'title2'}
+    ${'MOBILE_IOS'} | ${'title3'}
+    ${'MOBILE_IOS'} | ${'title4'}
+    ${'DESKTOP'}    | ${'title1'}
+    ${'DESKTOP'}    | ${'title2'}
+    ${'DESKTOP'}    | ${'title3'}
+    ${'DESKTOP'}    | ${'title4'}
```
Marcosld: IMO testing with 1 or 2 preset overrides is enough, because screenshot tests are quite heavyweight.

Marcosld (approved):

src/__screenshot_tests__/header-screenshot-test.tsx
BarbosaNath: Title 3 and 4 removed from tests to preserve storage!

src/header.tsx
BarbosaNath: Thanks for the tip, it's a better solution than mine! Implemented!

brtbrt (approved):

AlexandraGallipoliRodrigues (approved):

src/header.tsx
```diff
@@ -30,12 +31,15 @@ type OverridableTextProps = {
...
 
+type TitlePresetsType = 'title1' | 'title2' | 'title3' | 'title4';
+
 type HeaderProps = {
     headline?: React.ReactNode;
     pretitle?: RichText;
     pretitleAs?: HeadingType;
     title?: string;
     titleAs?: HeadingType;
+    titleTextPreset?: TitlePresetsType;
```
yceballost: I think another name for this prop would work better, like `titleSize` or `titleType`

We already have the Title components:
```
<Title1 />
<Title2 />
<Title3 />
<Title4 />
```

So, I prefer a reference for Title as a component instead of tokens. This sizes should match perfectly with Title components types.

what do you think?

yceballost (changes requested):

src/header.tsx
BarbosaNath: About this second change, do you mean doing something like this?

```ts
type TitlePresetsType = React.ComponentType<TitleProps>
type HeaderProps = {
    //...
    titleType?: TitlePresetsType;
    //...
};

//...
    const renderTitle = () => {
        if (!title) return undefined;

        const baseTitleProps = {
            children: title,
            as: titleAs,
            dataAttributes: {testid: 'title'},
        } as const;

        const TitleComponent = titleType ? titleType : small ? Title3 : Title4;

        return <TitleComponent {...baseTitleProps} />;
    };
//...
```
Since the type using strings limits to title1/2/3/4, I think it's almost the same in the end. But in the suggested changes it would have an additional import when using the `Header` component with custom title sizes.

yceballost: My comment is about how to name the prop! (I’m a designer, so I’m not talking about the code solution, my fellow engineers are better at that :D)

I suppose both this one and the current solution would work the same way because they both draw from the same token source

`<Header titleTextPreset="title1"  title="The last invoice is available" />`
vs
`<Header titleSize="title1"  title="The last invoice is available" />`
`<Header titleComponent="title1"  title="The last invoice is available" />`
`<Header titleType="title1"  title="The last invoice is available" />`

I dont have strong opinion about what is the best naming

BarbosaNath: Ooh, now I got it! I was confused about the second part of the message 😅. Just to triple-check, my action now is to rename `titleTextPreset` to something like `titleSize` without changing the implementation, right? I will get into it right away!

yceballost: exactly

BarbosaNath: Alright! I've just renamed it!

yceballost (approved):

brtbrt:
marked as blocked until mistica-design has been approved by at least one designer

PR #1721, opened 2026-09-08
feat(BoxedRowList): add border option for selected state

### LINKED ISSUE #1691: Add selectable card type and selected border for boxed rows
Add a selection mode to cards to support selectable cards using a checkbox or radio button, behaving as a single selectable unit within a group.

The goal is to define a clear and consistent selection pattern for cards and boxed rows. To ensure feature parity, add an outline/border option for the selected state of boxed rows, reinforcing the selection visually around the entire row.

[Cards specs](https://www.figma.com/design/tKdPOfcUALzVIh5oizFbm7/%F0%9F%94%B8-Cards-Specs?node-id=11030-16174&t=4Pb6KbuLjHZCqWGG-4)

[List Specs](https://www.figma.com/design/Be8QB9onmHunKCCAkIBAVr/%F0%9F%94%B8-Lists-Specs?node-id=6975-8017&t=gvqfhEjGfA7cgidl-4)

Design tickets: [selectable cards](https://github.com/Telefonica/mistica-design/issues/2622) [Add border option for selected state in Boxed Rows](https://github.com/Telefonica/mistica-design/issues/2639)

AlexandraGallipoliRodrigues:
[issue](https://github.com/Telefonica/mistica-design/issues/2639)
[playroom](https://mistica-hb8w63eb7-flows-projects-65bb050e.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8AlGBnADhAduglgG4wAyAhgJ4QCuALgHwA6OABC0gEIQAeLmZUKPhwBzAJoBeYAEYAbAF8mrNuwDKtMmADWLLJphS5i5ipWoB%2BCAHEATjUwscZALYHGITABsyOdy1gAZmTUnrQAamSe1G4g6Bo4UGQ2UO5KpqacPDBQKBAA7iT4cWnpGVzc2bl5JqXp-ILCIoYKNbUq6DCeMGAaAEZdrW0stPi0XRLuHACHBJDug202FhARUTG9ZLPzym0A9CVtmRU5%2Bbqd3X1dw6Pj7gCicVMJSX5LQiuR0ROx8YnJfvsFukjpVTh0uj0yP0YNcxjEAAo2GDOfDUZyvZarL7uTBIlFogEHYG7cqggpFBhApC7FDLWz2JRU0kncnFIFsEEs9kqepCUTNeTctjgi5QgY7WojOHfAByEBGAXwYE0lhwGD4EBsLGRZHwnm2QxYYAAFt0tL0eFJgI4XDB4Cx3Dq9QBaHDy-CK5UjXDodwAGn8MCCIXCnztwxs0RY8kFEtMgLjZmZVSFZwhlxgqaltxAcoVSpVuHVmGo6GNBqGJrNFu4VptrntONLxtd7s9ZG9eD8Me5CcOyfyqZFkOhWZuMQA8gEYDYNOg%2BDP0LhIvgAF4CTYVtpV7Q1utOBsOjyL5eeNfZZ0QALTmy%2BkDR2N7IkcklZFmFNm%2BROclOJ4XnEdxTabMYgAQR6IgkhYZxoAgFgIHQMBqDsLd0lgRCbHwTBO2%2BABVdAoO8FgyH4TCYBwMB8DIeDEOQ6jyJYIiSLPZVKIAZ98EBuXQPJRhNfdbUbEA-i0Z0YNgbtH3SBMmTfX9gPHb4rGoJI-iNCAYJYIJCE1UYIFQ0x0LATDsNVb5VAAyjlz4JJqJEFTkheLi-xYTCRGNWgpAACmtYdaGyANcAs9NsgAYWNHwRBheQAEoWAkegWC81MkAASUgHAODoWhcFTNh0twKQ-OyFgAH4WAKnAAAkYCSWgADE9S6KAWHtSqarqtB7O8GwpMNJIqOdbxek6KQ8vaAD-Ja8r3AARWoUYoNgLSyB0zCcrvVqj1AgBHiwtWo7TdI2gzaj6oZaAoTB1iWBJTtKDZtBEOxqASAAVK6DGAYrpqPJczxSe8hNoW69CRHBaHcc62lwREMHQby4oS%2BCcGCi4woi0QYC8gBCH6Yuh6TnxUAmgT7MxdnUTQtBKalyiUak0CwH0iFISgaEpHAQHkIA)


src/list.css.ts
```diff
@@ -11,8 +13,41 @@ export const row = style({
...
+export const selectionOutline = style({
+    position: 'relative',
+    borderRadius: vars.borderRadii.container,
+    selectors: {
+        '&::after': {
+            content: '',
+            position: 'absolute',
+            inset: -(selectionOutlineWidth + selectionOutlineOffset),
+            border: `${selectionOutlineWidth}px solid transparent`,
+            borderRadius: vars.borderRadii.container,
```
yceballost: you need to include a calc to avoid weird spacing in the corners

```suggestion
            borderRadius: `calc(${vars.borderRadii.container} + 3px)`,
```

yceballost (changes requested):

src/__stories__/list-story.tsx
```diff
@@ -509,6 +513,67 @@ BoxedRowListStory.argTypes = {
...
+                        Icon={selected ? IconHeartFilled : IconHeartRegular}
+                        aria-label={selected ? 'Unselect row' : 'Select row'}
+                        backgroundType={selected ? 'solid' : 'transparent'}
+                        type="brand"
+                        onPress={() => onSelectedChange(!selected)}
+                    />
+                )}
+            />
+        </BoxedRowList>
+    );
```
Marcosld, line 559: Do we really need these kinds of behaviors? Looking at the spec behavior is only defined for regular controls, not custom ones @AnaMontes11

src/list.tsx
```diff
@@ -38,7 +38,15 @@ import type {ExclusifyUnion} from './utils/utility-types';
 type ListContextType = {small: boolean};
 export const ListContext = React.createContext<ListContextType>({small: false});
 
-type Right = (({centerY}: {centerY: boolean}) => React.ReactNode) | React.ReactNode;
+const BoxedRowSelectionContext = React.createContext<((selected: boolean) => void) | undefined>(undefined);
+
+type RightProps = {
+    centerY: boolean;
+    selected: boolean;
+    onSelectedChange: (selected: boolean) => void;
```
Marcosld: This breaks current contract... on the other hand... do we really need this behavior? For me it would be ok to only apply this logic to regular radios, switches and checkboxes, not custom ones @yceballost ?

src/list.tsx
```diff
@@ -522,6 +567,7 @@ const RowContent = React.forwardRef<TouchableElement, RowContentProps & {hasDivi
         const ariaLabel = ariaLabelProp ?? (isInteractive ? computedAriaLabel : undefined);
 
         const radioContext = useRadioContext();
+        const onSelectedChange = React.useContext(BoxedRowSelectionContext);
```
Marcosld: I think we can remove this context and just pass the prop down

src/list.tsx
```diff
@@ -877,17 +953,28 @@ type BoxedRowProps = ExclusifyUnion<
...
+    const outsideVariant = useThemeVariant();
+    const [right, selected, setSelected] = useSelectableRight(props.right, hasControlProps(props));
+
+    return (
+        <InternalBoxed
+            overflow="visible"
+            className={classNames(
+                styles.boxed,
+                styles.selectionOutline,
+                styles.selectionOutlineColor[outsideVariant]
```
Marcosld: we can add these styles conditionally here instead of relying on dataAttribute selected prop, right?

src/__stories__/list-story.tsx
AlexandraGallipoliRodrigues: we talked about it in the meeting, but yeah, it is expected to have a custom control

src/list.tsx
AlexandraGallipoliRodrigues: Agreed. Since BoxedRow renders RowContent directly and this only needs to cross one level, the context is unnecessary. I’ll pass onSelectedChange as an internal prop instead

src/list.tsx
AlexandraGallipoliRodrigues: Yes, we can conditionally apply the outline color class based on selected. I’ll keep the base outline class always mounted so the selection and deselection transition still works

src/list.tsx
Marcosld: I see yo agreed to apply the behavior to custom controls too, but perhaps applying it to the `right` prop is way too much? It makes the API kinda complex

src/list.tsx
```diff
@@ -392,6 +398,43 @@ const useControlState = ({
...
+const SelectableToggleIconButton = ({
+    checked,
+    defaultChecked,
+    onChange,
+    onSelectedChange,
+    ...props
+}: ToggleIconButtonProps & {
+    onSelectedChange?: (selected: boolean) => void;
+}) => {
+    const initialSelected = React.useRef(!!defaultChecked).current;
```
Marcosld: This ref doesn't look like it is needed

src/list.tsx
```diff
@@ -392,6 +398,43 @@ const useControlState = ({
...
+        const result = onChange?.(selected);
+
+        if (checked !== undefined) {
+            return result;
+        }
+        if (result) {
+            return result.then(() => onSelectedChange?.(selected));
+        }
+        onSelectedChange?.(selected);
+    };
```
Marcosld: I think onSelectedChange should be called always unless onChange is a promise and it fails, right?

AlexandraGallipoliRodrigues: for uncontrolled usage, onSelectedChange is already called unless the async onChange fails. When checked is provided, the component is controlled, so the selection outline must follow checked; calling it immediately could leave the outline selected while the icon remains unchecked

src/list.tsx
AlexandraGallipoliRodrigues: you’re right, the ref isn’t necessary here. I’ve removed it and now use defaultChecked directly when synchronizing the selection state

src/list.tsx
AlexandraGallipoliRodrigues: oki, i refactored it, what do you think?

Marcosld: It looks better to me... yeah. Could we have a case where we have a ToggleIconButton but we don't want the card to be highlighted?? That seems like a reasonable use case to me... @AnaMontes11 Perhaps the icon is a padlock locked/unlocked ?

In that case I'd also remove that logic. The user can rely in the new `selected` prop and highlight the card if it suits the use case.

AnaMontes11: Yes, I agree with removing that logic. I think we are mixing two independent semantics under the same visual effect: selected in a selectable rowList normally means "this item is part of the selection for an action across multiple rows". The state of a ToggleIconButton, like a locked/unlocked padlock, a favorite, or play/pause, is the state of an isolated control inside the row, with no necessary relation to that selection. Exposing selected as a user controlled prop seems like the right approach.

AlexandraGallipoliRodrigues: makes sense, I’ll remove the automatic row selection behavior from ToggleIconButton and render it normally. Checkbox, switch and radio will keep their current automatic selection behavior, while toggle or custom controls can explicitly highlight the row using the new selected prop.

src/icon-button.tsx
```diff
@@ -209,6 +209,7 @@ export const InternalToggleIconButton = React.forwardRef<
             dataAttributes={{
                 testid: 'ToggleIconButton',
                 ...dataAttributes,
+                checked: checked ?? checkedState,
```
Marcosld: is this needed?

Marcosld (approved):

src/icon-button.tsx
AlexandraGallipoliRodrigues: No, this was left over from the previous :has([data-checked]) implementation. It’s no longer consumed now that ToggleIconButton doesn’t drive the row selection, so I’ll remove it, thanks!

src/__stories__/list-story.tsx
```diff
@@ -103,6 +106,7 @@ const Template: StoryComponent<
     const getControlProps = (index: number) => {
         let controlProps: any = {};
         const onPress = () => alert('Pressed');
+        const defaultControlValue = boxed ? index % 2 === 0 : true;
         switch (control) {
```
brtbrt, line 111:
```suggestion
        
        switch (control) {
```

pitimimí

src/list.tsx
```diff
@@ -446,361 +446,373 @@ const getNodeText = (node: HTMLElement | null): string => {
...
+
+        'aria-label': ariaLabel,
+        'aria-labelledby': props['aria-labelledby'],
+        'aria-description': props['aria-description'],
+        'aria-describedby': props['aria-describedby'],
+        'aria-current': props['aria-current'],
+    } as TouchableProps;
+
+    const [isChecked, toggle] = useControlState(props.switch || props.checkbox || {});
+    const selectedControl =
```
brtbrt:
```suggestion
    const isControlSelected =
```

src/__screenshot_tests__/list-screenshot-test.tsx
```diff
@@ -168,6 +168,21 @@ test.each(controlsWithOnPress)('Click control with onPress - %s', async (control
...
 });
 
+test('Selectable custom boxed row', async () => {
+    await openStoryPage({
+        id: 'components-lists--selectable-boxed-rows-story',
+        device: 'MOBILE_IOS',
+    });
+
+    const buttons = await screen.findAllByRole('button', {name: 'Select row'});
+    await buttons[0].click();
```
brtbrt, line 182: maybe two screenshot to make sure both are selected?

brtbrt:
designers: is this wanted?
image

just to make sure we're aligned

brtbrt (approved):
good job :)

I've struggled a bit because there's a lot of code churn, but I think it's an AI thing? maybe we can talk among us developers about this! to reduce cognitive charge when it comes to a PR, and focus on it properly

src/__screenshot_tests__/list-screenshot-test.tsx
AlexandraGallipoliRodrigues: what do you mean? both rows are selected but only one has the selected prop

src/list.tsx
AnaMontes11: Nice! thanks @AlexandraGallipoliRodrigues

AnaMontes11 (approved):
Nice job!

src/__screenshot_tests__/list-screenshot-test.tsx
brtbrt: clarified on chat

yceballost (approved):

PR #1722, opened 2026-09-10
feat(Skin): updates vivo-evo skin to use new responsiveLayoutMargin values

AngeloOBonfante:
This PR updates the vivo-evolution skin responsiveLayoutMargin to use 16px, updated with the branch used for https://github.com/Telefonica/mistica-design/pull/2828.

### DIFF
--- a/src/skins/vivo-evolution.tsx
+++ b/src/skins/vivo-evolution.tsx
@@ -840,7 +840,7 @@ export const getVivoEvolutionSkin: GetKnownSkin = () => {
                 right: {mobile: 16, desktop: 40},
             },
             boxedDefaultPadding: {left: {mobile: 16, desktop: 24}, right: {mobile: 16, desktop: 24}},
-            responsiveLayoutMargin: {mobile: 8, desktop: 48},
+            responsiveLayoutMargin: {mobile: 16, desktop: 48},
         },
     };
     return skin;


src/skins/vivo-evolution.tsx
```diff
@@ -840,7 +840,7 @@ export const getVivoEvolutionSkin: GetKnownSkin = () => {
                 right: {mobile: 16, desktop: 40},
             },
             boxedDefaultPadding: {left: {mobile: 16, desktop: 24}, right: {mobile: 16, desktop: 24}},
-            responsiveLayoutMargin: {mobile: 8, desktop: 48},
+            responsiveLayoutMargin: {mobile: 16, desktop: 48},
```
brtbrt, line 843: have you edited it by hand?
It looks it was not made by the action we should use, but in this very case content is identical to what the action produces, so it's not a big deal. 

but just to be clear: the recommended way is _using the actions_ 

like the following:
https://github.com/Telefonica/mistica-web/actions/runs/34592199248

Image

in this case you must use the `refs/pull/` ref thing because you're working on a fork,and ask for a run accept on the channel we have.

Hope it helps! 😀

brtbrt (approved):

AlexandraGallipoliRodrigues (approved):

yceballost (approved):

PR #1729, opened 2026-09-18
feat(Card): add selectable state

AlexandraGallipoliRodrigues:
Adds a selection outline to DataCard, MediaCard, CoverCard, NakedCard and AdvancedDataCard. The Checkbox, Switch and RadioButton controls included in their slots automatically communicate their state via an internal context. The `selected` property allows you to override this and connect custom controls. The colour of the outline depends on the outer theme and the card variant.

When there is a single control, the body of the card acts as the selection area, with visible focus and keyboard support. Selection takes precedence over navigation or the main `onPress` event, whilst footer actions remain independent. If there are multiple controls, each retains its own individual interaction.

[playroom](https://mistica-d2t80323u-flows-projects-65bb050e.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8AhCAPABABwIZSgEsA7AcwF5gAmAFgF8A%2BAHWM0yQGUAXXMAa0wBnPGBiUAzFUYs2bYAG0ZsuQGFcAJyjxMAEVw81mgDSZiuALYxtTEHoMaoNukaXLgr5ZkNbMAWRhEuN4urJ5sZpbWIP6B3jYhYWyE5rikMBzqYFEAFlxc2ILwAPRF2IRgggCu5gB02NkQXBCCRYRQRQCMAAwdABxFAJxdXUUAbMPxHmzOU5juoWHe2ioQAG4w6sGz4RZWmDYr65sOkwueyanpmTl5BcWl5VW19Y3Nre3dfYPDYxMgCZ4Zmc5NsvA5tAA5XB8AJbYGyCJ7GxQmFQOL-UEXNIZLL7EC5fKFEplCrVOoNJotNqdHr9IYjcZdU5hIFhYBLTAAQSgq1wxFEUDsQQcJkRUW5vP5ASF6NZsgAujUUtgABQq9ki0y7ExYq5gOgASkw5AYc1mkGIgi4mHUMGIsGOmmNmBVoK4hC4ABsYACwhauOoIJ7fZ5BDBvWAuAEQ8psJ7eDBLMRreQ8U1sDZZkaTWb4WwLVahJ7Gs7XXnZEgiKtQcorQBPb2UGueOMJpMp8ipmzpzPlxKYAD8uf7I9k2GaHsIEGIUVwACNBEHKlHmaOR%2BntB1RjG18p1IRSLlN9vm6OAF4ASXtMHQ2ioO93dFPnm0lWvADMSAFn0%2B%2B5hmH%2Bcj%2BoGnq-qOSBFFWAH9gaADcLCggW1q8O606ls2qDLk0xAAAr7ik6h1s%2BggpJ6nrPtOeEwIIgiUCq2bQbumC4N66hcOqz7KO6XpIiAnJgGAhAAM%2BsCQsDYHaRB2iuGKAZ4lg0ZcUQADK4EI4YwAJwmsMQECYLAc4wJgYAWHOhAaDUq6PgaP7Noxo78YJIkYUUKBYZR%2BEaHW9meHBCHlraXCVOorBluBcRyTCdaUNx3pgWu2QwPgnpfpQSAACqpJgXB1hJ5A2NggbmBANgMJlpCYPGhmehB5UMPFo6xWIwBNQ1I6wIIYD7tgqHEPlIDpdkhCCJgw3Mfp1FdYQPVTsQva7mGEZRlAlCLZpy1tf2wA1DtKq6jig5zPtmR0Jg2jAIam2JNtu2IsanZ4hKfICjKJwgJxh3AIIxZcNo8jfY08omHO7kzsxkazVd-bnQDXDA6DeEXIR2goZDBpQ7IRQ%2BcofnwnQ8FzfCgXBaFoKcDw-CYFFlCIqdwgJpQW7SH%2BSDZFQDDALTEFswBGFVsROWNsA8xMfpw2tnWUSkPujiyaL0ttOliatlGKyetUlpRLaEn6CquDLhAAC0n5w5g5gkCk6AqlQvRdNg6AmB077qAaNly0xpC4Ngd40A%2BiQsQexAXlG5iFHiVoaFwVkjnQGNsNj122tejpQGFTEHIl-Bzhg0fhZnfDZ%2BgH07JYlAAAYACSc7sdCG2A%2BeF2Xcf%2B-uuCG1V4b9RwGladOGjGW9xc2pJGx0cAwFBgAot67aGsapoT560%2BJtJzfKFjz7o-5u7AEnDreGnu42BwADuHr17nI6cGfXAX3JyiIuXVe04bgg3-XTdDxo5nt-OnfHz3QSfd1AD00PNUWw9k5j0XsvWeDFjLTgDFPGeq8PobzklvQmO894bAPh9A4lQrQQHMJAh0l9%2BxIBUA3HO98ES7CftXSwtcwCEKaOYT%2BtC2Dfzbh3T0XdAGzX7iZMB71OGkNHsAdU9dNKojnjmQ%2BEDKyEGrGI2sgtmoiwgWEM%2BUAuDZG0JIP2otEoHiPJgQxQ8whEGEPGSWeJ3zeiLu7LRygA6kCDiHMONhRDJg2OQiBAArVhhB3x1hWL45MUQfFRnUP40WhcOCEDPCQUgURs6aA2IbQucSmLpIdFEKg9shBBjaJgAAxMMXooxxC9BybuOcvA%2BDSwgG%2BHw0j%2BABEOjYCpXQqk1JsGdNM6g%2BT0yTlHZxLj8xBggOoKIJ8hoySMUxd8iDElnj2FQE8qjZCx0sf%2BPZbAkD01YNww2Q0CB2n6gGSoMBSrj3zp0ocNhADI5AMqITgILHITruCCUEh6YNFljJZyhWkwE-MQaM%2BCQDZygERURu4AVrl3iPFOCjRw2AAGK4FWDMj0JUJngXShAUgpBvQXgtG5PI04h7tNRJQNIXBuD6BgCqSujCYC13fNi3FUYy4mC5Z6MMiLRbTioXyNIq0YCMp4FGVlz8a7G25fuXlwqmK0oCHhCABRKCaK0eS6c2h9XEAABJJTYhiwgZFIXbJsKc3hNgogAEVKgen7lynFyr8XAsBGvTwb51VQE1dq4WeyjWGotKayOAAlGApBKjxnUN6sItrW6-2qg6vEvgNDCMwO6nlXqh67NoUCj6DKmWyrZS-PNyqYB8tzSxIVm9fzPmRcnPBtCDh8igPgAttCMrEtJTAI1lLsI0oeYGwMwbdUQLDZgI1ykID8BUMWMMUAY1xoTUmzwKaf72oQHiHQ1E5zFgAI43I0HUn1Q9-XjqDbREN2zZ3zsXXwAA8hJYg6740aC3coHdPC-7kX3TYFAp7z2xIJWuItgLvm%2BWbXJJAUb8BTgAOKBkqBmWhj9gCVoVcMogEAOGiztYB8u76gEQpGrATA8qmFEbXLBsI5NGlCBEM1WgzMtEKBsJyeIeIUA2EVMqNUWreoMT2Tg1FBzMBlyQwR7QWqnKsCrqJyGfLpOIeQxAEd1LtnKF5OrZqqnpy%2BtHCR6q5du4Rgo-3RT2kaPAGM8QOg9HJlsEk9AxBIFYGr3gTAlByZTOJHQZMmyYjVXgSKEy-gjGKxFDk6h9D2BeZyXZfdLs0QAjmXRJgAAZLlj6kn21aIOF5oMmBF3xqaJe5QlDqFF22Y-GwlhAhnLaLAOa8KXGwC5VVqhMjvzbPM--EAVnNI2ZAcI2WEmUV0XgUcvArAsZBbYCWsRoLwXWpK9C6AcKPoRc8L85RsGILRb4AnXGyhLpKFOxTc7LAIJoHQAwEAdAgA)

[issue](https://github.com/Telefonica/mistica-design/issues/2622)


src/card-selection-context.tsx
AlexandraGallipoliRodrigues, line 1: Cards receive controls through their slots, so this internal context connects their checked state and interaction handlers without introducing additional public control props. Rows receive their controls through dedicated props and can manage that connection directly. Custom controls opt in through selected.

src/card-selection-surface.tsx
AlexandraGallipoliRodrigues, line 1: This wrapper shares the selection surface between the standard cards and AdvancedDataCard. With a single supported control, it exposes the control’s role and checked state on the card and delegates interaction to its existing handlers, avoiding a second keyboard focus target.

src/__screenshot_tests__/__image_snapshots__/card-screenshot-test-tsx-selectable-advanced-card-with-keyboard-focus-1-snap.png
Marcosld, line 1: Lines disappear in the corners?

Marcosld (commented):
Looking at the spec it says the radio / checkbox should be rendered on the top right corner and "can be moved elsewhere", can you explain this @AnaMontes11 ? It means it can be moved into the slot or where? 
Also I don't see any screenshot of the top right default case.

brtbrt:
@AlexandraGallipoliRodrigues 
I think I've found a corner case scenario where the interface is not behaving like expected?

https://mistica-nuadkga37-flows-projects-65bb050e.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8AhCAPABABwIZSgEsA7AcwF5gAmAFgF8A%2BAHWM0yQGUAXXMAa0wBnPGBiUAzFUYs2bTj35CRY4AEYAbNNay5AFRjouNTACcYpAK4AbXCYYARXDwDCtqEgD0%2BwzWbadSI4ubjI6slyEXFZiTCBBuK4mUJiQrM4AFjD8AEYYsaFhQjDRYFwwUJQAZrhWgjBsdAVhsIJgJoTYERDE5LEACtaCuJg2KW6YEJjFKd1cJhBWOLbDYLgAttmEtiPDdSVghADPxAB0%2Bf5hXBDYAIKlhN2ClADaTYVIGVl8uehvhZh8GAAT16ID2WSuJnO-zCxHWMRAUCcvDcAFowJkcnkQH9CrYtqibNliqCOMUsgdutt4okoNCYbIMYQrFAzD1YmT9g84SZMDSQjiLoUPAwADS4zAAXUaQswIqanm4vD4fjkHiV-D8njQ6D8IFFIC4mTWMEECGeIAA8lQAPqES0cECSg0Ad0IUCNZvgzwA7OoABzSoA

if you have `selected={false}` and you focus it by tabbing (or keyboard navigation), the selection as well as the the focus is showed:

https://github.com/user-attachments/assets/379feb95-ba3f-4fdd-9d5f-6cedbedea0a5

src/card-internal.tsx
```diff
@@ -67,6 +69,8 @@ type ContainerProps = {
     type: CardType;
     size: CardSize;
     variant?: Variant;
+    selected?: boolean;
+    selectionVariant?: Variant;
```
brtbrt, line 73: is this explicitly wanted?

Image

as far as I see, we should use `variant` 🤔

src/card-internal.css.ts
```diff
@@ -269,3 +269,46 @@ export const stretchedLink = style({
...
+export const selectionOutline = style({
+    selectors: {
+        '&::after': {
+            content: '',
+            position: 'absolute',
+            inset: -(selectionOutlineWidth + selectionOutlineOffset),
+            border: `${selectionOutlineWidth}px solid transparent`,
+            borderRadius: `calc(${skinVars.borderRadii.container} + 3px)`,
+            pointerEvents: 'none',
+            transition: 'border-color 0.08s ease-in-out',
```
brtbrt, line 285: is the animation defined by the designers?

src/card-internal.tsx
```diff
@@ -249,6 +255,11 @@ const Container = React.forwardRef<HTMLDivElement, ContainerProps & MediaProps &
                 }}
             >
                 <div
+                    className={classnames(
+                        selected !== undefined && styles.selectionOutline,
+                        isNaked && styles.nakedSelectionOutline,
```
brtbrt, line 260: do we need this style if naked but not selected?

src/card-internal.tsx
```diff
@@ -198,7 +202,7 @@ type TouchableProps = {
 >;
 
 type TouchableCard<T> = T & TouchableProps;
-export type MaybeTouchableCard<T> = ExclusifyUnion<TouchableCard<T> | T>;
+export type MaybeTouchableCard<T> = ExclusifyUnion<TouchableCard<T> | T> & {selected?: boolean};
```
brtbrt: you got `selected` defined on https://github.com/Telefonica/mistica-web/blob/b57e50dea549779efc2abbd005c25a1c32be2102/src/card-internal.tsx#L72

I guess this is useless? Otherwise I can't understand what's going on here and the reason of having both

src/card-internal.tsx
```diff
@@ -249,6 +255,11 @@ const Container = React.forwardRef<HTMLDivElement, ContainerProps & MediaProps &
                 }}
             >
                 <div
+                    className={classnames(
+                        selected !== undefined && styles.selectionOutline,
```
brtbrt, line 259: it should be `!!selected` right? otherwise a falsy value wouldn't work. plus, selected is supposed to be boolean, so `true` wouldn't work either

brtbrt, line 259:
```suggestion
                        selected?? false === true && styles.containerOutlineSelection,
```

src/card-internal.css.ts
```diff
@@ -269,3 +269,46 @@ export const stretchedLink = style({
...
+        },
+    },
+});
+
+export const nakedSelectionOutline = style({
+    selectors: {'&::after': {borderRadius: 0}},
+});
```
brtbrt, line 292: wouldn't a `':hover': {` selector suffice?

src/__screenshot_tests__/__image_snapshots__/card-screenshot-test-tsx-selectable-naked-cards-with-checkbox-in-top-actions-1-snap.png
yceballost, line 1: Talked with @AnaMontes11, but in the naked card case we are going to use border radius because this is quite weird

Ana, update specs please

yceballost (approved):
amazing job 🥰

src/card-internal.tsx
brtbrt, line 260: is it me or this is more standard by using this syntax?

```
[styles.selectionOutline]: selected ?? false === true,
```

src/card-internal.tsx
```diff
@@ -249,6 +255,11 @@ const Container = React.forwardRef<HTMLDivElement, ContainerProps & MediaProps &
...
+                        isNaked && styles.nakedSelectionOutline,
+                        selected && styles.selectionOutlineColor[selectionVariant]
+                    )}
```
brtbrt, line 262: to be evaluated if we really gotta use the concept of "selectionVariant" or just the variant itself

src/card-internal.tsx
brtbrt: if we really need (and want) the explicit `selected` prop, I'd change the name of `MaybeTouchableCard` because it doesn't make really sense the association of `selected` to that. I mean, it works, but semantically it doesn't really make sense 🤔
WDYT?

brtbrt, line 205: if we really need (and want) the explicit `selected` prop, I'd change the name of `MaybeTouchableCard` because it doesn't make really sense the association of `selected` to that. I mean, it works, but semantically it doesn't really make sense 🤔
WDYT?

```suggestion
type SelectableCard = {selected?: boolean};
export type MaybeTouchableCard<T> = ExclusifyUnion<TouchableCard<T> | T>; 
export type InternalCard<T> = MaybeTouchableCard<T> & SelectableCard;
```

src/card-selection-context.tsx
```diff
@@ -0,0 +1,95 @@
...
+    context: SelectionContext & {surfaceRef: React.RefObject<HTMLDivElement | null>};
+    control?: CardSelectionControl;
+    isSelected?: boolean;
+    isSelectionMode: boolean;
+    topActionIndexes: Array<number>;
+};
+
+export const useSelectableCard = (selected?: boolean): CardSelection => {
+    const [controls, setControls] = React.useState<Controls>({});
+    const surfaceRef = React.useRef<HTMLDivElement>(null);
```
brtbrt, line 38: I'm struggling understanding what a surface is. where does this concept come from?

src/card-selection-context.tsx
```diff
@@ -0,0 +1,95 @@
...
+    const values = Object.values(controls);
+    const control = values.length === 1 && values[0].surface ? values[0] : undefined;
+    const isSelected = selected ?? (values.length ? values.some((control) => control.checked) : undefined);
```
brtbrt, line 41: I think the last `undefined` is not correct. `isSelected` should always be a boolean by definition

brtbrt:
In this PR we're gonna introduce two new things: 
- the selection border thing
- ‼️ the checkboxes/radio etc as `topAction` controls: this used to be allowed on `topActions` but "it was a bug". it wasn't a case allowed! So we must pay extra attention to the API (component signature). For instance, I've figured out that `onPress` is allowed as well as a selectable checkbox on the whole component. They compete for the same action, which we should avoid. My suggestion: let's NOT allow checkboxes/radios on `topActions` and let's create mutually exclusive props `topCheckbox` `topRadio` (or something like this), which does not allow neither `onPress` not `topActions`. This way, we'll be typescript safe and we'll improve the DX. Further side effect of introducing the checkboxes/radios must be carefully analyzed I think, but with this we should be good

a comment is worth sharing in a conversation I had with @yceballost 

> por si sirve de inspo, las cards tienen un sistema similar, debería la api/firma/comoloquieras llamar, ser igual?
> https://mistica-web.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8AlGBnADhAduglgG4wAyAhgJ4QCuALgHwA6OABC0gEIQAeLmZUKPhwBzAJoBeYACYALAF8mrNuwDKtMmADWLLJphTZABkXMVK1APwQA4gCcamFrABmZagBtaANTIfqBowgAEZkOGFkQUrm5qgQAO4k%2BOgMZjGxKAlp6eZk6OgwtFJIAJKQOKoAFhCYaCKeZHYsAPT08tk5bLT4tB6BIBwR4UEdnbDoYHb4mN24EkFiMB4eWSCjOXZWEL7%2B-aHhw2vKna3rGVnHnSx5BUXApeVVNXUNTa3tl53dvf0AgpiYPojT45caTaazHDzED2GAwHDAq4qTZCbZ%2BALQsgAoFHJGnEHneJnGI3QrFMq4J61GD1DyNFptYnmb59aEAFR6OKZKjBUxm1ihQQAIhhwfzcIikSwwJUYNpgjwpMBnDA3J4fOiYPAWG4PAUWPIPnjojkkM1MolkqkQWaUFt7I4TWwzepNFoTWauNwlLaMNg8ERSJQaNaQAAaEC0WUAWwwCAA2iAALIQQhWxogAC6Efi%2BCgUfQCYAzAA2IyZ%2BRAA

I do agree: the more coherent we are with that, the better!

doc/components.md
```diff
@@ -161,6 +161,32 @@ Common props: `color`, `truncate`, `textAlign`, `as` (HTML tag), `wordBreak`, `d
...
+If there are multiple controls, the outline is selected when any is checked and the controls keep their own
+interaction targets. A single control placed in footerSlot also activates the card when clicked, while footer
+actions remain independent.
+
+```tsx
+<DataCard title="Notifications" slot={<Switch name="notifications">Enable notifications</Switch>} />
+<DataCard title="Option" slot={<Checkbox name="option">Select option</Checkbox>} />
+```
+
+`selected` is the only additional card prop. It takes precedence over the controls' state and lets custom
```
brtbrt, line 182: as I commented on the general conversation https://github.com/Telefonica/mistica-web/pull/1729/#issuecomment-5781019300 I think we should reconsider this. `selected` is probably needed anyway, but I guess it should be allowed only along with `onPress` defined.

brtbrt (changes requested):

src/card-internal.css.ts
AlexandraGallipoliRodrigues: https://github.com/Telefonica/mistica-web/pull/1721/changes#diff-ef189c10c5db4811e07e6739abc2cac3f954714e8ce0536d28f16ca1b2f2c2e5R30
i took it from the rows but i'll delete it becaus its not in this spec

src/card-internal.tsx
AlexandraGallipoliRodrigues: This internal prop controls the selection outline color, not the selector variant. The token depends on both the surrounding theme and the card variant. For example, a brand card uses controlActivated on the default theme, but controlActivatedBrand on the brand theme. I've renamed it to selectionOutlineVariant to clarify it

src/card-internal.tsx
AlexandraGallipoliRodrigues: This kept a transparent border when unselected to support the transition; the border color was only applied when selected. With the transition removed, I've simplified the condition to !!selected

src/card-internal.tsx
AlexandraGallipoliRodrigues: Updated to the object syntax, using !!selected since the prop is already typed as an optional boolean

src/card-internal.css.ts
AlexandraGallipoliRodrigues: The selection outline must remain visible without hover. The pseudo-element lets us draw a rounded border separately from the focus ring. We also encountered square native outlines in Chromium 93; AdvancedDataCard uses a separate ::before for its focus ring to avoid clipping at the corners

src/card-internal.tsx
AlexandraGallipoliRodrigues: Agreed, the name no longer describes all supported modes. I'll rename it to reflect both touchable and selectable cards, keeping the mutually exclusive props introduced in the latest API

src/card-selection-context.tsx
AlexandraGallipoliRodrigues: Surface” refers to the card's interaction area, which receives focus, clicks and keyboard events instead of the visual selector. The ref connects the selector to that element. It is an internal implementation detail

src/card-selection-context.tsx
AlexandraGallipoliRodrigues: Agreed. I'll keep isSelected as a boolean and determine selection mode separately from the explicit selector props or the presence of selected

doc/components.md
AlexandraGallipoliRodrigues: selected also supports externally controlled cards and custom controls inside slots. Those cases don't require a card-level onPress, so making it mandatory would exclude supported use cases. The explicit selector props are mutually exclusive with onPress

src/card-internal.tsx
AlexandraGallipoliRodrigues: Updated to object syntax with !!selected

src/card-internal.tsx
AlexandraGallipoliRodrigues: Updated to object syntax with !!selected

src/card-internal.tsx
AlexandraGallipoliRodrigues: This determines the outline token from both the surrounding theme and the card variant. Renamed to selectionOutlineVariant; I've explained the token mapping in the other thread

AlexandraGallipoliRodrigues:
> In this PR we're gonna introduce two new things:
> 
> * the selection border thing
> * ‼️ the checkboxes/radio etc as `topAction` controls: this used to be allowed on `topActions` but "it was a bug". it wasn't a case allowed! So we must pay extra attention to the API (component signature). For instance, I've figured out that `onPress` is allowed as well as a selectable checkbox on the whole component. They compete for the same action, which we should avoid. My suggestion: let's NOT allow checkboxes/radios on `topActions` and let's create mutually exclusive props `topCheckbox` `topRadio` (or something like this), which does not allow neither `onPress` not `topActions`. This way, we'll be typescript safe and we'll improve the DX. Further side effect of introducing the checkboxes/radios must be carefully analyzed I think, but with this we should be good
> 
> a comment is worth sharing in a conversation I had with @yceballost
> 
> > por si sirve de inspo, las cards tienen un sistema similar, debería la api/firma/comoloquieras llamar, ser igual?
> > https://mistica-web.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8AlGBnADhAduglgG4wAyAhgJ4QCuALgHwA6OABC0gEIQAeLmZUKPhwBzAJoBeYACYALAF8mrNuwDKtMmADWLLJphTZABkXMVK1APwQA4gCcamFrABmZagBtaANTIfqBowgAEZkOGFkQUrm5qgQAO4k%2BOgMZjGxKAlp6eZk6OgwtFJIAJKQOKoAFhCYaCKeZHYsAPT08tk5bLT4tB6BIBwR4UEdnbDoYHb4mN24EkFiMB4eWSCjOXZWEL7%2B-aHhw2vKna3rGVnHnSx5BUXApeVVNXUNTa3tl53dvf0AgpiYPojT45caTaazHDzED2GAwHDAq4qTZCbZ%2BALQsgAoFHJGnEHneJnGI3QrFMq4J61GD1DyNFptYnmb59aEAFR6OKZKjBUxm1ihQQAIhhwfzcIikSwwJUYNpgjwpMBnDA3J4fOiYPAWG4PAUWPIPnjojkkM1MolkqkQWaUFt7I4TWwzepNFoTWauNwlLaMNg8ERSJQaNaQAAaEC0WUAWwwCAA2iAALIQQhWxogAC6Efi%2BCgUfQCYAzAA2IyZ%2BRAA
> 
> I do agree: the more coherent we are with that, the better!

Implemented following the Row API: checkbox, switch and radioValue are mutually exclusive with each other, navigation, top actions and dismiss. Controls inside slots no longer select the card automatically; custom selection uses selected

doc/components.md
Marcosld: This case looks broken: 
Image

src/community/advanced-data-card.tsx
```diff
@@ -287,14 +294,18 @@ type AdvancedDataCardProps = MaybeTouchableCard<{
     footerTextLinesMax?: number;
     buttonLink?: RendersNullableElement<typeof ButtonLink>;
     dataAttributes?: DataAttributes;
-    actions?: ReadonlyArray<CardAction | React.ReactElement>;
+    actions?: ReadonlyArray<CardAction>;
```
Marcosld: This could be a breaking change

src/community/advanced-data-card.tsx
```diff
@@ -326,14 +337,23 @@ export const AdvancedDataCard = React.forwardRef<HTMLDivElement, AdvancedDataCar
...
     ) => {
-        const isTouchable = !!touchableProps.href || !!touchableProps.onPress || !!touchableProps.to;
-
+        const hasSelector = !!checkbox || !!switchProps || radioValue !== undefined;
+        const selection = useSelectableCard(selected, hasSelector);
+        const outsideVariant = useRawThemeVariant();
+        const isTouchable =
+            !!selection.control ||
+            (!selection.isSelectionMode &&
+                (!!touchableProps.href || !!touchableProps.onPress || !!touchableProps.to));
```
Marcosld: Isn't this already guarenteed by typescript types?

src/community/advanced-data-card.tsx
```diff
@@ -326,14 +337,23 @@ export const AdvancedDataCard = React.forwardRef<HTMLDivElement, AdvancedDataCar
...
         const footerProps = {button, footerImage, footerText, footerTextLinesMax, buttonLink};
 
         const hasFooter = !!button || !!footerImage || !!footerText || !!buttonLink;
         const hasSlots = !!slot?.length;
 
-        const topActionsCount = (actions?.length || 0) + (onClose ? 1 : 0);
+        const topActionsCount = selection.isSelectionMode
+            ? hasSelector
```
Marcosld: weird if, probably it could be simplified

src/card-selection-surface.tsx
```diff
@@ -0,0 +1,98 @@
...
+    const id = React.useId();
+    const config = checkbox || switchProps;
+    if (!config && radioValue === undefined) {
+        return null;
+    }
+    const Control = switchProps ? Switch : Checkbox;
+    return (
+        <ThemeVariant variant={variant}>
+            <div className={styles.topActionsContainer} data-testid="cardSelector">
+                <CardSelectionContext.Provider value={selection.context}>
```
Marcosld: This context doesn't look necessary at all... controls could just receive the props and you wouldn't need to modify their code either.. right?

doc/components.md
AlexandraGallipoliRodrigues: Fixed. The reserved space could shrink and was based on an icon button's width. It now accommodates the switch dimensions, including iOS. Added screenshot coverage for Blau, Vivo and Vivo Evolution at 360px

src/community/advanced-data-card.tsx
AlexandraGallipoliRodrigues: True, restored

src/community/advanced-data-card.tsx
AlexandraGallipoliRodrigues: The explicit selector props exclude navigation through TypeScript. However, selected can also be used independently for external selection, so that case still needs a runtime check. I've made the distinction explicit

src/community/advanced-data-card.tsx
AlexandraGallipoliRodrigues: simplified the nested condition. It now counts one explicit selector, or the visible top actions and dismiss button

src/card-selection-surface.tsx
AlexandraGallipoliRodrigues: Agreed, restored

AnaMontes11 (approved):
Good job!

src/card-internal.tsx
```diff
@@ -67,6 +70,8 @@ type ContainerProps = {
     type: CardType;
     size: CardSize;
     variant?: Variant;
+    selected?: boolean;
+    selectionOutlineVariant?: Variant;
```
brtbrt: is this needed? can't we use the card variant? I'm not finding any doc talking about this feat

src/theme-variant-context.tsx
```diff
@@ -30,5 +30,6 @@ export const ThemeVariant = ({variant, children}: ThemeVariantProps): JSX.Elemen
     <ThemeVariantContext.Provider value={variant ?? 'default'}>{children}</ThemeVariantContext.Provider>
 );
 
-export const useThemeVariant = (): NonDeprecatedVariant =>
-    normalizeVariant(React.useContext(ThemeVariantContext));
+export const useRawThemeVariant = (): Variant => React.useContext(ThemeVariantContext);
+
+export const useThemeVariant = (): NonDeprecatedVariant => normalizeVariant(useRawThemeVariant());
```
Marcosld: what is this for? why do we need a new hook?

src/community/advanced-data-card.css.ts
```diff
@@ -269,3 +270,28 @@ export const flexColumn = style({
...
+globalStyle(`${selectionTouchable} > div:focus-visible::before`, {
+    content: '""',
+    position: 'absolute',
+    inset: 0,
+    border: '1.5px solid',
+    borderColor: skinVars.colors.controlActivated,
+    borderRadius: 'inherit',
+    pointerEvents: 'none',
+    zIndex: 2,
+});
```
Marcosld: does this only happen for chrome 93?

src/community/advanced-data-card.css.ts
```diff
@@ -269,3 +270,28 @@ export const flexColumn = style({
...
+
+const switchWidth = {default: 34 + 8, ios: 51};
```
Marcosld: why 34+8? why not a fixed number?

src/community/advanced-data-card.tsx
```diff
@@ -319,21 +330,29 @@ export const AdvancedDataCard = React.forwardRef<HTMLDivElement, AdvancedDataCar
...
             'aria-label': ariaLabelProp,
-            onClose,
+            onClose: onCloseProp,
             ...touchableProps
         },
         ref
     ) => {
-        const isTouchable = !!touchableProps.href || !!touchableProps.onPress || !!touchableProps.to;
-
+        const hasSelector = !!checkbox || !!switchProps || radioValue !== undefined;
```
Marcosld: isn't it enough with the useSelectableCard hook?

src/community/advanced-data-card.tsx
```diff
@@ -319,21 +330,29 @@ export const AdvancedDataCard = React.forwardRef<HTMLDivElement, AdvancedDataCar
...
+        const selection = useSelectableCard(selected, {checkbox, switch: switchProps, radioValue});
```
Marcosld: I think I'd move the hook to card-internal instead of its own file

src/card-selection.tsx
```diff
@@ -0,0 +1,65 @@
...
+};
+
+export type CardSelectionProps = ExclusifyUnion<
+    {checkbox: ControlProps} | {switch: ControlProps} | {radioValue: string}
+>;
+
+type CardSelection = {
+    checkbox?: ControlProps;
+    switch?: ControlProps;
+    radioValue?: string;
```
brtbrt: we're passing radioValue + switch + checkbox to always return them as-is. isn't that weird?

src/community/advanced-data-card.tsx
```diff
@@ -319,21 +330,29 @@ export const AdvancedDataCard = React.forwardRef<HTMLDivElement, AdvancedDataCar
...
+        const actions = selection.isSelectionMode ? undefined : actionsProp;
+        const onClose = selection.isSelectionMode ? undefined : onCloseProp;
```
Marcosld: I think we should avoid these kind of checks and rely on typescript or we will have our code full of checks

src/community/advanced-data-card.tsx
Marcosld: mmm if selected is used independently why do we need this logic? It shouldn't affect touchable behavior of the component, right? I mean, a card can have a touch behavior and also be selected or not if controlled externally

brtbrt:
do we want this case to be handed like this?
```
        checkbox={{name: "datacard-checkbox", value: true}}
        selected={false}
```

[mistica example](https://mistica-k7g4j95rl-flows-projects-65bb050e.vercel.app/playroom#?code=N4Igxg9gJgpiBcIA8AhCAPABABwIZSgEsA7AcwF5gAmAFgF8A%2BAHWM0yQGUAXXMAa0wBnPGBiUAzFUYs2bTj35CRY4AEYAbNNay5AFRjouNTACcYpAK4AbXCYYARXDwDCtqEgD0%2BwzWbadSI4ubjI6slyEXFZiTCBBuK4mUJiQrM4AFjD8AEYYsaFhmLCCYCaE2BEQxOSxAArWgriYNilumBCYMFYpVVwmEN14Jk1guAC22YS2zU2CXVlghADPxAB0%2Bf5hYJk5GJTAxOMw8JixUE68bgC021l8ueixADSYAG64VhbHmH1fdHQFMJzaJgLgwKCUABmHzmAM2bA8fgCHm4vD4SPYKIU6JYnjQ6AYIDoQA)

maybe we should not allow `value` if `selected` is true? or even better (in my head) we should not allow selected + checkbox (and the other control props!) and they should be mutually exclusive?

src/community/advanced-data-card.tsx
```diff
@@ -419,10 +456,14 @@ export const AdvancedDataCard = React.forwardRef<HTMLDivElement, AdvancedDataCar
...
                         )}
-                    </Touchable>
+                    </CardSelectionSurface>
                     {hasFooter && <CardFooter {...footerProps} />}
                 </Boxed>
-                <TopActions actions={actions} onClose={onClose} />
+                <TopActions
+                    actions={actions}
+                    onClose={onClose}
+                    variant={selection.isSelectionMode ? 'default' : undefined}
```
Marcosld: what is this new check for?

src/card-internal.tsx
```diff
@@ -1379,7 +1437,7 @@ export const InternalCard = React.forwardRef<HTMLDivElement, MaybeTouchableCard<
 
         const shouldShowVideo = hasMediaVideo || hasBackgroundVideo;
         const {video, videoAction} = useVideoWithControls({
-            src: shouldShowVideo ? videoSrc : undefined,
+            src: shouldShowVideo && !isSelectionMode ? videoSrc : undefined,
```
Marcosld: we can't play video on selection mode?

src/card-internal.tsx
```diff
@@ -1341,17 +1386,30 @@ export const InternalCard = React.forwardRef<HTMLDivElement, MaybeTouchableCard<
...
+            checkbox,
+            switch: switchProps,
+            radioValue,
             segregateTouchableContent = false,
             touchableAriaLabel: touchableAriaLabelProp,
             ...touchableProps
         },
         ref
     ): JSX.Element => {
+        const hasSelector = !!checkbox || !!switchProps || radioValue !== undefined;
```
Marcosld: same than in the other component

src/card-internal.tsx
```diff
@@ -1403,11 +1462,13 @@ export const InternalCard = React.forwardRef<HTMLDivElement, MaybeTouchableCard<
...
         const shouldShowFooter =
-            (showFooterProp && (hasButtons || !!footerSlot)) || (hasButtons && isTouchable);
+            (showFooterProp && (hasButtons || !!footerSlot)) ||
+            (hasButtons && (isTouchable || isSelectionMode));
 
         const showButtonsInBody = !shouldShowFooter && hasButtons;
 
-        const topActionsLengthWithoutVideo = (topActions?.length || 0) + (onClose ? 1 : 0);
+        const topActionsLengthWithoutVideo =
+            (isSelectionMode ? (hasSelector ? 1 : 0) : topActions?.length || 0) + (onClose ? 1 : 0);
```
Marcosld: simplify

src/card-selection-surface.tsx
```diff
@@ -0,0 +1,66 @@
...
+        return (
+            <div className={classnames(touchableStyles.notTouchable, props.className)} style={props.style}>
+                {props.children}
+            </div>
+        );
+    }
+
+    const render = ({controlElement}: {controlElement: React.ReactElement}) => (
+        <>
+            <ThemeVariant variant={contentVariant}>{props.children}</ThemeVariant>
```
Marcosld: what is the theme variant trickery for?

src/card-selection-surface.tsx
```diff
@@ -0,0 +1,66 @@
...
+import * as styles from './card-internal.css';
+import * as touchableStyles from './touchable.css';
+
+import type {useSelectableCard} from './card-selection';
+import type {Variant} from './theme-variant-context';
+import type {TouchableProps} from './touchable';
+
+type Props = TouchableProps & {selection: ReturnType<typeof useSelectableCard>; variant?: Variant};
+
+export const CardSelectionSurface = ({selection, variant, ...props}: Props): JSX.Element => {
```
Marcosld: we should probably move all this to card-internal, right?

src/card-selection.tsx
```diff
@@ -0,0 +1,65 @@
...
+    hasSelector: boolean;
+    isSelected: boolean;
+    isSelectionMode: boolean;
+    onChange: () => void;
+};
+
+export const useSelectableCard = (
```
brtbrt: I'd docblock it

src/card-selection-surface.tsx
```diff
@@ -0,0 +1,66 @@
...
+    const render = ({controlElement}: {controlElement: React.ReactElement}) => (
+        <>
+            <ThemeVariant variant={contentVariant}>{props.children}</ThemeVariant>
+            <div className={styles.topActionsContainer} data-testid="cardSelector" aria-hidden>
+                {controlElement}
+            </div>
+        </>
+    );
+    const config = selection.checkbox || selection.switch;
+    const Control = selection.switch ? Switch : Checkbox;
```
brtbrt: I think we do got room for improvement on this semantic:

1. `config` is way too generic. maybe `nonRadioSelectionConfig` or something like this?
2. `Control` can be spared in case we use a `RadioButton`

src/card-selection-surface.tsx
brtbrt: the fact we got two variants: I don't understand it 🥲 please help

src/card-selection-surface.tsx
brtbrt: I understand the concept of surface, but I'd be stick to the `Row` way of handling this, in order to not introduce further concepts we already have on other similar components and behavior.
Maybe Claudio suggested you the name, and it's actually not that bad, but in this case we need to steer its behavior, because it's way too general purpose :-)
Long story short: 

according to this:

```
    const [isChecked, toggle] = useControlState(props.switch || props.checkbox || {});
    const isControlSelected =
        props.switch || props.checkbox
            ? isChecked
            : props.radioValue !== undefined
              ? radioContext.selectedValue === props.radioValue
              : undefined;
```

I'd call it `CardWithControl` or something like this (the "control" thing is the concept I'd like to convey)

Plus: I'm a big fan of SRP, but in this case I'd say we can spare this module by placing this logic into `card-internal` itself.

last thing `useSelectableCard`: mh, I see extra things we probably don't need it. wouldn't it be enough `useControlState` like on the row?
long story short: I'd be stick the most to the row solution

